PR #4, opened 2020-04-27
Enable dictionary-structured VSS files.

magnusfeuer:
Supports both old-style, list-structured and new-style, dictionary-structured vspec files.
See: https://github.com/GENIVI/vehicle_signal_specification/issues/149

### DIFF
--- a/vspec.py
+++ b/vspec.py
@@ -13,6 +13,8 @@ import yaml
 import json
 import os
 import uuid
+import collections
+
 
 class VSpecError(Exception):
     def __init__(self, *args, **kwargs):
@@ -29,7 +31,7 @@ class VSpecError(Exception):
 #
 class SignalUUIDManager:
     NAMESPACE = "vehicle_signal_specification"
-    
+
     def __init__(self):
         self.signal_uuid_db_set = {}
         self.namespace_uuid = uuid.uuid5(uuid.NAMESPACE_OID, self.NAMESPACE)
@@ -212,6 +214,23 @@ def load(file_name, include_paths):
     return deep_model["children"]
 
 
+def convert_yaml_to_list(raw_yaml):
+    if isinstance(raw_yaml, list):
+        return raw_yaml
+
+    # Sort the dictionary according to line number.
+    # The reason is that when the YAML file is loaded
+    # the object order is not preserved in the created
+    # dictionary
+    raw_yaml = collections.OrderedDict(sorted(raw_yaml.items(), key=lambda x: x[1]['$line$']))
+    lst = []
+    for elem in raw_yaml:
+        if isinstance(raw_yaml[elem], dict):
+            raw_yaml[elem]['$name$'] = elem
+            lst.append(raw_yaml[elem])
+
+    return lst
+
 def load_flat_model(file_name, prefix, include_paths):
     # Hooks into YAML parser to add line numbers
     # and file name into each elemeent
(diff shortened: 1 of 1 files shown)


vspec.py
```diff
@@ -225,15 +244,31 @@ def yaml_compose_node(parent, index):
...
-        # Find the name of the branch / signal and convert
-        # it to a dictionary element '$name$'
+
+        # Replace
+        # { 'Vehicle.Speed': { 'datatype': 'boolean', 'type': 'sensor' }}
+        # with
+        # { '$name$': 'Vehicle.Speed', 'datatype': 'boolean', 'type': 'sensor' }
+
         for key, val in list(mapping.items()):
+            # Don't convert metadata.
```
gunnarx, line 271: Moderately confusing comment since we often call the listed attributes in VSS (datatype, description, ...) "metadata".  Here you mean maybe something like 'special directives' ?

vspec.py
```diff
@@ -212,6 +214,23 @@ def load(file_name, include_paths):
...
 
+def convert_yaml_to_list(raw_yaml):
+    if isinstance(raw_yaml, list):
+        return raw_yaml
+
+    # Sort the dictionary according to line number.
+    # The reason is that when the YAML file is loaded
+    # the object order is not preserved in the created
+    # dictionary
+    raw_yaml = collections.OrderedDict(sorted(raw_yaml.items(), key=lambda x: x[1]['$line$']))
```
gunnarx, line 225: Here (and even more elsewhere) "line_number" might a bit more understandable than "line"

vspec.py
```diff
@@ -225,15 +244,31 @@ def yaml_compose_node(parent, index):
         except yaml.parser.ParserError as e:
             raise VSpecError(file_name, line + 1, e)
 
-        node.__line__ = line + 1
-        node.__file_name__ = file_name
+        if node.value == '$include$':
+            node.value = f'$include${load_flat_model.include_index}'
```
gunnarx, line 248: @magnusfeuer Is this **f** in ```f'$include...'``` special syntax or a typo?
Python seems to complain about it:
```
    node.value = f'$include${load_flat_model.include_index}'
                                                           ^
SyntaxError: invalid syntax
```
If it's some python3 unique thing, it doesn't make sense since the top of the file specifies python2 as interpreter...

SebastianSchildt: @gunnarx  I am confused, where does it mention py2? The setup.py explicitly mentions py 3, and so does vspec2json.py.  vspec.py has no shebang at all (and if I understand correctly isn't called directly anyway), or am I missing something?

SebastianSchildt:
```  
File "vspec.py", line 123, in get_or_assign_signal_uuid
    sys.exit(255)
NameError: name 'sys' is not defined

```

because vspec.py is missing `import sys`

SebastianSchildt:
Ok, not a comprehensive test, but without triggering the error mentioned above I can create a JSON from VSS master (so I guess "old-style input"), that can be parsed & used by kuksa.val. So yeah, "works for me". 👍

vspec.py
gunnarx: I'm so sorry, it was my own program that was requiring python2 and I was totally confused apparently.  Thanks @SebastianSchildt for checking my BS...  So we can conclude that the syntax is python3 supported.  I'll fix my program to be python3 compliant and move on.

danielwilms:
The old list vspec file works fine for me with the old and new implementation. If I use the object-style, like e.g.:

```
#
# The vehicle branch for highlevel vehicle signals and attributes.
#
Vehicle:
  type: branch
  description: High-level vehicle data.

# Include the Vehicle/Vehicle.vspec file and attach all its signals under the
# Vehicle branch created above.

#include Vehicle/Vehicle.vspec Vehicle
```

I get the following issue:
```
./vss-tools/vspec2csv.py -i:spec/VehicleSignalSpecification.id -I ./spec ./spec/VehicleSignalSpecification.vspec vss_rel_$(cat VERSION).csv
Error: ./spec/VehicleSignalSpecification.vspec: 22: while parsing a block mapping
  in "<unicode string>", line 22, column 1:
    Vehicle:
    ^
expected <block end>, but found '-'
  in "<unicode string>", line 29, column 1:
    - $include$:
    ^
make: *** [csv] Error 255
```
Can you guys confirm it?

SebastianSchildt:
I can confirm that (with vspec2json) I did
 - using this branch and vanilla VSS model -> vspec2json -> works
 - replacing "VehicleSignalSpecification.vspec" with your snippet -> vspec2json -> breaks

magnusfeuer:
The include bug is trickier to fix than I first thought. I'll continue to work on it tomorrow.

magnusfeuer:
Fixed the include issue.

Tests done:
Original VSS spec.
VSS spec converted from list to object format, with original `#include` retained
VSS spec converted from list to object format, with original `#include` converted to `$inlcude:`
VSS spec with only the `Cabin/` directory converted to object format, with original `#include` retained
VSS spec with only the `Cabin/` directory converted to object format, with  `#include` converted to `$inlcude:`

All test cases were run through vspec2json.py, and all generated JSON specs had the same md5 checksum.

@danielwilms Please retest.

Note: The code is getting **really** ugly, and is in need of a rewrite.

danielwilms:
> @danielwilms Please retest.

Thanks @magnusfeuer, I confirm that it works.

SebastianSchildt:
I don't see any blockers. Why isn't this merged yet? I think currently this not being merged blocks useful work on https://github.com/GENIVI/vss-tools/pull/7 

Do we still have a problem here? @magnusfeuer @gunnarx

PR #6, opened 2020-06-02
VSS to Franca IDL generator

gunnarx:
OK, so this is the first preview -- NOW UPDATED --, I just wanted to put it up for a first look.

I might want to squash the commits (what is your opinion?) before we merge.

Also I have a few more compatibility tests to run on the results, but the result looks kind of OK I think.

This converter translates VSS to Franca interfaces and primarily the feature that mirrors the concept of a signal, which is *attribute* because it is a named data item that can most usually also be "observed" i.e. subscribe to changes.

In addition it allows a hierarchical nested interfaces (namespaces) that mirror the branches and a pattern matching to specify what to include.  Through the parameter it is very flexible to decide how this mapping to packages, interface hierarchy and included parts shall be done.   It is expected to run the program multiple times to create separate franca files, for separate interfaces, thus dividing the large VSS file into multiple Franca interface files.

This addresses https://github.com/GENIVI/vehicle_signal_specification/issues/148


gunnarx:
Since the minor changes are not squashed, it is probably easiest view [the whole final file](https://github.com/GENIVI/vss-tools/blob/a9f4ffae9ccddacbd45176aecf1561ecbd17d14f/vspeccsv2franca.py)

gunnarx:
This needs some usage documentation also.  To get started here are some hints and examples:

First, as you can see this program takes the complete signal list from the **generated CSV file**.
That way I did not need to traverse the tree datatype.  String-matching against the fully qualified path was a much easier approach.

--> You must first run  ```make``` or specifically ```vspec2csv.py``` to generate **VSS in CSV format**

Example 1:   with comments
```
./vspeccsv2franca.py 
 -V 1,2                       # This defines the Franca interface version in major,minor (as supported by Franca IDL itself)
 -v "VSS-2.0--internaltag"    # This optionally defines a constant string for the VSS version, in the result to read out the VSS version.  The constant is placed inside the namespace to avoid clashes.
 -p MyFrancaPackage          # This defines the Franca package (required) 
-n Vehicle 
-n Drivetrain                                 
-n FuelSystem                 # This is a sequence of nested interface name. They must match the VSS hierarchy in both name (case sensitive) and order!
-s Vehicle.Drivetrain.Fuel*   # This is a path pattern match to specify which nodes to include 
../vss_rel_2.0.0-alpha+006.csv  
Drivetrain-FuelSystem.fidl
```

generates:
```
// [trimmed comments]

package MyFrancaPackage

interface Vehicle {
    interface Drivetrain {
        interface FuelSystem {
            version { major 1, minor 2 }
            const UTF8String VSS_INTERFACE_VERSION = "VSS-2.0--internaltag"
            attribute String FuelType
            attribute String HybridType
            attribute UInt16 TankCapacity
            attribute UInt8 Level
            attribute UInt32 Range
            attribute Float InstantConsumption
            attribute Float AverageConsumption
            attribute Float ConsumptionSinceStart
            attribute UInt32 TimeSinceStart
            attribute Boolean EngineStopStartEnabled
            attribute Boolean LowFuelLevel
        }
    }
}
// End of file
```

Noteworthy:
1. The interface hierarchy guarantees namespacing of the node names.  For simplicity, it is required to use the exact matching name, and order, of branches that matches the VSS tree.
2. Datatypes are translated from to the native types that Franca includes out of the box.  All of them seem to map (enum still pending?) without using complex datatypes in Franca, which is also supported of course.

Then here is another example. If running exactly the same as the above but omitting the ```-n Fuelsystem``` from the interface hierarchy we get this:

```
// [trimmed comments]

package MyFrancaPackage

interface Vehicle {
    interface Drivetrain {
        <<-- COMMENT: NOTE LACK OF FuelSystem INTERFACE HERE -->>
        version { major 1, minor 2 }
        const UTF8String VSS_INTERFACE_VERSION = "VSS-2.0--internaltag"
        attribute String FuelSystem_FuelType
        attribute String FuelSystem_HybridType
        attribute UInt16 FuelSystem_TankCapacity
        attribute UInt8 FuelSystem_Level
        attribute UInt32 FuelSystem_Range
        attribute Float FuelSystem_InstantConsumption
        attribute Float FuelSystem_AverageConsumption
        attribute Float FuelSystem_ConsumptionSinceStart
        attribute UInt32 FuelSystem_TimeSinceStart
        attribute Boolean FuelSystem_EngineStopStartEnabled
        attribute Boolean FuelSystem_LowFuelLevel
    }
}
// End of file
```
Noteworthy:
1. The branch "FuelSystem" that exists in the full hierarchy was here _not_ specified to be a nested Interface.  That is therefore prepended to the attribute name.  Thus the full "namespacing" of node names is still preserved and name clashes should be impossible.

(This is particularly useful if the pattern matching was set up to match more than one subtree under Drivetrain)

By using different sequences of -n <interfacename> and -s <pattern> it is possible to arrange the wanted signals in the wanted namespaces.  All branches are thus not required to be its own nested interface, but for those that are specified, it is  **required** to use the exact names and define them in the same order as the VSS branch names. 

It is possible to flatten the namespace hierarchy by using the dot notation instead:
E.g. ```-n Vehicle.Drivetrain -n FuelSystem``` would yield:
```
interface Vehicle.Drivetrain {
   interface FuelSystem {
      attribute String FuelType
      ...
```
or ```-n Vehicle.Drivetrain.FuelSystem```  would yield:
```
interface Vehicle.Drivetrain.FuelSystem {
    attribute String FuelType
    ...
```
As we can see for the signalname ```FuelType```, the namespacing and uniqueness of node names should be preserved in every choice of interface hiearchy, and there is flexibility in how to translate each VSS subtree to a Franca IDL interface.

Note that all names and matches are **Case Sensitive** and must match what VSS is using.

vspeccsv2franca.py
```diff
@@ -0,0 +1,225 @@
...
+
+    try:
+        franca_out = open (args[1], "w")
+    except OSError as e:
+        print("Error opening file %s for writing?" % args[1])
+        print("Error: {}".format(e))
+        exit(255)
+
+
+    print(signal_patterns)
```
gunnarx, line 184: This is a debug printout -- should be removed.

gunnarx:
@gunnarx TODO: Remove debug printout for [ signal patterns ], and clarify output of number of matches.

gunnarx:
Turns out still some work to make the output compatible with code-generators.  Working on it and an update will come.

gunnarx:
Updated, as seen above.  In my tests, the resulting .fidl files are now accepted by CommonAPI code generators, and appropriate code is generated.  I hope to publish my testing environment soon to allow others to reproduce that.   The changed solution however means there are even more configurability flags to the VSS to FIDL converter so this needs some documentation and examples.

danielwilms:
>     attribute String FuelSystem_FuelType
        attribute String FuelSystem_HybridType
        attribute UInt16 FuelSystem_TankCapacity
        attribute UInt8 FuelSystem_Level
        attribute UInt32 FuelSystem_Range
        attribute Float FuelSystem_InstantConsumption
        attribute Float FuelSystem_AverageConsumption
        attribute Float FuelSystem_ConsumptionSinceStart
        attribute UInt32 FuelSystem_TimeSinceStart
        attribute Boolean FuelSystem_EngineStopStartEnabled
        attribute Boolean FuelSystem_LowFuelLevel

@gunnarx: everything maps to an attribute, so quite a bit of information (attribute vs. sensor, actuator) is lost. Does that matter in Franca? I would expect something like const/readonly/read-write to map with the node types mentioned earlier.

gunnarx:
Agree, it needs some rework to handle VSS attribute as a Franca IDL const instead.

(By the way, this is an unfortunate name clash?  attribute in VSS is a constant, whereas in Franca it means an observable (changeable) property.

gunnarx:
REMINDER: Array support must be added also, see #22

gunnarx:
Rebased to fix conflict, moved to contrib, and removed uuid -i flag, as done for other tools.
The build failure can't be fixed in this PR alone since it stems from the Makefile in the VSS parent.  (The path and name of the tool has changed), but I might try sending a corresponding PR to VSS repo that pulls in this commit in the submodule to test it.

gunnarx:
I will close this until it becomes a higher priority to fix this.  I think putting work on a VSC<->Franca translation is better spent time in the short term, and then further discussion about relationship and joint usage of VSS and VSC.

Issue #5, opened 2020-05-31
Parsing into typesafe structure


Floix:
VSS tools parsing result currently parses into a Dictionary structure. This has some disadvantages over using a typesafe (as far as it gets in python) tree structure.
1.) Missing documentation and autocompletion during development time. 
2.) Missing checks and validations (e.g., various ways to express a unit like l, liters, liter, Liter, LITER)
3.) Implementation of traversal operations during generation steps.

We'd like to introduce anytree (https://anytree.readthedocs.io/en/latest/) along with entity classes and enumerations for VSS Nodes. We'll benefit from this library traversal functions (Post-Order, Pre-Order, Level-Order traversal), document our node structure (meta-model) as code annotations and simplify development.

Sponsor: @danielwilms 
Presented at VSS call 27.04.2020

danielwilms:
Reference to [issue 150](../../vehicle_signal_specification/issues/150) in the VSS repo

Floix:
From https://github.com/GENIVI/vehicle_signal_specification/issues/150 this PR fixes:
* Node types: one of branch, attribute, sensor, actuator, stream
* Data types, we have a similar list
* Units: we have a similar list, but this one might change with new signals.

gunnarx:
I'll probably see the advantages better when starting to use Anytree traversal functions.

I'd just point out that implementing checks such as number 2) seems possible to achieve without turning the strings to classes and enums.  Just implement the checks to enforce some convention I mean.
 
> 2.) Missing checks and validations (e.g., various ways to express a unit like l, liters, liter, Liter, LITER)

gunnarx:
We merged because wanted most of the code available.  This is still not explained however:
https://github.com/GENIVI/vss-tools/pull/7#discussion_r447283118

gunnarx:
I'm closing this since the issue is done but opening a new one to explain the UINT/INT discrepancy.

PR #7, opened 2020-06-03
Feature/typesafe parsing

Floix:
Still a preview, so please don't merge yet. Improves issue described here: https://github.com/GENIVI/vss-tools/issues/5

I'd like to ask for comments (e.g., @danielwilms @gunnarx @magnusfeuer)

What's still missing:
* Additional tests on nested structures
* Additional documentation
* Versioning for proper deprecation of features
* Migration of CSV converter to demonstrate approach and guide other migrations

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,0 +1,62 @@
+# VSS Tools Usage
+
+## Contents
+1. Prerequisites
+2. Project Setup
+3. Run Tests
+4. Use Project
+
+## 1. Prerequisites
+* Python 3.7 installed
+* If the installation (pip install) is executed behind a (corporate) proxy, the following environments variables must be set: http_proxy and https_proxy (including authentication e.g., http://${proxy_username):$(proxy_password)@yourproxy.yourdomain)
+* If you do not run with administration rights, you may need to configure pip target path to write to your user home directory.
+
+```
+On Unix and Mac OS X the configuration file is: $HOME/.pip/pip.conf
+If the file does not exist, create an empty file with that name.
+
+Add or replace the following lines:
+[global]
+target=/somedir/where/your/account/can/write/to
+
+On Windows, the configuration file is: %HOME%\pip\pip.ini 
+If the file does not exist, create an empty file with that name.
+
+Add or replace the following lines:
+[global]
+target=C:\SomeDir\Where\Your\Account\Can\Write\To
+```
+
+* If you use a custom pip installation directory, set the PYTHONPATH enviornment variable to the directory that you set in the pip.ini file.
+
+## 2. Project Setup
+* Checkout vss-tools as submodule of the Vehicle Signal Specification repository (git clone --recurse-submodules -j8 https://github.com/GENIVI/vehicle_signal_specification.git)
+* RUN  ```pip install -e .``` from the vss-tools project root directory
+* Run  ```pip install -r requirements.txt```  from the vss-tools project root directory
+
+## 3. Run Tests
+Run the following command in your root directory
+* RUN ```pytest tests ```
+
+## 4. Use Project
+
+Folder structure for the example below
+```bash
+.
+└── vehicle_signal_specification
+    ├── VERSION
+    ├── spec
+    │   │   ├── VehicleSignalSpecification.id
+    │   └── VehicleSignalSpecification.vspec
+    ├──vss-tools  
+    ├── vspec2c.py
+    ├── vspec2cnative.py
+    ├── vspec2csv.py
+    ├── vspec2franca.py
+    ├── vspec2json.py
+    └── vspec2json.py
+```
+
+* CSV File generation -> vss.csv:
+ ```python /vspec2csv.py -i:../vehicle_signal_specification/spec/VehicleSignalSpecification.id ../vehicle_signal_specification/spec/VehicleSignalSpecification.vspec vss.csv```
+
--- a/model/constants.py
+++ b/model/constants.py
@@ -1,0 +1,142 @@
+#!/usr/bin/python
+
+#
+#
+#
+# All files and artifacts in this repository are licensed under the
+# provisions of the license provided by the LICENSE file in this repository.
+#
+
+#
+# Constant Types and Mappings
+#
+
+# noinspection PyPackageRequirements
+from enum import Enum
+
+
+class StringStyle(Enum):
+    NONE = "none"
+    CAMEL_CASE = "camelCase"
+    CAMEL_BACK = "camelBack"
+    CAPITAL_CASE = "capitalcase"
+    CONST_CASE = "constcase"
+    LOWER_CASE = "lowercase"
+    PASCAL_CASE = "pascalcase"
+    SENTENCE_CASE = "sentencecase"
+    SNAKE_CASE = "snakecase"
+    SPINAL_CASE = "spinalcase"
+    TITLE_CASE = "titlecase"
+    TRIM_CASE = "trimcase"
+    UPPER_CASE = "uppercase"
+    ALPHANUM_CASE = "alphanumcase"
+
+    @staticmethod
+    def from_str(name):
+        for style in StringStyle:
+            if style.value == name:
+                return style
+        raise Exception("Invalid String style %s only support %s" % (name, StringStyle.values()))
+
+    @staticmethod
+    def values():
+        return list(map(lambda v: v.value, StringStyle))
+
+
+class Unit(Enum):
+    MILIMETER = "mm"
+    CENTIMETER = "cm"
+    METER = "m"
+    KILOMETER = "km"
+    KILOMETERPERHOUR = "km/h"
+    METERSPERSECONDSQUARED = "m/s2"
+    LITER = "l"
+    DEGREECELCIUS = "celsius"
+    DEGREE = "degrees"
+    DEGREEPERSECOND = "degrees/s"
+    KILOWATT = "kW"
+    KILOWATTHOURS = "kWh"
+    KILOGRAMM = "kg"
+    VOLT = "V"
+    AMPERE = "A"
+    SECOND = "s"
+    MINUTE = "min"
+    WEEKS = "weeks"
+    MONTHS = "months"
+    UNIXTIMESTAMP = "UNIX Timestamp"
+    PASCAL = "pa"
+    KILOPASCAL = "kpa"
+    PERCENT = "percent"
+    CUBICMETER = "cm3"
+    HORSEPOWER = "PS"
+    STARS = "stars"
+    GRAMMPERSECOND = "g/s"
+    GRAMMPERKM = "g/km"
+    KILOWATTHOURSPER100KM = "kWh/100km"
+    LITERPER100KM = "l/100km"
+    LITERPERHOUR = "l/h"
+    MILESPERGALLON = "mpg"
+    POUND = "lbs"
+    NETWONMETER = "N.m"
+    REVOLUTIONSPERMINUTE = "rpm"
+    INCH = "inch"
+    RATIO = "ratio"
+
+    @staticmethod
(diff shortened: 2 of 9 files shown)


model/vsstree.py
```diff
@@ -0,0 +1,141 @@
...
+
+    def is_orphan(self):
+        if self.type == VSSType.BRANCH or self.type == VSSType.RBRANCH:
+            return self.is_leaf
+        return False
+
+    def is_deprecated(self):
+        return self.deprecation is not None
+
+    def supports_unit(self):
```
gunnarx, line 139: A lot of these functions I see are not used in the current source code.  I think you have some idea for future usage?  Let's discuss those details, please.

model/vsstree.py
```diff
@@ -0,0 +1,141 @@
...
+    deprecation = None
+
+    aggregate = False
+    child_type = None
+    child_properties = 0
+    prop_name = None
+    prop_type = None
+    prop_format = None
+    prop_unit = None
+    prop_value = None
```
gunnarx, line 39: Please explain these.

model/constants.py
```diff
@@ -110,7 +122,7 @@ class VSSDataType(Enum):
     INT32 = "Int32"
     UINT32 = "UInt32"
     INT64 = "Int64"
-    UINT64 = "UInt64"
+    UINT64 = "Int64"
```
gunnarx, line 125: ?

gunnarx:
Thanks for the PR @Floix 

I have a few review comments and then mostly questions about how this fits into the bigger context and what future use we will have that warrants building all this "structure".    I imagine defining all those classes are needed for AnyTree to be used?    I think once we have an example of practical usage, the advantages might be clearer.

vspec.py
```diff
@@ -203,6 +210,9 @@ def assign_signal_uuids(flat_model):
     return flat_model
 
 
+@deprecation.deprecated(deprecated_in="2.0", removed_in="2.0",
```
gunnarx, line 212: Can we talk about the intention/usage of the deprecation library?  
Is it for features that may be added/removed from VSS, or from the tooling?

gunnarx:
General on the set of commits:
I assume we will squash and clean up the commits a bit before merging but just some example feedback on the current commits:
* 2c7f685 says it changes requirements and readme, but it also has source file changes in --- vspec.py and vspec2csv.py -- since 3a5a961 already modifies those files, I'd put the changes there instead.
Some others are also making changes in unrelated files, for example 49ae68d9 not only adds tests but also removes properties from the vsstree implementation.

SebastianSchildt:
Highlevel question: Instead of the enum of allowed units (that may fall appart and needs some fixing whenever introducing a new "useful" unit, should we use a mechanism that parses all "sane" units, such as https://pint.readthedocs.io/en/0.12/   This is just arguing form a "painless tooling" point of view, because I think here is not the discussion about a whitelist of all legal  units in VSS forever... (which might more be along the lines of ISO 80000 aka SI Units)

README.md
```diff
@@ -0,0 +1,54 @@
+# VSS Tools Usage
+
+## Contents
+1. Prerequisites
+2. Project Setup
+3. Run Project
+
+## 1. Prerequisites
+* Python 3.7 installed
+* If executed behind a corporate proxy http_proxy and https_proxy environment variable set (including authentication e.g., http://${USERNAME):$(PASSWORD)@yourproxy.local)
```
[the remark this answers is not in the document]
Floix: I'll update the comment to make it more clear

README.md
```diff
@@ -0,0 +1,54 @@
...
+* If you use a custom pip installation directory, set the PYTHONPATH enviornment variable to the directory that you set in the pip.ini file.
+
+## 2. Project Setup
+* Create a VERSION file in the parent directory of project root
+* RUN  ```pip install -e .``` from the project root directory 
+* Run  ```pip install -r requirements.txt``` 
+
+## 3. Run Project
+
+Folder structure for the example below
```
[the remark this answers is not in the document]
Floix: I updated the setup config for using vss-tools as submodule checkout of vehicle_signal_specification repository.

README.md
```diff
@@ -0,0 +1,54 @@
...
+```bash
+.
+├── VERSION
+├── vehicle_signal_specification
+│   ├── spec
+│   │   │   ├── VehicleSignalSpecification.id
+│   │   └── VehicleSignalSpecification.vspec
+└── vss-tools
```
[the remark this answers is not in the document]
Floix: Updated the readme for using vss-tools as submodule of vehicle_signal_specification project.

model/constants.py
```diff
@@ -0,0 +1,131 @@
...
+    SENTENCE_CASE = "sentencecase"
+    SNAKE_CASE = "snakecase"
+    SPINAL_CASE = "spinalcase"
+    TITLE_CASE = "titlecase"
+    TRIM_CASE = "trimcase"
+    UPPER_CASE = "uppercase"
+    ALPHANUM_CASE = "alphanumcase"
+
+
+class Unit(Enum):
```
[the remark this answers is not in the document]
Floix: Direct string matching is basically what we head before. I added these enums, to have type safe (as far as python gets here) usage of Units, Types and Datatypes. This prevents us from using multiple strings for the same unit (e.g., l, liters, liter) as a check during running vss-tools would fail and allows for type safe generation of code or other artifacts. Right now you'd have to do string matching on "l", "liter", "liters", on a lower-case cased string. In addition, during development you'd always get a list of allowed units, types and datatypes.

model/constants.py
```diff
@@ -0,0 +1,131 @@
...
+            if vss_type.value == name:
+                return vss_type
+        raise Exception("Invalid VSS data type %s only support %s" % (name, VSSType.values()))
+
+    @staticmethod
+    def values():
+        return list(map(lambda v: v.value, VSSType))
+
+
+class VSSDataType(Enum):
```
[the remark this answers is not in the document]
Floix: See https://github.com/GENIVI/vss-tools/pull/7#discussion_r436282090

model/vsstree.py
```diff
@@ -0,0 +1,141 @@
...
+    def is_private(self):
+        node = self
+        while node:
+            if node.name == "Private":
+                return True
+            node = node.parent
+        return False
+
+    @staticmethod
+    def case_conversion(source, style: StringStyle):
```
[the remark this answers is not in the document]
Floix: This is an optional suggestion. I had the need to generate types out of VSS in multiple programming and definition languages. All had different conventions. In order to support that, I added StringStyle and conversion libraries that should cover all kinds of variable naming conventions. However, we can get rid of that and just allow the generator code to decide for that. It probably boils down to adding some convenience to the core framework so that generators do not need to re-invent the wheel here.

model/vsstree.py
Floix: Not all VSS leafs have a unit defined. This might make a difference for generating code or other sorts of definitions from the model. Thus, I added little helper functions like this to delegate unit support to the VSSNode instead of handling it in the generators.

model/constants.py
Floix: Not sure what the question here is, can you elaborate?

vspec.py
Floix: I did not want to change the load function from vspec.py as this change would be breaking. I'm also not familiar with all the generators, thus I thought the best way would be to introduce a second similar function (load_tree) and deprecate the old one. I'd recommend to get rid of the load function sooner or later and us the typed model tree to reduce complexity in the generators.

Floix:
> Highlevel question: Instead of the enum of allowed units (that may fall appart and needs some fixing whenever introducing a new "useful" unit, should we use a mechanism that parses all "sane" units, such as https://pint.readthedocs.io/en/0.12/ This is just arguing from a "painless tooling" point of view, because I think here is not the discussion about a whitelist of all legal units in VSS forever... (which might more be along the lines of ISO 80000 aka SI Units)

I took a look at pint. It seems that it's mostly designed for unit conversion and comparison. In our case, we have a rather static model with a defined set of metrics. The purpose of adding the Unit Constants was to make explicit which units are currently supported and what abbreviations we use. Pint would not really support that. 

Furthermore, the Unit enum helps us with code completion. Pint will only throw runtime errors if you select a unit that is not supported. We could define our own set of metrics here, however this is not simpler or less explicit as the enum right now. We'd simply replace one list with another and loose code completion. 

The current approach allows us to define supported units, name them the same way and we can easily extend it by just adding another entry to the enum.

Floix:
> Still a preview, so please don't merge yet. Improves issue described here: #5
> 
> I'd like to ask for comments (e.g., @danielwilms @gunnarx @magnusfeuer)
> 
> What's still missing:
> 
> ~~* Additional tests on nested structures~~
> ~~* Additional documentation~~
> ~~* Migration of CSV converter to demonstrate approach and guide other migrations~~
> * Versioning for proper deprecation of features

For versioning I'd need some input from you guys for deprecating the load function. In addition, I tried to address the comments. Let me know if you are satisfied with my answers.

The conversion from the old CSV generator to a version that uses the VSSNode class and anytree is a good example on how we can get rid of a lot of code in the generators.

magnusfeuer:
Started going through the install, code, and reviews right now. I will not comment on old questions about enums, etc, since I am sure that the rest of the gang has done a thorough job.

I can rewrite `vspec2c.py`, `vspec2json.py`, and `vspec2csv.py` to use VSSNode once I've understood the code.

Vspec2cnative is Ulf's code, so I will leave that to him.

I believe we agreed on deprecating `vspec2franca.py`

Floix:
@magnusfeuer vspec2csv.py is already migrated within this PR as an example implementation.

Floix:
You can start with the migration if you like. The old versions will still work though, as the current load function is still available and only marked as deprecated.

@gunnarx can you check if the changes you requested are acceptable now?

model/constants.py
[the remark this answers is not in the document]
gunnarx: If it's not clear what I meant, here is an example:
[simple string match](https://gist.github.com/gunnarx/f4e3551e69926855544ef518a4d9b0e7)

It seems to me this does the same thing when it comes to checking VSS files to fulfil a standard ("prevents us from using multiple strings for the same unit (e.g., l, liters, liter)") and is a lot less complex.  

No need to refer to _what we had before_ because I am agreeing it should not  remain exactly the same, instead discussing a way forward.

> Right now you'd have to do string matching on "l", "liter", "liters", on a lower-case cased string

No, I am not advocating that the VSS file should be able to use different spellings for the same thing.  I agree with this check in principle.

> In addition, during development you'd always get a list of allowed units, types and datatypes

When writing custom code that is referring to specific data types then yes, you will be warned if you misspell the name of the units that you explicitly listed, or get IDE support to write them correctly.  I just found the writing of such code to be a rare occasion and I am not convinced that it is worth all the boiler plate code.  But this comes down to a preference.  Whereas I often lean on the side of more formalism, I am in VSS partial to the simplicity that we have created, and keeping the amount of code to maintain to a minimum.
As you also noted, even calling this "type safe" has its limits...

model/constants.py
gunnarx: The question is why UINT is mapped to INT (with no explanation).
However, it might be related to this? 
https://github.com/GENIVI/vehicle_signal_specification/issues/163

gunnarx: ... or the question could be why you changed something at all in this file.  It seems unrelated to the main topic here?

danielwilms:
As discussed in VSS meeting @2020/07/14 PR ready to be merged.
@Floix will resolve the merge conflict. 
@gunnarx, please remove the requested changes flag. 
@magnusfeuer, @gunnarx: please merge when it's ready

Issue #15, opened 2020-08-25
name 'convert_yaml_to_list' is not defined


SebastianSchildt:
No matter waht I do, with current tools I can not generate anything, not even the CSV example

```
$ python3 ./vspec2csv.py -i:../spec/VehicleSignalSpecification.id ../spec/VehicleSignalSpecification.vspec vss.csv
Traceback (most recent call last):
  File "./vspec2csv.py", line 97, in <module>
    tree = vspec.load_tree(args[0], include_dirs)
  File "/home/sebastian/vehicle_signal_specification/vss-tools/vspec.py", line 225, in load_tree
    flat_model = load_flat_model(file_name, "", include_paths)
  File "/home/sebastian/vehicle_signal_specification/vss-tools/vspec.py", line 316, in load_flat_model
    raw_yaml = convert_yaml_to_list(raw_yaml)
NameError: name 'convert_yaml_to_list' is not defined
```

Any pointers @floix @gunnarx ?

SebastianSchildt:
@danielwilms

danielwilms:
@SebastianSchildt same here, I'll have a look.

PR #16, opened 2020-08-26
Bugfix: Add previously removed method

danielwilms:
With commit #4497a47 method convert_yaml_to_list
has been removed. Re-added it.

Fixes: #15

### DIFF
--- a/vspec.py
+++ b/vspec.py
@@ -219,6 +219,25 @@ def load(file_name, include_paths):
     deep_model = create_nested_model(absolute_path_flat_model_with_id, file_name)
     cleanup_deep_model(deep_model)
     return deep_model["children"]
+    
+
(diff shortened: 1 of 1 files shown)


SebastianSchildt:
I can confirm this is running and does indeed fix https://github.com/GENIVI/vss-tools/issues/15#

Should be merged ASAP @gunnarx @danielwilms  
While I am not sure it is "bugfree" , it does bring our tools back from "impossible to run" to "works"

gunnarx:
Review comment @danielwilms :

A better summary than "add previously removed method" is to actually name the method.   Always try to provide as much useful information (in few words) in the summary:
For example:
 "add convert_yaml_to_list method that was removed by mistake"

> Should be merged ASAP

OK sure. I'll merge and we'll continue from there.

Issue #11, opened 2020-08-11
Updating vehicle_signal_specification to use the latest vss-tools master, make fails:


gunnarx:
Attempting to update to the latest master commit of vss-tools, running make inside VSS causes the following for me:

... (successful parts removed) ...

```
cd vehicle_signal_specification
   (update submodule)
make
...
./vss-tools/vspec2json.py -i:spec/VehicleSignalSpecification.id -I ./spec ./spec/VehicleSignalSpecification.vspec vss_rel_$(cat VERSION).json
Traceback (most recent call last):
  File "./vss-tools/vspec2json.py", line 14, in <module>
    import vspec
  File "/tmp/a/vehicle_signal_specification/vss-tools/vspec.py", line 17, in <module>
    from anytree import (Resolver, ChildResolverError, LevelOrderIter)
ModuleNotFoundError: No module named 'anytree'
make: *** [Makefile:15: json] Error 1
```

Hmm... but **anytree** seems to be installed using pip already?
However, it's in /usr/**local**/lib/python3.7/site-packages

Is this a local problem only for me?  Please confirm.

Of course I could send a PR with an update and vehicle_signal_specification should run the automated build checks, but probably someone with more knowledge can comment on the above --> @Floix?  @magnusfeuer what is your experience?

==> So what is required to make it work - a code change or instructions in a README for how to install the right prerequisites?

chrizmc:
Some time ago I had a similar problem. I changed the first line of the file from

#!/usr/bin/python3

to 

#!/usr/local/bin/python3

Then it should work.

SebastianSchildt:
I will check

Floix:
> Some time ago I had a similar problem. I changed the first line of the file from
> 
> #!/usr/bin/python3
> 
> to
> 
> #!/usr/local/bin/python3
> 
> Then it should work.

Is there a better way to remove/reduce the dependency to dev machine settings? E.g., just execute python without absolute paths or a build container?

gunnarx:
> Is there a better way to remove/reduce the dependency to dev machine settings? E.g., just execute python without absolute paths or a build container?

@chrizmc, Yes I agree with @Floix.  I asked for guidance precisely because I don't want to just "make it work on my machine".   We need to decide on a strategy, with possibly a specific environment that is the supported one.  I noticed more path/version issues in the Travis environment, documented in #12

SebastianSchildt:
[...]

PR #17, opened 2020-08-26
Changed interpreter path

danielwilms:
Using env instead of direct python path
in order to make it more independent of the
python installation and underlying distribution.

Fixes: #11

### DIFF
--- a/vspec2c.py
+++ b/vspec2c.py
@@ -1,4 +1,4 @@
-#!/usr/bin/python3
+#!/usr/bin/env python3
 
 #
 # (C) 2019 Jaguar Land Rover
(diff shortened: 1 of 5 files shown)


SebastianSchildt (approved):
Not sure it "fixes" #11 for @gunnarx  , but definitely an improvement

gunnarx:
Sorry, I missed this one and added my own commit in #19
 
I would put your commit into it to be polite but you missed changing the files in the ocf/ subdirectory. 
;-) ;-)

...so if you're OK with it, we can merge #19.

danielwilms:
@gunnarx fine for me. I'll close this one.

Issue #14, opened 2020-08-25
vspec2c not supporting array datatypes


SebastianSchildt:
As discovered here https://github.com/GENIVI/vehicle_signal_specification/pull/177  the vspec2c generator does not support arrays

danielwilms:
@magnusfeuer: Could you please have a look.

danielwilms:
@danielwilms has a look

s-e-martin:
JLR moves to deprecate `vspec2c` unless somebody else needs it (see https://github.com/GENIVI/vss-tools/issues/23#issuecomment-700876515)

PR #22, opened 2020-09-15
Added array support to the tools

danielwilms:
Added arrays to the type check, so that they are included.

Fixes: #14


gunnarx:
This looks correct to me (when I consider what was decided by the group on the syntax of the arrays).  Should just be merged, but then of course the tools may need to start supporting the arrays in conversions also.

PR #24, opened 2020-09-22
Cnative array support.

UlfBj:


gunnarx:
GitHub said:
"Rebasing the commits of this branch on top of the base branch cannot be performed automatically as this would create a different result than a regular merge"
I found the same when trying to rebase locally.  I will therefore not care about cleaning up the history as we had originally planned, but simply add a merge-commit with all of the above.

PR #25, opened 2020-10-27
cnative support for instantiation.

UlfBj:


danielwilms (approved):
Works for me

gunnarx:
No issues from me.  Just note that this code has its own checking of datatypes and still accepts alternative spelling.
So PR #26 should likely be updated to ensure the change is done everywhere at the same time?

gunnarx:
I can't help but write down, since I noticed it, that `strlen` is used instead of `strnlen` on what I believe are *unchecked* input strings?  Sure, it's not strcpy, so perhaps not as bad, but it's still worth noting that a C-library, *if* it is going to be used widely, should be carefully checked and made safe against buffer overflows.

danielwilms:
remove non-related commits from PR

PR #28, opened 2020-11-23
cnative and gonative tools replaced by native tool.

UlfBj:
The native tool creates a "native" tree file, see its format in the README in the native directory. 
Two tree parsers providing a set of convenience services that a VS server is likely to use are available, one written in C, the other in Go. 
If this PR is accepted, then the Makefile in the VSS repo needs an update, which is submitted there.


UlfBj:
The conflicting files
c_native/cnativenodelib.c
vspec2cnative.py
should both be deleted as the cnative tool is removed.

gunnarx:
Rebased to get rid of GitHub complaints.  (There were no conflicts when rebasing)

SebastianSchildt:
Regarding the discussion here: https://github.com/GENIVI/vss-tools/issues/27

Personally I think, this is not the most "elegant" way to build this, as it is some kind of intermediary between loading a fully fledged YAML/JSON and compiling everything into the binary.

But as it is working, and not replacing any "better" solution and is in use, I would propose to merge this, as a working example/workflow how you could use VSS in C and Go, unless somebody commits to change it in the short-term.

As it does not state to be a super generic solution, or standard pattern to consume VSS, I think it is fine.

danielwilms (approved):

gunnarx (changes requested):
No.

magnusfeuer:
No

SebastianSchildt (approved):

chheis:
Yes (merge it)

UlfBj:
Yes to merge.

danielwilms:
@UlfBj change wording

UlfBj:
Native tool renamed to binary.

vspec2binary.py
```diff
@@ -33,16 +33,29 @@ def usage():
     sys.exit(255)
 
 import os.path
+<<<<<<< HEAD:vspec2binary.py
```
gunnarx, line 36: You have some conflict markers here that should not be in the commit.
You might just edit this file manually, keep the lines that should be there, and 
```git add vspec2binary.py
git commit --amend
```
to fix the commit.  (Or create a new commit but even with what is there now it would be worthwhile to squash the changes together before merging, IMHO)

gunnarx (changes requested):

UlfBj:
Merge conflicts fixed.

danielwilms:
to be merged 2021/01/26

gunnarx:
As requested, I am merging this.  But there was a minor conflict again after latest merges to master.  We also discussed squashing some of the uneeded merge and fix commits, so it's now done and it is only one commit.

PR #32, opened 2021-02-01
Native to binary scripts update.

UlfBj:


UlfBj:
The script files setup.py and three go.mod files under binary/go_parser had not been updated with the change from native to binary.

gunnarx:
Yes. Rebased and merged.

PR #34, opened 2021-02-09
Support deprecation in csv generation

danielwilms:
* Added deprecation to vss element
* Added deprecation to csv output
* Fixed missing UUID

Fixes: #33

PR #35, opened 2021-02-09
Feature/json to anytree

SebastianSchildt:
This ports JSON generation to the "new" anytree based parser. (It performs more checks, and the old one is deprecated)

It fixes 1 bug in anytree vspec parser: Deprecation is supported now

The created file is almost identical to compare change the line

```python
    json.dump(tree, json_out, indent=2)
```

in the old version to 

```python
    json.dump(tree, json_out, indent=2, sort_keys=True)
```

This gives the following  diff 

```diff
--- new.json    2021-02-09 12:27:34.693521102 +0100
+++ old.json    2021-02-09 12:02:03.687367155 +0100
@@ -12031,6 +12031,12 @@
         "type": "branch",
         "uuid": "12f35ec7bd1c58d1a329565ce3d053d5"
       },
+      "Private": {
+        "children": {},
+        "description": "Uncontrolled branch where non-public signals can be defined.",
+        "type": "branch",
+        "uuid": "4161866b048a5b76aa3124dec82e0260"
+      },
       "RoofLoad": {
         "datatype": "int16",
         "description": "The permitted total weight of cargo and installations (e.g. a roof rack) on top of the vehicle.",
@@ -12256,4 +12262,4 @@
     "type": "branch",
     "uuid": "ccc825f94139544dbb5f4bfd033bece6"
   }
-}
\ No newline at end of file
+}
```

Because it i seems, the anytree parser seems to ignore it. I am not quite sure why (because it generate empty "media")
Anyway, as long as you can not prove it is my fault, I consider it a separate bug :)

### DIFF
--- a/vspec2json.py
+++ b/vspec2json.py
@@ -1,7 +1,8 @@
 #!/usr/bin/env python3
 
 #
-# (C) 2016 Jaguar Land Rover
+# (c) 2016 Jaguar Land Rover
+# (c) 2021 Robert Bosch GmbH
 #
 # All files and artifacts in this repository are licensed under the
 # provisions of the license provided by the LICENSE file in this repository.
@@ -14,6 +15,8 @@ import sys
 import vspec
 import json
 import getopt
+from model.vsstree import VSSNode, VSSType
+
 
 def usage():
     print("Usage:", sys.argv[0], "[-I include_dir] ... [-i prefix:id_file] vspec_file json_file")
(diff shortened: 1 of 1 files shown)


SebastianSchildt:
@rfrowe  maybe take a look as you are a JSON consumer?
@magnusfeuer  Originally was your code I think, maybe give it a rain check...

rfrowe:
@SebastianSchildt Assuming that diff is correct, this change will be fine. I can test the newly generated file if desired

SebastianSchildt:
Yes pls test. That diff just implies the new JSON file is missing the (empty) "Vehicle.Private" branch. But that is a bug of the new parser, so will be repaired seperately.

You can generate a complete JSON by checking this branch and a current spec out and run

```
python3 vspec2json.py -i :my.id -I ../spec/ ../spec/VehicleSignalSpecification.vspec new.json
```

Or just trust me :) and use this: 
[new.json.txt](https://github.com/GENIVI/vss-tools/files/5954675/new.json.txt)

If you tooling can't smell any difference that would be a good indication, that this PR is ok.

danielwilms:
@SebastianSchildt, this is the cause:
```
return render_tree(dict_tree, True)
```
Removing the second parameter or setting it to `False` solves your issue.

https://github.com/GENIVI/vss-tools/blob/eacd7e00976ef56cdf1e482e9b6b0aad2931154d/vspec.py#L253

SebastianSchildt:
I am happy to change that, but note it is a "generic" setting, that is currently not easily reachable from outside, so it would change for all tools (e.g. CSV). I think that is the desired behaviour (do not ignore anything in your vspec), or do you disagree?

danielwilms:
Sorry, needs clarification :) before the change to anytree private was included. I think it's hard to figure out for someone using it, why the private branch disappears by default. But it can be a separate discussion as well and doesn't have to be as part of this PR.

rfrowe:
@SebastianSchildt I checked your provided JSON with my tooling. I found instances of `value` on attributes which is illegal as per https://github.com/GENIVI/vehicle_signal_specification/issues/198

rfrowe:
Aside from that it was valid

SebastianSchildt:
This is know rebased on #34 and should be merged after

The problem with the extra  values is a spec bug and is tracked here https://github.com/GENIVI/vehicle_signal_specification/pull/237

SebastianSchildt:
Will need to rebase this correctly, then mark as mergeable again

danielwilms (approved):

gunnarx:
Fixed some whitespace errors and amended, but I could not push back to branch boschglobal:feature/json_to_anytree so I merged direct to master.

PR #37, opened 2021-02-09
FIx CSV build failure (vsstree.py: Add deprecation key)

gunnarx:
The CSV converter was failing because the new deprecation: key was
unknown.  

Other converters also use vsstree definition, so I think this affects more
things, but it is noticed when Travis fails for CSV.


gunnarx:
Closed in favor of #34

PR #40, opened 2021-02-12
Go parser go.mod updates.

UlfBj:


danielwilms:
@gunnarx pls have a look

gunnarx:
Lots of commits, merges and a conflict with master => I rebased on master and squashed everything into one commit and updated the PR.
@UlfBj Please check if 6cd6f71 contains everything we need and if it is correct.

gunnarx:
Hmm, it does not seem quite correct to put this into only one commit.  6cd6f71 contains more than go.mod updates.  I think you had also other "bug fixes" in this PR.

UlfBj:
Yes, it slipped into the previous PR. 
The initial parts were about go script updates, the latter about bug fixes in the parser code.
Shall I do anything to brake it up into two?

gunnarx:
OK, I could not easily separate different parts (and not sure what is the right separation either).  So, because the changes include both go.mod updates, and several code updates, I now just renamed the squashed commit to "Updates to binary tool ".  Not ideal but the best I could do.  If the latest commit looks OK I propose we merge it.

UlfBj:
Ok. Thanks.

danielwilms:
@danielwilms run code and check PR

danielwilms:
decision to merge 09/03/21

PR #41, opened 2021-02-18
vspec2proto PoC

SebastianSchildt:
protobuf generator from hell.

But it does work. 

```
$ python3 vspec2proto.py   -i :test.id -I ../vehicle_signal_specification/spec/ ../spec/VehicleSignalSpecification.vspec    vsstest.proto
$ protoc -I=. --cpp_out=. vsstest.proto 
```

It is a nested message structure, beware when loading the generated cpp and h in your IDE, might eat some GB in VSCode :grin:

### DIFF
--- a/vspec2proto.py
+++ b/vspec2proto.py
@@ -1,0 +1,135 @@
+#!/usr/bin/env python3
+
+#
+# (c) 2016 Jaguar Land Rover
+# (c) 2021 Robert Bosch GmbH
+#
+# All files and artifacts in this repository are licensed under the
+# provisions of the license provided by the LICENSE file in this repository.
+#
+#
+# Concert vspec file to JSON
+#
+
+import sys
+import vspec
+import json
+import getopt
+import re
+
+from model.vsstree import VSSNode, VSSType
+
+
+def usage():
+    print("Usage:", sys.argv[0], "[-I include_dir] ... [-i prefix:id_file] vspec_file proto_file")
+    print("  -I include_dir       Add include directory to search for included vspec")
+    print("                       files. Can be used multiple timees.")
+    print()
+    print("  -i prefix:uuid_file  File to use for storing generated UUIDs for signals with")
+    print("                       a given path prefix. Can be used multiple times to store")
+    print("                       UUIDs for signal sub-trees in different files.")
+    print()
+    print(" vspec_file            The vehicle specification file to parse.")
+    print(" json_file             The file to output the JSON objects to.")
+    sys.exit(255)
+
+
+
+def getDataType(node):
+    um =  re.compile('uint.*')
+    sm =  re.compile('int.*')
+    m64 = re.compile('.int64')
+
+    if str(node.data_type.value) == "boolean":
+        return "bool"
+    
+    elif m64.match(str(node.data_type.value)) != None:
+        return str(node.data_type.value)
+        
+    elif um.match(str(node.data_type.value)) != None:
+        return "uint32"
+
+    elif sm.match(str(node.data_type.value)) != None:
+        return "int32"
+           
+    elif str(node.data_type.value) == "float":
+        return "float"
+    elif str(node.data_type.value) == "double":
+        return "double"
+    elif str(node.data_type.value) == "string":
+        return "string"
+    elif str(node.data_type.value) == "string[]": #probably need more generic array handling
+        return "repeated string"
+    else:
+        print("Unsupported datatype "+str(node.data_type.value))
+        return None
+
+
+
+def createMessage(file, prefix, node):
+    dt = getDataType(node)
+    if dt is None:
+        print("Skipping "+node.name)
+        return
+
+    file.write("message "+prefix+" {\n")
+    file.write("    "+dt+" value = 1;\n")
+    file.write("}\n\n")
+    
+
+def export_node(file, prefix, node):
+    if node.type == VSSType.SENSOR or node.type == VSSType.ATTRIBUTE:
+        createMessage(file, prefix, node)
+    elif node.type == VSSType.BRANCH:
+        file.write("message "+prefix+" {\n")
+        for child in node.children:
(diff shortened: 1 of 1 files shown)


vspec2proto.py
```diff
@@ -0,0 +1,135 @@
...
+    print("Usage:", sys.argv[0], "[-I include_dir] ... [-i prefix:id_file] vspec_file proto_file")
+    print("  -I include_dir       Add include directory to search for included vspec")
+    print("                       files. Can be used multiple timees.")
+    print()
+    print("  -i prefix:uuid_file  File to use for storing generated UUIDs for signals with")
+    print("                       a given path prefix. Can be used multiple times to store")
+    print("                       UUIDs for signal sub-trees in different files.")
+    print()
+    print(" vspec_file            The vehicle specification file to parse.")
+    print(" json_file             The file to output the JSON objects to.")
```
erikbosch, line 33: Should be "proto_file", not "json_file"

vspec2proto.py
```diff
@@ -0,0 +1,135 @@
...
+
+            [prefix, file_name] = id_spec
+            vspec.db_mgr.create_signal_uuid_db(prefix, file_name)
+        else:
+            usage()
+
+    if len(args) != 2:
+        usage()
+
+    json_out = open (args[1], "w")
```
erikbosch, line 124: proto_out?

AnotherDaniel:
So I was too slow for both writing code, and doing a review - but for what it's worth I look at it as well, tested it locally, and approve :)

tobiask84:
It looks to me like this will nest the resulting structure.

The [docs suggest here](https://developers.google.com/protocol-buffers/docs/proto3#nested) that you can access a nested structure like this:

`If you want to reuse this message type outside its parent message type, you refer to it as _Parent_._Type_`

Additionally the names seem to be prefixed here. Wouldn't that mean that you now have to enter ever prefix multiple times and you would have very long and overly complicated names for accessing attributes?

To me it seems better to not nest the resulting structure.

I already implemented a version that is not nested and has some other improvements.

tobiask84:
For example take this here which I just generated:
`Vehicle.Vehicle_Body.Vehicle_Body_Windshield.Vehicle_Body_Windshield_Rear.Vehicle_Body_Windshield_Rear_WasherFluid.Vehicle_Body_Windshield_Rear_WasherFluid_LevelLow.value`

vspec2proto.py
```diff
@@ -0,0 +1,135 @@
...
+        return "string"
+    elif str(node.data_type.value) == "string[]": #probably need more generic array handling
+        return "repeated string"
+    else:
+        print("Unsupported datatype "+str(node.data_type.value))
+        return None
+
+
+
+def createMessage(file, prefix, node):
```
tobiask84, line 69: This method body can start with:
`file.write("message "+prefix+" {\n")`
and end with:
`file.write("}\n\n")`

This will be written to the file either way. Currently both of these write statements exist in the if clause (in createMessage) and in the else clause of this method. By putting it here, outside the if clause, it will only be needed once.

tobiask84:
Another thing that I don't like about this solution here is that it nests the lowest level (ie the leaves of the tree) into separate messages. That creates inefficient data structures.

For example created with this solution:
```
message Vehicle_Cabin_Seat_Row1_Pos4_Occupant {
message Vehicle_Cabin_Seat_Row1_Pos4_Occupant_Identifier {
message Vehicle_Cabin_Seat_Row1_Pos4_Occupant_Identifier_Subject {
    string value = 1;
}

message Vehicle_Cabin_Seat_Row1_Pos4_Occupant_Identifier_Issuer {
    string value = 1;
}

}

}
message Vehicle_Cabin_Seat_Row1_Pos4_IsBelted {
    bool value = 1;
}

message Vehicle_Cabin_Seat_Row1_Pos4_Heating {
    int32 value = 1;
}

message Vehicle_Cabin_Seat_Row1_Pos4_Massage {
    uint32 value = 1;
}

message Vehicle_Cabin_Seat_Row1_Pos4_Recline {
    int32 value = 1;
}

message Vehicle_Cabin_Seat_Row1_Pos4_Position {
    uint32 value = 1;
}
....
```

How it would be a lot better:

```
message VehicleCabinSeatRow3Pos3 {
  bool HasPassenger = 1;
  VehicleCabinSeatRow3Pos3Occupant Occupant = 2;
  bool IsBelted = 3;
  int32 Heating = 4;
  uint32 Massage = 5;
  int32 Recline = 6;
  uint32 Position = 7;
  VehicleCabinSeatRow3Pos3Cushion Cushion = 8;
  VehicleCabinSeatRow3Pos3Lumbar Lumbar = 9;
  VehicleCabinSeatRow3Pos3SideBolster SideBolster = 10;
  VehicleCabinSeatRow3Pos3HeadRestraint HeadRestraint = 11;
  VehicleCabinSeatRow3Pos3Airbag Airbag = 12;
  VehicleCabinSeatRow3Pos3Switch Switch = 13;
}
```

tobiask84:
I created a less bloated version without these problems here:
https://github.com/GENIVI/vss-tools/pull/42

But it seems to me that enums and other things will need to be implemented as well. But if we add a POC we might as well take a better one.

SebastianSchildt:
I suggest we continue with @tobiask84 PR #42 , while I also may repair details in this one, I have no real need for this atm., so lets continue with the better one 😁

gunnarx:
> Another thing that I don't like about this solution here is that it nests the lowest level (ie the leaves of the tree) into separate messages. That creates inefficient data structures.

I have a similar challenge with converter to Franca IDL which is also a full featured IDL.  My conclusion is that the converter needs some parameters to control the different ways you can translate branches, either into nested namespaces / structures (maybe multiple levels) or a more "flat" approach.  The converter might be able to do it in different ways.   (Conventions/recommendations for how to do it in a particular case might follow separately).

danielwilms:
@gunnarx: close this one, continue in #42

PR #39, opened 2021-02-11
Feature/parser check core attributes

SebastianSchildt:
This fixes #38 

merge order #34  #35  and then this

This does not change the default behaviour (i.e. CSV parser not changed, output the same), but will print warnings if non-core attribtes are found.

With current vspec

```
$ python3 vspec2json.py  -i :test.id -I ../vehicle_signal_specification/spec/ ../spec/VehicleSignalSpecification.vspec    vsstest.json
Loading vspec...
Warning: Non-core attribute "value" in elment CurrentOverallWeight found.
Warning: Non-core attribute "value" in elment CurbWeight found.
Warning: Non-core attribute "value" in elment GrossWeight found.
Warning: Non-core attribute "value" in elment MaxTowWeight found.
Warning: Non-core attribute "value" in elment MaxTowBallWeight found.
Warning: Non-core attribute "value" in elment Length found.
Warning: Non-core attribute "value" in elment Height found.
Warning: Non-core attribute "value" in elment Width found.
Warning: Non-core attribute "value" in elment ChargePortFlap found.
Warning: Non-core attribute "value" in elment ChargePlugType found.
Warning: Non-core attribute "value" in elment Mode found.
Warning: Non-core attribute "value" in elment StartStopCharging found.
Warning: Non-core attribute "value" in elment Mode found.
Recursing tree and creating JSON...
All done.
```
(for layers non-core attributes might be desired)

Once all spec bugs are fixed, we can switch Travis to run strict mode with `-s` parameter

```
python3 vspec2json.py -s  -i :test.id -I ../vehicle_signal_specification/spec/ ../spec/VehicleSignalSpecification.vspec    vsstest.json
Loading vspec...
Warning: Non-core attribute "value" in elment CurrentOverallWeight found.
You asked for strict checking. Terminating.
```
@danielwilms  pls check

PR #43, opened 2021-03-03
Fixing unit and datatypes tests

erikbosch:
I noticed when running "pytests test" according to the README that there were some errors due to previously changed units/types. 
This PR fixes that.

But how come that it has not been detected. Shall we consider these tests as obsolete and remove them?
Or possibly instead include them in Travis-tests so the automated tests will fail on errors?


SebastianSchildt:
Never used/seen them.

Since they are there, I suggest in this PR also adding them to travis.yml, so if in the future it breaks, we can check why.

erikbosch:
Sure, no problem, but what do we consider as the preferred way.

- Adding it directly to travis.yml, like "PYTHONPATH=. pytest"
- Or adding it to the Makefile in vehicle-signal-specification like other tools called indirectly from the Makefile, so that it will be included if you run "make all"?

erikbosch:
Note: This PR depends on that make target "tests" have been merged, see https://github.com/GENIVI/vehicle_signal_specification/pull/249

SebastianSchildt:
I vote merging this, once we get https://github.com/GENIVI/vehicle_signal_specification/pull/249   merged.

In the long term, build recipes/vss-tools helper scripts should move to vss-tools, but lets discuss this separately from the issue here.

This is about automatically running the tests we have, and I believe not test should go to waste :)

gunnarx:
It failed again.  Looks like it is about the **value** metadata.  Was it recently changed to **default**?  @erikbosch can you check?

erikbosch:
Now it worked @gunnarx . It seems that I didn't rebase my own vss-tools to latest master when I tested myself , so having a test on "value" still worked locally. But that was changed about 14 days ago by #39 so I had to remove the test on "value" here as well, so now the tests pass.

PR #44, opened 2021-03-10
Move ocf related tools and create contrib/ dir

gunnarx:
Introducing contrib/ dir as was proposed on a recent VSS meeting.

I looked around the internet and a typical definition of contrib is either
- code contributed by external parties and not necessarily maintained by the core project maintainers
- code that is not really part of the core project or "product" as such (not officially part of debian distribution, in the case of debian-contrib), but that is supported by the core project, and works well together with the project.

This is roughly what we are looking for here.  Fulfilling all or some of these points:

- Code is not guaranteed to be maintained at all times (for example tools in contrib/ might not be included in the automated tests used as acceptance criteria for pull requests to VSS, i.e. they might sometimes break)
- Supported but not strongly endorsed by the project  (for example, the solution or format might still be under discussion), or quite simply work-in-progress that we still want to make visible.
- Code libraries, converters, etc. are not clearly adopted or used (yet) by other projects.

As a first example, the OCF tooling has not been used or updated for a long time, so it should not be in the primary directory at the moment.  Do you agree?


erikbosch:
To know if a tool is "compliant with the current version" the first step is to make sure that it can be tested to some degree by our Continuous Integration-environment  (E.g. added to https://github.com/GENIVI/vehicle_signal_specification/blob/master/Makefile and/or https://github.com/GENIVI/vss-tools/blob/master/.travis.yml) 

As of now neither vspec2c nor ocf are part of our travis tests, making them candidates for "contrib" as I see it.

gunnarx:
Thanks for feedback, trying to evaluate and answer here:
(quotes in opposite order because it worked better)

@adobekan wrote:

> I would go for, if a tool is complaint with current version and if issues are being resolved?

This is roughly equal to the first bullet point, in the second section

>  I am not sure if one the main classifiers should be "if something is used or not".

I wasn't being overly formal and deliberately wrote it fulfils "all or some" so not sure if each of the three bullets should be considered a "main" classifier.  

So I see your argument that the first one (third bullet in my list) is a weaker point, and consider removing it, but then I think I change my mind back again.  Here is what I was thinking when I wrote it:  We should generally strive for making sure that all code is updated, runs successfully, etc.  So there _might_ be tools that we manage to keep up to date, but are still not so widely used.  So in your view does that mean they should **_not_** be moved to contrib?  I don't think contrib is necessarily a location only for outdated non-maintained code.  The example I took with debian and other explanations of contrib do not suggest that the code is incompatible.   Do we not agree that there could be other reasons to put code in contrib/ ?   
What would be your list of conditions?

The file that started the contrib proposal was protobuf and it fits this, in my opinion.  It appears to be a complete tool, it is up to date (using the anytree methodology), I suppose it runs successfully and if it were included in the CI test it would not fail the test.  But at the same time I had concerns about if it is generating the right thing, and I propose to wait for programs to start using its output (this way we will automatically get the answer if it is generating "the right thing" or it will change).  This triggered the proposal to move it to contrib/

So I'm leaning back towards also keeping _also_ "things that are less used" in contrib.  It's a case-by-case evaluation though.

> Otherwise, I would link the tool to last working version and deprecate it from the current working branch. 

Trying to follow what you mean by otherwise.  Which scenario are you speaking about exactly?

How do you link a tool in git... if for example the implementation of vspec2foo is including vspec.py but vspec.py has changed so that vspec2foo.py is broken on the working branch.

When you say deprecate from working branch do you mean to remove it in the working (in our case master) branch?  In that scenario, what is the use of the contrib for?   contrib/ could also be used as a mild way of "deprecating" as far as I can tell.

> Additionally, open topics that are still in discussion or not yet finalized we could add to directory experimental/contrib.

But also this is roughly covered in existing bullet points, is it not?  

I guessed here that you mean "experimental/" **_or_** "contrib/"  (not a directory contrib/ inside of a directory experimental/).  The PR is a concrete proposal, so please provide a counter proposal if I misunderstood.

Honestly I'm just responding to feedback here as good practice.   I'd much prefer not to overengineer this with too many directories and conditions and spend too much time on a triviality!

Shall we continue this discussion to get to a formal list of conditions or are we OK with a rough one?  Actually, in the end what is written in the commit message is the most important.  The discussion in the PR tends to fade. What do all of you think, is it worth optimizing this list of conditions?

If you have the time, feel free to summarize, rewrite and propose a new list of conditions.  

>  https://github.com/git/git/tree/master/contrib

Yes...I spent the time to read it.  Does it describe a significantly different view from what was already covered?

gunnarx:
@erikbosch - Yes, the automated test condition covered in the first bullet of my 3-point list.
Agree, vspec2c could maybe also move to contrib.  It should be a separate commit so it's not holding this up.

adobekan:
@gunnarx 
To make it a bit more clear. My comment was related to, if the tooling is broken then deprecate it if no-one is reacting to the open issue. Why, moving it to contrib. Nevertheless, it seems that you had quite some things covered in the comment and that it was discussed in the meeting, my bad I was missing context.
Also, since ocf will be moved to new directory. The tool is looking for vspec2ocf.csv and vspecocfmap.csv in directory tools/ocf. Should we handle this change in the current pull request or you prefer to have separated pull request?

 @erikbosch, would be great!

gunnarx:
@adobekan Let's do a separate pull request.  Happy to get a quick fix for the ocf script if someone wants to make it, even if interest is relatively low on this tool at the moment.

gunnarx:
> My comment was related to, if the tooling is broken then deprecate it if no-one is reacting to the open issue. Why, moving it to contrib.

Makes sense.  At some point some things are no longer needed and might need to get removed.  I think it's something we can handle on a case by case basis.

gunnarx:
I decided to fix the path issue.  Also found that the program was failing on printfs.  Needed to be fixed to be working with python3.

SebastianSchildt:
Discussion March 23rd: Merge this @gunnarx

gunnarx:
There is still a bug (it is depending on an old vspec.py feature that is now missing).  I'll merge for now as agreed so we can move on.

PR #45, opened 2021-03-11
Go parser functional update, and go script fix

UlfBj:


SebastianSchildt:
Discussion March 23rd: "Content" is fine, and can be merged.  It seems the PR is still build in such a way, that it is conflicted.

@UlfBj  repairs @gunnarx  supports

gunnarx:
Fresh start approach:

"genivi" is whatever the name of the GENIVI upstream remote is.  
Check your remotes with:
```
git remote -v
```
"Fresh start" approach from within your working dir.
```bash
git fetch genivi
git reset genivi/master
```
**or** make a fresh checkout of course.

But the above will keep your local changes.  `reset` will move your current branch to point to genivi/master instead.  BUT since it is not a reset --hard, the difference between the files you had and the files in genivi/master will be kept in the working directory.  So if you are lucky you already have everything you need - check `git diff`.  If you do a fresh checkout you will need to copy in the right files and not forget one.

Then, with either of the above approaches:
- Edit and commit the changes you want to make in one or a few commits on top of current master.

After committing, push to your remote.  Use 'origin' if that is your own UlfBj remote:
```bash
git push origin HEAD:go_parser_functional_update
```
The syntax "HEAD:branch_name" will always push what you have committed to a remote branch using that name.  So it does not matter on which local branch you are, or if you have a detached HEAD (no branch), it will push the top of your made commits  (HEAD) to the branch name.   

Use a branch name because it is better to send a PR from a new branch instead of your Github master.

+ Create the PR inside  GitHub from your new branch.


--------------
Once you are on track again, you should try a general rebase workflow:

Do this before starting (or regularly during) any ongoing work:

"genivi" is whatever the name of the GENIVI upstream remote is.  
```bash
git fetch genivi
git rebase genivi/master
```
It is not very obvious but git rebase <something> means rebase "on top of " <something> (and leave the result in your current working branch)

More explicitly:   "Merges" any unique commits that you have made in my current HEAD (current branch) but that are missing in <something> with anything that is in <something> but is missing in your branch.  HOWEVER, it does not create a merge commit -- instead it recalculates all your changes and places them linearly "on top of" <something> (just as if they were performed last, after any other changes).  This workflow creates a clean, linear, git history.

If you get conflicts when rebasing then you would have the same conflicts using merge.
Simply resolve them this way:

- Edit the conflicting files, remove conflict markers and leave what you want the combined result to be.
- `git add <conflicting file(s)>`
then:
  `git rebase --continue`
Repeat this until the rebase is done.

If you get stuck with an unfinished rebase you can always:
`git rebase --abort`
and everything will wind back to what you had before.

PR #47, opened 2021-03-31
Align travis file with counterpart in vss

erikbosch:
With this you only need to change Makefile in vehicle_signal_specification to change the targets run by travis.

This is the same change as done in .travis.yml in vehicle_signal_specification recently


danielwilms:
@gunnarx please merge

PR #49, opened 2021-04-07
Moving vspec2c folder to contrib

erikbosch:
This is a follow up to latest commit from Gunnar


erikbosch:
@gunnarx Should not this folder also be moved to contrib, as you moved vspec2c.py?

Note: Build fails as the makefile in vss tried to do clean in the old location, so a dependent change is needed there as well

.travis.yml
```diff
@@ -4,6 +4,7 @@ python: 3.8
 
 install:
   - pip install -r requirements.txt
+  - python3 setup.py -q install
```
erikbosch, line 7: Adding same setup as in vss, with that we detect any problems with setup also here

setup.py
```diff
@@ -18,6 +18,6 @@
     url='https://github.com/GENIVI/vss-tools',
     license='Mozilla Public License v2',
     py_modules=['vspec'],
-    scripts=['vspec2csv.py', 'vspec2franca.py', 'vspec2binary.py', 'vspec2json.py', 'vspec2c.py'],
+    scripts=['vspec2csv.py', 'vspec2franca.py', 'vspec2binary.py', 'vspec2json.py', 'contrib/vspec2c.py', 'contrib/vspec2protobuf.py', 'contrib/ocf/vspec2ocf.py'],
```
erikbosch, line 21: Path change to vspec2c.py needed

erikbosch:
This PR shall likely pass Travis builds as soon as https://github.com/GENIVI/vehicle_signal_specification/pull/263 has been merged

SebastianSchildt (approved):

danielwilms:
merge, as decided in 2021/04/13

PR #51, opened 2021-04-13
Fix module searchpath

### LINKED ISSUE #50: Protobuf broken due to move to contrib
I think the move to contrib killed protobuf 

```
make protobuf
./vss-tools/contrib/vspec2protobuf.py -i:spec/VehicleSignalSpecification.id -I ./spec ./spec/VehicleSignalSpecification.vspec vss_rel_$(cat VERSION).proto
  File "./vss-tools/contrib/vspec2protobuf.py", line 13, in <module>
    import vspec
ModuleNotFoundError: No module named 'vspec'
make: *** [Makefile:37: protobuf] Error 1
```

I think we need to change the search path for python includes to include ../ , anyway, just writing this issue for ow, so we do not forget

SebastianSchildt:
Fixes #50 

(I think this is better than setting environments in Makefile as it would be surprising to a user if it only runs correctly when starting from a specific Makefile)

### DIFF
--- a/contrib/vspec2protobuf.py
+++ b/contrib/vspec2protobuf.py
@@ -10,6 +10,11 @@
 #
 
 import sys
+import os
+#Add path to main py vspec  parser
+myDir= os.path.dirname(os.path.realpath(__file__))
+sys.path.append(os.path.join(myDir, ".."))
+
 import vspec
 import getopt
 from anytree import RenderTree, PreOrderIter


gunnarx:
Looks like a good change.  IMHO the explanation you gave in the PR could be useful in the commit message.  Also the short message could say what you actually did.  ("Fix something" is usually not such good information) For example:

`Makefile: Module search path use current and parent dir`

danielwilms (approved):

gunnarx:
Amended with comment as above, but I could not push back to PR (no permission)
-> so it's merged outside of GitHub's knowledge and this will get status "Closed" instead.

PR #52, opened 2021-04-14
Adding gitignore for vspec2c

erikbosch:
To avoid the risk that someone merge them by mistake


erikbosch:
Note - fails as the updated Makefile (in https://github.com/GENIVI/vehicle_signal_specification/pull/260 ) has not yet been merged

gunnarx:
We should be able to use *.a and *.o shouldn't we?

erikbosch:
@gunnarx And now also the build for this trivial change pass (failed before)

PR #55, opened 2021-04-21
Fixing instantiation problems

### LINKED ISSUE #54: Instance handling problems in vss-tools
This is a follow up to https://github.com/GENIVI/vehicle_signal_specification/issues/265

As of now there seems to be 3 issues/limitations that better shall be fixed.

- If you have a branch `A` and then a signal `AB` then it is treated as `A.B` by tooling. See e.g. how `AxleCount` gets represented as `Vehicle.Chassis.Axle.Row1Count` and `Vehicle.Chassis.Axle.Row2Count`
- Currently signals within instantiated branch needs to be defined directly after the branch to be correctly generated, otherwise they will be ignored or not correctly generated. (Not a problem in current VSS model)
- Currently instantiations cannot be redefined, to e.g. change number of axles or number of seat rows compared to default.

A PR to be address this is being prepared

erikbosch:
Fixes #54

- Now a signal "XXXYYY" is not considered being part of branch "XXX"
- Unrelated signals can now exist between branch and signals
- Redefinition of instantiation is now possible

### DIFF
--- a/vspec.py
+++ b/vspec.py
@@ -1,4 +1,5 @@
 #
+# (C) 2021 Robert Bosch GmbH
 # (C) 2018 Volvo Cars
 # (C) 2016 Jaguar Land Rover
 #
@@ -464,11 +465,6 @@ def expand_includes(flat_model, prefix, include_paths):
 def expand_instances(flat_model):
     # create instances from specification
 
-    new_flat_model = []
-    instantiation = []
-    base_name = ''
-    instances = None
-    inst_path = ''
     cont = True
 
     def extend_entry(new_entry, instance, name):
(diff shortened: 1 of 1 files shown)


erikbosch:
With this change the method shall be more robust. See #54 for test cases covered. There is however a small performance penalty involved, additional iterations are required compared to previous solution. Due to the limited number of "instances" in VSS the impact is however very small.

SebastianSchildt (approved):
I ran it (worked fine) and checked the JSON output.  JSON looks and works fine.

danielwilms:
merge for now 18/05/21

PR #57, opened 2021-05-11
Adding tool support for ttl/VSSo

danielwilms:


danielwilms:
> Do you intend to add it as a "ttl" target in the vss makefile, and possibly include it in "travis_optional", so we later will detect if it breaks after a vss change. I tried to run it locally but apparently I missed python rdflib. Do we possibly need to add rdflib to requirements.txt file?

makes sense. I'll add the missing requirements to the txt file.

contrib/vspec2ttl/vspec2ttl.py
```diff
@@ -0,0 +1,320 @@
...
+# All files and artifacts in this repository are licensed under the
+# provisions of the license provided by the LICENSE file in this repository.
+#
+
+#
+# Convert vspec file to TTL
+#
+import os
+import sys
+#Add path to main py vspec  parser
```
gunnarx, line 18: \# (space)
and what is main py vspec <space> parser?   I suppose you mean `vspec.py`.

... and I wonder if there is a better way to do this for the contrib tools.  It seems a bit of a hack to have the program depend on where it is located in the tree.

contrib/vspec2ttl/vspec2ttl.py
```diff
@@ -0,0 +1,320 @@
...
+
+from model.vsstree import VSSNode, VSSType
+
+from rdflib import Graph, Literal, RDF, URIRef, Namespace
+from rdflib.namespace import RDFS, XSD, OWL, XMLNS, RDF, FOAF, SKOS, SDO, NamespaceManager
+from enum import Enum
+
+from vssotypes import VssoConcepts, DataTypes, DataUnits, Namespaces
+
+
```
gunnarx, line 35: Why so many empty lines?

contrib/vspec2ttl/vspec2ttl.py
```diff
@@ -0,0 +1,320 @@
...
+    g.add((partOf,RDFS.subPropertyOf, OWL.topObjectProperty))
+    g.add((partOf,RDFS.domain, VssoConcepts.VEHICLE_COMP.uri))
+    g.add((partOf,RDFS.label, Literal("partOf",lang="en")))
+
+    hasComponentInstance = VssoConcepts.HAS_COMP_INST.uri
+    g.add((hasComponentInstance, RDF.type, OWL.DatatypeProperty))
+    g.add((hasComponentInstance,RDFS.subPropertyOf, OWL.topDatatypeProperty))
+    g.add((hasComponentInstance,RDFS.label, Literal(VssoConcepts.HAS_COMP_INST.value,lang="en")))
+
+
```
gunnarx, line 92: Why extra lines?

gunnarx:
Minor feedback on this, nothing major.   Fix them if you can, but basically it works for me.

contrib/vspec2ttl/vspec2ttl.py
erikbosch: For some other tools that has been moved to contrib we solved it in the Makefile by setting PYTHONPATH before calling the script.


```
c:
	(cd ${TOOLSDIR}/contrib/vspec2c/; make )
	PYTHONPATH=${TOOLSDIR} ${TOOLSDIR}/contrib/vspec2c.py -i:./spec/VehicleSignalSpecification.id -I ./spec ./spec/VehicleSignalSpecification.vspec vss_rel_$$(cat VERSION).h vss_rel_$$(cat VERSION)_macro.h
```

felix-loesch:
Anees and I just did a review of the pull request and we encountered the following open points:
- Line 239: I think we agreed on treating concrete VSS branches as sub-classes of the class "VehicleComponent". In Line 239 you are still creating instances instead of creating owl:Class and sub-classing these from VehicleComponent. 
- When creating new classes you create them both as owl:Class and rdfs:Class (see e.g. Line 256, 257, 267, 268). We think that it will be sufficient to just use owl:Class
- Line 269: Is a vehicle signal always an observable signal? In Line 272 the subclassing is added if it is a actuatable signal. This means that a signal can either be Observable or Observable&Actuatable. Is this always valid, meaning that ActuatableSignal cannot be used without Observable?
- Line 245: You are using BELONGS_TO which is a high-level object property relating StaticVehicleProperty or DynamicVehiclePropety to VehicleComponent. If a sub-class of StaticVehicleProperty or DynamicVehicleProperty shall be linked to a sub-class of VehicleComponent, we need to introduce sub-properties of the BELONGS_TO object property. This can be achieved by suffixing the object-propert definition with the name of sub-class of VehicleComponent. 
- Line 248ff: You still try to link data types and units to sub-classes of StaticVehicleProperty or DynamicVehicleProperty. However, this is not allowed in OWL. Solution Option A: create OWL restrictions to store that information at the class level or solution option B: create master instances to store that information.

contrib/vspec2ttl/vspec2ttl.py
SebastianSchildt: I think putting it in the python file is the correct way to do this, otherwise the tool is not usable without secret knowledge of some ENV variables that need to be set. As it is expected you check out the whole tools-repo it does not depend on where you locate it.

But that is just an opinion, I am fine either way 👍

danielwilms:
> When creating new classes you create them both as owl:Class and rdfs:Class (see e.g. Line 256, 257, 267, 268). We think that it will be sufficient to just use owl:Class

fixed

danielwilms:
> Line 269: Is a vehicle signal always an observable signal? In Line 272 the subclassing is added if it is a actuatable signal. This means that a signal can either be Observable or Observable&Actuatable. Is this always valid, meaning that ActuatableSignal cannot be used without Observable?

this is the definition in VSS, so we kept the concept in the ontology

danielwilms:
> Line 248ff: You still try to link data types and units to sub-classes of StaticVehicleProperty or DynamicVehicleProperty. However, this is not allowed in OWL. Solution Option A: create OWL restrictions to store that information at the class level or solution option B: create master instances to store that information.

Removed range includes, added issue to solve it properly:
https://github.com/w3c/vsso/issues/13

danielwilms:
> Line 245: You are using BELONGS_TO which is a high-level object property relating StaticVehicleProperty or DynamicVehiclePropety to VehicleComponent. If a sub-class of StaticVehicleProperty or DynamicVehicleProperty shall be linked to a sub-class of VehicleComponent, we need to introduce sub-properties of the BELONGS_TO object property. This can be achieved by suffixing the object-propert definition with the name of sub-class of VehicleComponent.

right now it's an annotation property, as I hadn't introduced the issues with the subclasses yet. But do I understand you right, @felix-loesch, that you want to have sub-properties in order to express the connection to the right vehicle component on ontology level?

danielwilms:
> Line 239: I think we agreed on treating concrete VSS branches as sub-classes of the class "VehicleComponent". In Line 239 you are still creating instances instead of creating owl:Class and sub-classing these from VehicleComponent.

fixed

contrib/vspec2ttl/vspec2ttl.py
danielwilms: I'd vote for the python file as well.

danielwilms:
@felix-loesch, could you please have look and give feedback

danielwilms:
can be merged: meeting 15/06/21

PR #58, opened 2021-05-14
Adding framework for non-supported core attributes

erikbosch:
Previously "arraysize" was reported as a non-core attribute.
It is however listed as an allowed attribute in our documentation.
With this change it is instead csv/json/protobuf tooling
that reports that they do not support arraysize.

Fixes https://github.com/GENIVI/vehicle_signal_specification/issues/275

Below is an example of how it would like if you would try to use both a non-core attribute as well as the core attribute "arraysize" and try to generate csv:

```
# make csv
./vss-tools/vspec2csv.py -i:spec/VehicleSignalSpecification.id -I ./spec ./spec/VehicleSignalSpecification.vspec vss_rel_$(cat VERSION).csv
Warning: Non-core attribute "asjhasdjh" in element SeatRowCount found.
Warning: Core attribute "arraysize" used in element SeatPosCount is not supported in this context.
```
It is the tool (in the case vspec2csv) that needs to specify that it does not support "arraysize". If you run the tool in strict mode both warnings will be treated as errors.

### DIFF
--- a/model/vsstree.py
+++ b/model/vsstree.py
@@ -1,7 +1,7 @@
 #!/usr/bin/env python3
 
 #
-#
+# (C) 2021 Robert Bosch GmbH
 #
 # All files and artifacts in this repository are licensed under the
 # provisions of the license provided by the LICENSE file in this repository.
@@ -43,7 +43,8 @@ class VSSNode(Node):
     instances = None
     deprecation = ""
 
-    def __init__(self, name, source_dict: dict, parent=None, children=None, break_on_noncore_attribute=False):
+    def __init__(self, name, source_dict: dict, parent=None, children=None, break_on_noncore_attribute=False,
+                 unsupported_vss_attributes=[]):
         """Creates an VSS Node object from parsed yaml instance represented as a dict.
 
             Args:
@@ -59,7 +60,7 @@ class VSSNode(Node):
 
         super().__init__(name, parent, children)
         try:
-            VSSNode.validate_vss_element(source_dict, name)
+            VSSNode.validate_vss_element(source_dict, name, unsupported_vss_attributes)
         except NonCoreAttributeException as e:
             print("Warning: {}".format(e))
             if break_on_noncore_attribute:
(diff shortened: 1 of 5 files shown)


model/vsstree.py
```diff
@@ -1,7 +1,7 @@
 #!/usr/bin/env python3
 
 #
-#
+# (C) 2021 Robert Bosch GmbH
```
gunnarx, line 4: Correct to add this when modifying code.

In addition, it looks like Felix Willnecker committed most of this new vsstree.py module when it was rewritten.  Should this be complemented by adding BMW's copyright line? (@danielwilms ?)

Note:  There are some not insignificant changes by Felix also in vspec.py

vspec.py
```diff
@@ -240,17 +241,22 @@ def convert_yaml_to_list(raw_yaml):
 
     return lst
 
-
-
-def load_tree(file_name, include_paths, exclude_private=False, break_on_noncore_attribute=False):
+# As of today the VSS language/metamodel specification contain some attributes (e.g. arraysize) 
+# that is ignored by most tools in this repository. A tool can specify which VSS core attributes that it
```
gunnarx, line 245: The point about it being ignored by most tools today seems not relevant, and I would like to remove it.

But idea and documentation that tools can specify the unsupported attributes seems reasonable **if** some attributes might not serve a purpose for certain tools?  But do we have such an today, a tool that actually does not need a certain piece of metadata?  

The original reason stated (is ignored by some/most tools) seems for most cases to be a bug in the tool, rather than a logical behavior?

In summary I support introducing this feature only if we know a case for it today.  Otherwise it is better to wait until that case comes up.

gunnarx (changes requested):
First, thanks for taking this on because it seems to be something worth looking at.

I am requesting changes because of at least the comment, but possibly more, after discussion of the feedback.

I am wondering if for a format translation tool like vspec2csv it should not be doing anything in particular with the data.  I mean a code generator might actually need to know the array size to do something with it.

Tools can still validate that only valid metadata names are used of course.  

But beyond that I think it is a matter of specification of intended behavior of the tool.  If the CSV file should include array size or not.  If it should then it should (and the bug should be fixed), but if it is intended behavior that arraysize is not included in the CSV output, then I don't really see a need for a warning - the tool is operating as intended?

In both cases, there would not be any statement that array size is not supported by the tool.

vspec.py
erikbosch: Can we assume that all tools shall support all metamodel elements? If we consider it as natural that csv/json (but possibly not protobuf as it is a "contrib" tool) shall support "arraysize", then I agree that we instead shall create issues in vss-tools on the bug that csv/json tooling does not support arraysize.

danielwilms:
Discussion:
- remove warning for arraysize
- add tickets for tools

erikbosch:
Based on comments I close this and instead open #59 that only adds support for arraysize

PR #64, opened 2021-05-31
vspec2graphql.py: (New GraphQL schema generator)

gunnarx:
Based on the the protobuf generator and the output according to the "vehicle.ts" file that was previously provided for the CCS-arch-demo (which we only had the file for, not the generator that created it!)

### DIFF
--- a/contrib/vspec2graphql.py
+++ b/contrib/vspec2graphql.py
@@ -50,12 +50,15 @@ indent = '   '
 def traverse_tree(tree, graphql_file):
     tree_node: VSSNode
     for tree_node in filter(lambda n: n.is_branch(), PreOrderIter(tree)):
-        graphql_file.write(indent + '"""\n')
-        graphql_file.write(indent + tree_node.description + "\n")
-        graphql_file.write(indent + '"""\n')
-        graphql_file.write(indent + f"type {tree_node.qualified_name('_')} {{" + "\n")
-        print_message_body(tree_node.children, graphql_file)
-        graphql_file.write(indent + "}\n\n")
+        if len(tree_node.children) == 0:
+            print(f"WARNING: Skipping branch with no children: {tree_node.qualified_name('.')}")
+        else:
+            graphql_file.write(indent + '"""\n')
+            graphql_file.write(indent + tree_node.description + "\n")
+            graphql_file.write(indent + '"""\n')
+            graphql_file.write(indent + f"type {tree_node.qualified_name('_')} {{" + "\n")
+            print_message_body(tree_node.children, graphql_file)
+            graphql_file.write(indent + "}\n\n")
 
 def print_message_body(nodes, graphql_file):
     for i, node in enumerate(nodes, 1):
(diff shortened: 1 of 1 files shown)


contrib/vspec2graphql.py
```diff
@@ -0,0 +1,134 @@
...
+        graphql_file.write(indent + f"type {tree_node.qualified_name('_')} {{" + "\n")
+        print_message_body(tree_node.children, graphql_file)
+        graphql_file.write(indent + "}\n\n")
+
+def print_message_body(nodes, graphql_file):
+    for i, node in enumerate(nodes, 1):
+        data_type = node.qualified_name('_')
+        if not node.is_branch():
+            dt_val = node.data_type.value
+            # TODO: FIX ARRAYS!
```
erikbosch, line 64: What is it that is left to do?

gunnarx: Good catch.  I already added array support, so this comment should be removed.

gunnarx: New version below.  I rewrote a few comments for clarity.

contrib/vspec2graphql.py
```diff
@@ -0,0 +1,134 @@
...
+        # CamelCase but starting lower case
+        name = node.name[0].lower() + node.name[1:]
+        graphql_file.write(indent * 2 + '"""\n')
+        graphql_file.write(indent * 2 + node.description + "\n")
+        graphql_file.write(indent * 2 + '"""\n')
+        graphql_file.write(indent * 2 + f"{name}: {data_type}\n\n")
+
+def usage():
+    print(
+        """Usage: vspec2graphqlbuf.py [-I include_dir] ... [-i prefix:id_file] vspec_file output_file
```
erikbosch, line 80: Not "vspec2graphqlbuf.py" right? Remove "buf"?

contrib/vspec2graphql.py
```diff
@@ -0,0 +1,134 @@
...
+  -I include_dir       Add include directory to search for included vspec
+                       files. Can be used multiple times.
+
+  -i prefix:uuid_file  File to use for storing generated UUIDs for signals with
```
erikbosch, line 84: called id_file on line 80, uuid_file here. Use same?

contrib/vspec2graphql.py
```diff
@@ -0,0 +1,134 @@
...
+                       a given path prefix. Can be used multiple times to store
+                       UUIDs for signal sub-trees in different files.
+
+ vspec_file            The vehicle specification file to parse.
+ graphql_file            The file to output the graphql file to.)
```
erikbosch, line 89: Called output_file on oline 80, graphql_file here, align?

gunnarx: Fixed

contrib/vspec2graphql.py
gunnarx: Aligned. 

I'm wondering if a tool like this should trigger a rewrite of the UUID file or not.  The others do, so this one also does, but I'm not personally so invested in the UUID concept.

contrib/vspec2graphql.py
gunnarx: Fixed

contrib/vspec2graphql.py
erikbosch: Do we have any documentation somewhere on how uuid and prefix are supposed to be used? Is it someone that actually use them and the generated uuid_file today (Could be a topic for general tool documentation)

gunnarx:
I added a check to not generate types for empty branches, but that is not enough because a branch could have a sub-branch but the sub-branch is empty.  So it needs a better solution:

For example.

1. Recursive/iterative traversal down the branch to see if there exists _any_ non-branch children
2. or add a file that specifies a filter for what to include in the generation (so that problematic branches in current VSS master can be avoided)

(I think the empty branches should be considered as a bug in the VSS catalog but that doesn't help the issue at the moment)

gunnarx:
Merging, as agreed in https://github.com/GENIVI/vehicle_signal_specification/pull/290.  A better empty-branch check is still pending but we might just avoid having empty branches instead.  We will see.

Issue #66, opened 2021-06-09
broken installation


barbieri:
While it's common to use this package right from the development source, it would be nice if it could be installed and used as dependency in other projects.

However the `setup.py` is not installing `model` and it fails to import:

```
$ python3 vspec2csv.py
Traceback (most recent call last):
  File "/private/tmp/test-vss/bin/vspec2csv.py", line 18, in <module>
    import vspec
  File "/tmp/test-vss/lib/python3.9/site-packages/vspec.py", line 22, in <module>
    from model.vsstree import VSSNode
ModuleNotFoundError: No module named 'model'
```

Also, it would be nice to list the runtime dependencies in `setup.py` under `install_requires` as per https://packaging.python.org/discussions/install-requires-vs-requirements/ leaving in https://github.com/GENIVI/vss-tools/blob/master/requirements.txt just the development dependencies (also, could be replaced by newer tooling such as `pipenv` and its `Pipfile`).

PR #67, opened 2021-06-09
Install fixes

barbieri:
Removed unused dependencies from `requirements.txt`, update Python version to 3.7 in `setup.py` and fix the installation of the package. Last but not least, add the `Pipfile` and instructions on how to use `pipenv`.

Unfortunately the existing layout of a single `vspec.py` file and a `model/` directory does not match the Python package expectations, in particular `model` is a too broad/generic name and may cause issues. To solve this, a `vspec/` package was created and `vspec.py` was moved as its `__init__.py` (existing imports will work, unchanged); while `model/` was made a sub-package, however the imports had to be changed to be `vspec.model`.

Closes: #66


gunnarx (approved):
As discussed we should fix the build check and make sure it runs OK but otherwise OK from me.

barbieri:
already rebased, would be nice to have more people to test it (install and see if the installed binaries work, it does here)

Pipfile
```diff
@@ -0,0 +1,17 @@
...
+stringcase = ">=1.2.0"
+deprecation = ">=2.1.0"
+pytest = ">=2.7.2"
+PyYAML = ">=5.1"
+vss-tools = {editable = true, path = "."}
+
+[dev-packages]
+
+[requires]
+python_version = "3.7"
```
SebastianSchildt, line 17: This makes it fail on Ubuntu LTS, where default Python is 3.8.5.   as ">=" does not seem to be supported maybe just use "3" ?

SebastianSchildt (approved):
I can confirm this still works for JSON and it looks fin. And I can also generate CSV.

And just because I noticed: Small problem in the readme (not created in this PR):

* CSV File generation -> vss.csv:
 ```python /vspec2csv.py -i:../vehicle_signal_specification/spec/VehicleSignalSpecification.id ../vehicle_signal_specification/spec/VehicleSignalSpecification.vspec vss.csv```

This should be ```./vspec2csv.py```  and the other paths are also weird, when executing from cwd vss-tools (directly or in venv)  it should be something like

```python ./vspec2csv.py -i:../tmp/asdasd ../spec/VehicleSignalSpecification.vspec vss.csv```

Pipfile
barbieri: We should agree on a version (for the Pipfile) so everything matches (see the `Pipfile.lock`, where exact versions are used).

While I'm totally fine to move to 3.8.5 or something else, the way to use `pipenv` is to pair it with [`pyenv`](https://github.com/pyenv/pyenv), then when you `pipenv` it will use the version.

Let's agree on a version and I'll document the `pyenv` in the `README.md`

barbieri: just did this, fixed to 3.8.5 and explained how to use `pyenv`. This helps everythone (being on Ubuntu or not) to have the same setup.

Of course, we should still test other systems in the CI, usually via a python matrix (ie: latest 3.7, 3.8, 3.9...)

barbieri:
@SebastianSchildt also fixed (in another commit, but in this same PR) the `vspec2csv` as you pointed out

SebastianSchildt (approved):

PR #69, opened 2021-06-15
Removing uuid parameter support

erikbosch:
We discussed at a previous meeting that the id-parameter in tools is not used and does not make sense. Here is a change reflecting that. Parameter support is being removed, but internal model of creating UUID remains.

Intermediate build problems expected as make-files needs to be adapted to the change

### DIFF
--- a/contrib/vspec2c.py
+++ b/contrib/vspec2c.py
@@ -17,13 +17,9 @@ import getopt
 import hashlib
 import json
 def usage():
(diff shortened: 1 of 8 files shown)


SebastianSchildt:
I "like"  this change, but did not try it out yet

gunnarx:
OK, merged github action into master, and then rebased on that.  Now it builds OK.

PR #71, opened 2021-06-22
Build test using Github Actions workflow (vss-tools)

gunnarx:
Still having some issues with Travis builds so here is a stopgap using the built in runners for GitHub Actions / Workflows.

I did not remove .travis.yml in this commit since we probably want to test this first and we might sort out Travis.

### DIFF
--- a/.github/workflows/buildcheck.yml
+++ b/.github/workflows/buildcheck.yml
@@ -1,0 +1,45 @@
+name: Standard Build Check
+
+on: [push, pull_request]
+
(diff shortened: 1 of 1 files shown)


gunnarx:
Note:  I had to push it to a branch on **this** repository also, to trigger the build check? Before that build check was only visible on the source repo. Anyway, see results above (or  [with this link](https://github.com/GENIVI/vss-tools/pull/71/checks?check_run_id=2884936125))

PR #73, opened 2021-06-24
README.md: add link to VSS Tools section of official online VSS doc

slawr:
The official online VSS documentation contains a section on the VSS Tools.
Follow the pattern set by the VSS README.md and link to it to aid
discoverability by newcomers.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,5 +1,7 @@
 # VSS Tools Usage
 
+Additional information can be found in the official VSS documentation at: [VSS Tools](https://genivi.github.io/vehicle_signal_specification/tools/)
+
 ## Contents
 1. Prerequisites
 2. Project Setup


SebastianSchildt:
I am not sure, we should link it (at this point), as I think the documentation in the main hugo docs is outdated. (And it should be discussed where tools documentation should live).

My view is: the main VSS hugo page at http://localhost:1313/vehicle_signal_specification/   should not contain any tools documentation at all, and only provide a lnik to this repo.

IN _here_ we can decide whether github Readmes are enough (I think so), or we want to set up a separate Hugo instance

SebastianSchildt:
See also #74

gunnarx:
Agree with Sebastian. Let's restructure the documentation of the tools.

slawr:
Agree on the points about the different documentation levels. My thinking behind the PR was my discovery path through the doc. I found the online VSS doc early on and therefore also info on vss-tools. I have also seen links to vss-tools source in CVII. The README however limits itself to install, run and test. There is no information about what the tools are intended to do. That however can be found in the online doc including a useful diagram putting it in context with the vspec files. 

I therefore thought a simple link would help discoverability until the documentation catches up, but its a minor/trival change and not a hill I am looking to die on.

I feel I lack the deep experience of the tools to attempt rewriting the documentation. That would be best handled/discussed in an enhancement ticket I feel. So I am leaning towards simply closing this?

danielwilms:
@gunnarx, please close as tools are not described in vss anymore.

slawr:
Closing as requested by community

PR #78, opened 2021-07-01
Changing instances variable initalization. 

### LINKED ISSUE #77: Variable initialized with None being iterated
At the subfunction  `expand_instances` inside `expand_instances` (yes, both have the same name) we can find [a declaration of a variable called **instances**](https://github.com/GENIVI/vss-tools/blob/14d044699bc358590009aaf1740864c149d3c9be/vspec/__init__.py#L396) and later [the iteration of the same variable](https://github.com/GENIVI/vss-tools/blob/14d044699bc358590009aaf1740864c149d3c9be/vspec/__init__.py#L429). The problem arises as between them there is no guarantee that it will be initialized with a list or any other iterable object, causing it to raise an _Exception_.

oliveirarleo:
resolves #77

### DIFF
--- a/vspec/__init__.py
+++ b/vspec/__init__.py
@@ -393,7 +393,7 @@ def expand_instances(flat_model):
         new_flat_model = []
         elements_to_instantiate = []
         base_name = ''
-        instances = None
+        instances = []
         current_inst_path = ''
 
         for elem in flat_model:


SebastianSchildt (approved):
Looks sane to me, but not tested. Vote to merge, once somebody triggers CI and it runs successfully.

danielwilms:
@gunnarx, please merge.

PR #79, opened 2021-07-01
Adding the datatypes HSVColor and DateTime.

### LINKED ISSUE #76: Change the Unit and VSSDataType to support custom types and units

```
    LUX="lux"
    HERTZ="Hz"
    MILLISECOND="ms"
```

oliveirarleo:
Adding the datatypes HSVColor and DateTime.

The type DateTime is added according to the [ISO 8601](https://en.wikipedia.org/wiki/ISO_8601)
The type HSVColor is added according to the [HSL Format](https://drafts.csswg.org/css-color/#the-hsl-notation)
resolves #76

### DIFF
--- a/tests/model/test_contants.py
+++ b/tests/model/test_contants.py
@@ -105,6 +105,8 @@ class TestConstantsNode(unittest.TestCase):
         self.assertEqual(VSSDataType.DOUBLE, VSSDataType.from_str("double"))
         self.assertEqual(VSSDataType.STRING, VSSDataType.from_str("string"))
         self.assertEqual(VSSDataType.UNIX_TIMESTAMP, VSSDataType.from_str("UNIX Timestamp"))
+        self.assertEqual(VSSDataType.DATETIME, VSSDataType.from_str("DateTime"))
+        self.assertEqual(VSSDataType.HSVCOLOR, VSSDataType.from_str("HSVColor"))
 
     def test_invalid_vss_data_types(self):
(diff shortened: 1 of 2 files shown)


barbieri (approved):
thanks, let's see what @gunnarx and @Floix think

barbieri:
@oliveirarleo please add `tests/model/test_contants.py` covering them

barbieri (approved):

SebastianSchildt:
If theres is consensus to merge I will not block, but I feel this is the "most wrong" way to add units. We create a randomly curated list of "allowed" units, some of which are in VSS and some not, and probably the first thing everybody needs to to, once he uses vSS besides the standard catalogue is hacking in more units in his fork  (which I already did, and which also ultimately leads to this PR)

Anyway, my thoughts on this are here already: https://github.com/GENIVI/vss-tools/issues/76#issuecomment-872043539

danielwilms:
@barbieri: create an issue 
* for datatypes and 
* a second one for units (< how to deal with additional units curated list or not)  
* list possible datatypes as start to be more "future proof"

@danielwilms: create issue for units mismatch code vs. docs

barbieri:
@danielwilms created #89 and #91 as requested. Since you did not create the issue on the units mismatch, I did that as #90.

danielwilms:
@barbieri That's generous ;) thanks a lot!

danielwilms:
@gunnarx, please close the PR

gunnarx:
Refer to tickets for more info.

PR #80, opened 2021-07-01
Adding the units for illuminance, frequency and time, following the SI

### LINKED ISSUE #76: Change the Unit and VSSDataType to support custom types and units

oliveirarleo:
Adding the units for illuminance, frequency, and time, following the SI notation as described in [International_System_of_Units wiki page](https://en.wikipedia.org/wiki/International_System_of_Units#Units_and_prefixes.)

Adding the following units:
 - Lux as 'lx'
 - Hertz as 'Hz'
 - Millisecond as 'ms'
resolves #76

### DIFF
--- a/tests/model/test_contants.py
+++ b/tests/model/test_contants.py
@@ -47,6 +47,7 @@ class TestConstantsNode(unittest.TestCase):
         self.assertEqual(Unit.VOLT, Unit.from_str("V"))
         self.assertEqual(Unit.AMPERE, Unit.from_str("A"))
         self.assertEqual(Unit.SECOND, Unit.from_str("s"))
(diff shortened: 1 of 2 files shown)


barbieri (approved):
looks good to me, let's see what @Floix, @gunnarx and others will say

barbieri:
@oliveirarleo would be nice to add the tests `tests/model/test_contants.py`

barbieri (approved):

barbieri (approved):

barbieri (approved):

PR #81, opened 2021-07-02
Enum improvements

barbieri:
based on #76 discussion, add script to auto-document the enums (done for `Unit`)

Also use a new `EnumMeta` subclass to offer `values()` and an efficient `from_str()` in every enumeration that uses it.


SebastianSchildt:
No problems with the "contrib" part, but need to discuss whether we want this change to the model/constants.py

I agree there should be a "single source of truth" wrt to accepted units in VSS standard catalogue/tooling/docs. I doubt that Python _code_ is the correct place for that.

See discussion here: https://github.com/GENIVI/vss-tools/issues/76

barbieri:
@SebastianSchildt the choice to have the code as the source of truth is that the code is the only thing that really works. Paper takes whatever, the formats (even markdown) are too relaxed. Code is a bit more strict, we can add unit tests and have it to output the documentation.

contrib/document_constants.py
```diff
@@ -0,0 +1,155 @@
...
+        rows = [
+            (f"`{m.name}`", m.value) for m in self.cls.__members__.values()
+        ]
+        lines.append(create_table(self.table_header, rows))
+        lines.append("")
+        return "\n".join(lines)
+
+
+DocumentGenerators = Union[Type[DocumentEnum]]
+Documents = Union[DocumentEnum]
```
gunnarx, line 154: As far as I understand a Union type with only one choice is not useful.

So this
```
DocumentGenerators = Union[Type[DocumentEnum]]
Documents = Union[DocumentEnum]
```
is really the same as this (?)
```
DocumentGenerators = Type[DocumentEnum]
Documents = DocumentEnum
```

If you agree, I guess you could also use the original types directly in the expressions that follow instead of giving them a new name.  Unless the "renaming" is done just for easier understanding.  What's the idea here?

contrib/document_constants.py
```diff
@@ -0,0 +1,155 @@
...
+collect_map: Mapping[str, DocumentGenerators] = {
+    "Enumerations": DocumentEnum,
+}
+
```
gunnarx, line 158: Do you expect more than one entry in the map later on?  If not, the iteration over a map done in collect() seems unnecessary.

contrib/document_constants.py
```diff
@@ -0,0 +1,155 @@
...
+
+def collect() -> Mapping[str, Sequence[Documents]]:
+    collected = {}
+    for name in constants.__all__:
+        element = getattr(constants, name)
+        for key, doc in collect_map.items():
+            if issubclass(element, doc.kind):
```
gunnarx, line 165: Similar to above... if there's only one entry in the map, then I imagine you already know if this if statement is true or false.

vspec/model/constants.py
```diff
@@ -59,6 +66,14 @@ class StringStyle(Enum, metaclass=EnumMetaWithReverseLookup):
 
 
 class Unit(Enum, metaclass=EnumMetaWithReverseLookup):
+    """Data Unit Types
```
gunnarx, line 86: I would rebase and split these doc strings into a separate commit.  It seems unrelated to the other code changes.

gunnarx (changes requested):

vspec/model/constants.py
gunnarx: or maybe... why not put it together with the previous change, where you did a significant rewrite of the Unit constants.  Feels like it more belongs there.

contrib/document_constants.py
```diff
@@ -0,0 +1,155 @@
...
+    a human friendly table (the text representation looks like a table).
+
+    To avoid too much resizing when the cells change, the sizes are rounded
+    using `size_rounding`.
+    """
+    def round(n):
+        remainder = n % size_rounding
+        if not remainder:
+            return n
+        return n + remainder
```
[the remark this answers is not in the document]
barbieri: I didn't do much thinking on this one, I just wanted to have some way to avoid resizing, we can indeed change it to be as you said, it would make more sense, indeed.

contrib/document_constants.py
```diff
@@ -0,0 +1,155 @@
...
+        for i in range(count):
+            max_lengths[i] = max(max_lengths[i], round(len(r[i])))
+
+    fmt = "|"
+    for ml in max_lengths:
+        fmt += f" %-{ml}s |"
+
+    ret = [
+        fmt % header,
+        "|" + "|".join(f":{'-' * n}-" for n in max_lengths) + "|",
```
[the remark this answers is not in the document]
barbieri: I can rewrite in a more traditional way, were it creates a list of the strings formatted... however in here it's doing the markdown table separator:

```
|:-----|:-----|:---------|
```

the number of `-` depends on the column width.

vspec/model/constants.py
```diff
@@ -74,43 +91,63 @@ class Unit(Enum, metaclass=EnumMetaWithReverseLookup):
...
+        "Acceleration"
+    )
+    LITER = VSSConstant("l", "Liter", "Volume")
+    DEGREECELCIUS = VSSConstant("celsius", "Degrees Celsius", "Temperature")
+    DEGREE = VSSConstant("degrees", "Angle in degrees", "Angle")
+    DEGREEPERSECOND = VSSConstant(
+        "degrees/s",
+        "Angular speed in degrees/s",
+        "Angular Speed"
+    )
```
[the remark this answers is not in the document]
barbieri: no idea :-D

vspec/model/constants.py
```diff
@@ -26,6 +26,23 @@
...
+    """String subclass that can tag it with description and domain.
+    """
+    __slots__ = ("description", "domain")
+
+    def __new__(
+        cls,
+        value: str,
+        description: str = "",
+        domain: str = "",
+    ) -> None:
```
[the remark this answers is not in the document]
barbieri: with the types signatures line become too long, it's over 80 columns and flake8 and other PEP8 checkers will complain:

```py
    def __new__(cls, value: str, description: str = "", domain: str = "") -> None:
```

is 83 chars wide

vspec/model/constants.py
```diff
@@ -13,7 +13,7 @@
 
 # noinspection PyPackageRequirements
 from enum import Enum, EnumMeta
-from typing import Sequence, Type, TypeVar
+from typing import Optional, Sequence, Type, TypeVar
```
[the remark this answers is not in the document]
barbieri: it's not anymore... in the original code it was in domain/description, but making them `=""` simplified the rest of the code.

good catch :-)

contrib/document_constants.py
barbieri: yes, it's the same. The idea is that eventually we'll be documenting more than `Enum` and adding it there would be slightly more straightforward.

Say we add some document generator for other non-string classes, just `, OtherType` and it's done. But I don't mind just skipping the `Union[]` until we reach there

contrib/document_constants.py
barbieri: as per above, in the future we may document more than just enums, say some other classes or even some more general constants, like a `SEPARATOR` or the likes.

vspec/model/constants.py
barbieri: let me know if they should be merged or not

contrib/document_constants.py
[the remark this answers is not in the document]
barbieri: fixed in the updated version, also with doctests :-)

contrib/document_constants.py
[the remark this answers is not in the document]
barbieri: improved in the new version, also with doctests

vspec/model/constants.py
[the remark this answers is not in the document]
barbieri: removed in the new version

barbieri:
@gunnarx rebased and fixed some issues. Now we have some doctests:

```sh
python3 -m doctest contrib/document_constants.py
```

vspec/model/constants.py
```diff
@@ -59,46 +83,79 @@ class StringStyle(Enum, metaclass=EnumMetaWithReverseLookup):
...
+    DEGREE = VSSConstant("degrees", "Angle in degrees", "Angle")
+    DEGREEPERSECOND = VSSConstant(
+        "degrees/s",
+        "Angular speed in degrees/s",
+        "Angular Speed"
+    )
+    KILOWATT = VSSConstant("kW", "Kilowatt", "Power")
+    KILOWATTHOURS = VSSConstant("kWh", "Kilowatt hours", "Power")
+    KILOGRAMM = VSSConstant("kg", "Kilograms", "Weight")
+    VOLT = VSSConstant("V", "Potential difference in volt", "Electrical")
```
Floix, line 115: Why do we document units?

I mean, If you do want to understand volts, why not look it up: https://en.wikipedia.org/wiki/Volt

contrib/document_constants.py
```diff
@@ -0,0 +1,297 @@
...
+    ret = [
+        fmt % filter_visible_columns(header),
+        create_table_header_separator(visible_columns_lengths),
+    ]
+    for r in rows:
+        ret.append(fmt % filter_visible_columns(r))
+    return "\n".join(ret)
+
+
+class DocumentEnum:
```
Floix, line 199: I'd recommend to check for Markdown implementations first, before we implement this ourselves. Maybe https://github.com/Python-Markdown/markdown is worth a shot.

Floix (changes requested):

contrib/document_constants.py
barbieri: not sure what you mean, I'm just generating https://raw.githubusercontent.com/GENIVI/vehicle_signal_specification/master/docs-gen/content/rule_set/data_entry/data_unit_types.md so it can be in sync with code

like the other comment, it's from that same file... my idea is just to write it in one place (ie: code) and have the documentation to be generated from it.

vspec/model/constants.py
```diff
@@ -59,46 +83,79 @@ class StringStyle(Enum, metaclass=EnumMetaWithReverseLookup):
...
+    MONTHS = VSSConstant("months", "Months", "Time")
+    UNIXTIMESTAMP = VSSConstant(
+        "UNIX Timestamp",
+        "Seconds since January 1st 1970 UTC",
+        "Time"
+    )
+    PASCAL = VSSConstant("Pa", "Pascal",  "Pressure")
+    KILOPASCAL = VSSConstant("kPa", "Kilo-Pascal", "Pressure")
+    PERCENT = VSSConstant("percent", "Percent", "Relation")
+    CUBICMETER = VSSConstant("cm^3", "Cubic Meters", "Volume")
```
erikbosch, line 129: Cubic centimeters

vspec/model/constants.py
erikbosch: I can see some use of "Domain", like "Power" or "Volume" to be able to quickly find out which units that VSS (and related tools) support in a certain domain. For example, if I want to specify a signal for the maximum power consumption of the front wipers, I could check for units marked with "Power" and find "kW", "kWh" and "PS" and select one of them. That goes quicker than checking unit one by one, not all persons know in their head what e.g. "PS" and "lbs" are used for.

(But if used for that purpose "Power" is not a good description for "kWh", it would rather be "Energy")

erikbosch: Documentation becomes more essential if we have units that might be ambiguous. A typical example could be "stars" below. To be useful I think that it needs some explanation. Do we e.g. refer to a particular star rating (e.g. Euro NCAP crash rating) or could it be any star rating using any scale? If we give some documentation, like "rating, minimum value is 0, maximum value is context/signal dependent" then that information could possible be used by UI designers as a hint on how to present a signal using "star" in a UI.

contrib/document_constants.py
Floix: I figured that from the code, but if we check the underlying issue this is not solving it? This makes it even harder to add units and the documentation we add is to brief to add any value imho. Easy fix would be: Update the .md file and be done with it. Why wouldn't we do that?

In addition, you wrote a markdown generator. Why reinvent the wheel, if we can reuse existing markdown libs?

vspec/model/constants.py
barbieri: but which one is wrong (which one should I change)? The value or the name, see https://github.com/GENIVI/vss-tools/blob/master/vspec/model/constants.py#L85 it's already `CUBICMETER = "cm^3"` 👀

contrib/document_constants.py
barbieri: Because currently they diverged, see #90. The current process that needs to add documentation to one repo, code to another leads to divergences.

My proposal is to keep it in a single place (code) and provide a tool that generates the documentation. So https://github.com/GENIVI/vehicle_signal_specification tooling can be updated to execute this script *before* the markdown is converted to HTML to be used in the site.

As for the markdown generator, I'm just formatting the tables like they were in the https://raw.githubusercontent.com/GENIVI/vehicle_signal_specification/master/docs-gen/content/rule_set/data_entry/data_unit_types.md I didn't find any generator for that, do you? The repository you sent is something else, it **reads** markdown and outputs html. Note that this can be skipped, if nobody bothers to have a human readable (aligned) table, we can just output it in a simpler way that markdown will still understand and correctly format the HTML table.

vspec/model/constants.py
erikbosch: Our only use of `cm^3` in VSS is for displacement where we in the description clearly says cubic centimetres (and that is the unit that makes sense, if you are to represent it as a uint16. So I would say changing to "CUBICCENTIMETER" makes sense. I have a vague memory that we spotted this error earlier at some time, but it seems we did not fix it.

(And it seems that we prefer US spelling generally, e.g. centimeter and tire)

```
Displacement:
  datatype: uint16
  type: attribute
  description: Displacement in cubic centimetres.
  unit: cm^3
```

contrib/document_constants.py
Floix: Ok, so your proposal is to generate markdown to afterwards manually add it to the wiki pages? What's the benifit here? The steps are essentially the same: You add/remove/change a unit, you run the tools, you update the documentation (which is not copy/paste instead of edit which may safe you 5 mins of work).

However, with this approach if you change the format (e.g. add/remove a column) you now have to do code changes to be in sync, whereas in the "classical" approach you just edit the markdown file.

Imho, we add  complexitiy to the tools here that make documentation more complex to change and instead of simplifing adding units as initially requested in https://github.com/GENIVI/vss-tools/issues/76 we make situation worse.

I understand, that those units and the docs should be in sync. Fair point. Let's just update the documentation and be done with it?

In addition, we have a setup here that we'd like to contribute where you can locally change the units config for your needs, which will help prototyping (as requested in https://github.com/GENIVI/vss-tools/issues/76). With the changes you propose here, this will get more complicated just to make a markdown table copyable. I think we should not add this complexitity.

barbieri: I don't know why you consider it manual. Since the CI scripts already do checkout the `vss-tools` code, it's a matter or running another command to output the markdown file in addition to to convert it to HTML. It needs automation, but doesn't require manual intervention, is that your understanding?

As for adding a new column bringing more work, it should be the same, no? You'd need to add the information to the markdown or Python code. You'd need to add a need to add the table header in markdown, so you do in this class (add a string to `table_header` tuple below. Shall you want to avoid adding the new column to all of the entries (you'd have to do that in markdown as well), you can define a default value in `VSSConstant` or right in the loop below (that calls`rows.append`), which you can't do in markdown. That said, if you could share more details on your Point of View, why it would be more work, let me know.

I can't comment on the code I didn't see, but the registry or file format can incorporate the documentation in a similar way. Say you have a constant name and a value, you can add description and domain as well. Even if that's optional, no?

vspec/model/constants.py
barbieri: fixed in the updated branch, I added this as a first commit, independent from the VSSConstants and scripts so it can be merged regardless of the tools. I won't open another PR yet, but if this PR is rejected then I can create another one with just thix fix

contrib/document_constants.py
Floix: It's manual, since there is not automation in place that does change the markdown file automatically.

I assume it's an order of magnitude more complex to change python code than the markdown itself and requires a different skillset.

vspec/model/constants.py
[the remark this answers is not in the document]
gunnarx: Can't believe PEP8 requires ~80 columns... outdated crap. :-1:   Well... then we stick with this style I guess.

vspec/model/constants.py
gunnarx: > let me know if they should be merged or not

In my opinion, no need to squash them together (so I didn't ask for it).

contrib/document_constants.py
gunnarx: Interesting exchange of ideas here.  I'll try to be brief, but a lot to cover.  

I often like the simple approach of just changing the files (documentation) and move on like @floix suggests.  But we have seen at least one case of this failing.  There is a proposal here so we should look at it I think.  But there are some points to cover first:

1) Whether or not the project should allow for adding your own units is still an open question!   We're debating the pros and cons.   (And #76 is perhaps no longer really an appropriate ticket to track the discussions we are having on the weekly VSS development call, primarily because we want to treat datatypes and units separately).  #91 was created to clarify the current state of the conversation in the VSS development group.

2) I agree with the logic that this PR basically depends on **if** there should be extensible units in the first place.
But see below for details on how to resolve that first.  We should recognize that it is also possible to remove code, if a strategy is later changed.  In some cases the right thing to do is to fix things for the _current_ situation.

The rest is best answered by quoting:

@floix wrote:
> It's manual, since there is not automation in place that does change the markdown file automatically.

I don't understand why you keep pushing this since @barbieri explained the plan for CI to do this?  Actually, GitHub actions already rebuild the docs for every change on master branch for a few weeks now, ref: [this line in the workflow](https://github.com/GENIVI/vehicle_signal_specification/blob/049b53fefd1f0bccc4a9f431a893cc03ecd10ae2/.github/workflows/buildcheck.yml#L46), and it could take care of running this tool too.  If you here only mean that this PR does not **yet** include the change to add that to the GitHub Actions workflow, then that seems an unnecessary nitpick.  (We actually still have the complication in VSS development that the Makefile is currently owned by **vehicle_signal_specification** repository instead of **vss-tools** repository, so it is not really possible to include the change in the same PR anyhow).

> I assume it's an order of magnitude more complex to change python code than the markdown itself and requires a different skillset.

Reasonable point although "order of magnitude" is a subjective evaluation.  But the alternative to this PR seems to be "change _both_ **code** and **docs**, or change only **code**", so does it really apply_

Basically we seem to have these possible futures:

A) A mechanism is added to add your own custom units by a separate configuration file (no "code" change)
or:
B) Possible units are specified in the vss-tools python code, as today:
    B1) Units are specified in code, and have to also be added to the markdown documentation at the same time.
    B2) Documentation is generated from the specification of units in the code file.  (=this PR)

@floix:
> In addition, we have a setup here that we'd like to contribute 
...

In such situations, I think it's fine to say:  Please wait with this change a week or two, because we have another idea coming.  But it cannot hold up the progress for a very long time  - then you have to present the idea at least (even if the code is taking a bit longer).  And in this case here, it is the fundamental question (#76, #91) that need to be dealt with primarily.

@floix:
> where you can locally change the units config for your needs, which will help prototyping (as requested in #76).

I would note that "we have something we'd like to contribute" is just more along the same lines as some other PRs that we found were coming too early**.  I feel that vss-tools is struggling from code change proposals to vss-tools being used to push some fairly major fundamental policy changes for VSS development (as I see it).  I think we see that it is not a good way -- this is why we raise issues and discuss the really fundamental things separately.  Then the PRs should just reflect the necessary code changes for the decision.  (And while I'm on the subject I have to say something I kept quiet about last week:  #91 is a VSS question first, before it becomes a vss-tools question, so the issue should have been raised in the VSS repository and not vss-tools.  *Because* these things are complex and have long lasting effects, I think we should be careful about the process.)

(** I think there was consensus on the VSS call that @barbieri could propose this PR so it could be evaluated.)

gunnarx: Minor answer also to this (it is less about the big picture and more about details)

@Floix:
> In addition, you wrote a markdown generator. Why reinvent the wheel, if we can reuse existing markdown libs?

My view.  It is not _always_ better to add additional dependencies, if the code is trivial and can be written without adding  dependencies to something that we only use a tiny part of.  But this has to be evaluated for each case.  So I would second @barbieri 's question whether there is a reasonable choice to use that can generate these types of tables easily, and if so let's compare with that choice...

contrib/document_constants.py
gunnarx: On this and the 2 points above I'd like to know what the plan is to document more things.  (And if that still fits into the same file, i.e. document_constants.py?)  
If we don't know the plan I still request the code to be simplified down to what it is actually doing now _only_.

contrib/document_constants.py
gunnarx: see next

contrib/document_constants.py
gunnarx: see next

contrib/document_constants.py
barbieri: ok, since right now all constants are enums, I'll simplify this part.

contrib/document_constants.py
barbieri: about the last point, also note I can shrink the code if there is no need to render properly aligned *textual* table. Markdown itself will allow tables that are not perfectly aligned, so it's much simpler to generate (no need to measure, create the formatting strings and then loop again to print the table).

However, since the code is not that big and being able to read a nicely aligned textual table is always helpful, I'd like to keep what I did.

contrib/document_constants.py
barbieri: it's done in the updated branch. Please notice the `issubclass(element, Enum)` remains as we need to be defensive on the `constants.__all__`, in case any non-`Enum` symbol is added to `__all__` it would break the script.

but the `collect_map`, `kind` and all are gone.

contrib/document_constants.py
```diff
@@ -0,0 +1,280 @@
...
+
+
+def main():
+    ap = get_args()
+    args = ap.parse_args()
+
+    out = args.output_file
+    collected = collect()
+    for doc in collected:
+        if args.name and args.name != doc.name:
```
barbieri, line 273: @gunnarx I still keep this `--name` and check so we can use the same script to generate different markup files for a single type, like we have:
- https://github.com/GENIVI/vehicle_signal_specification/blob/master/docs-gen/content/rule_set/data_entry/data_types.md
- https://github.com/GENIVI/vehicle_signal_specification/blob/master/docs-gen/content/rule_set/data_entry/data_unit_types.md

(Note: I just realized/noticed that the `data_types.md` contains some different columns, so shall we keep this tool I'll update the `VSSConstant` and this tool to handle different columns, likely `VSSUnitConstant` and `VSSDataTypeConstant`)

gunnarx:
I think we agreed last week's meeting that this PR is obsolete because new solutions have been agreed, and that it should be closed.

PR #83, opened 2021-07-05
Using separator "/" at qualified_name()

### LINKED ISSUE #82: Use of DEFAULT_SEPARATOR in detect_and_merge
In these two lines of code
https://github.com/GENIVI/vss-tools/blob/14d044699bc358590009aaf1740864c149d3c9be/vspec/__init__.py#L748
https://github.com/GENIVI/vss-tools/blob/14d044699bc358590009aaf1740864c149d3c9be/vspec/__init__.py#L752

The method `qualified_name()` is used without any parameter, so it uses the `DEFAULT_SEPARATOR`, [which is the character "."](https://github.com/GENIVI/vss-tools/blob/7e36e5a357d276e2a25f273b7774f6b5e4c1eddd/vspec/model/vsstree.py#L21), but when r.get is called with the `qualified_name` with the separator `"."` it returns an error, as the following example:
`anytree.resolver.ResolverError: unknown root node '/Vehicle.Private.Lifecycle'. root is '/Vehicle'.`
because the resolver tries to split the string as nodes with the separator `"/"` by default. 

A simple way to detect this is in these two lines:
https://github.com/GENIVI/vss-tools/blob/14d044699bc358590009aaf1740864c149d3c9be/vspec/__init__.py#L748-L749
The line 784 uses the `qualified_name()` with the `DEFAULT_SEPARATOR` and then try to remove the string `"/Private"` which can only exists if the `DEFAULT_SEPARATOR` is `"/"` or if you call `qualified_name("/") `

oliveirarleo:
Using separator "/ "in detect_and_merge(), since later the name
generated is used by a resolver that gives an error when called
with DEFAULT_SEPARATOR ".".
resolves #82

### DIFF
--- a/vspec/__init__.py
+++ b/vspec/__init__.py
@@ -745,11 +745,11 @@ def detect_and_merge(tree_root: VSSNode, private_root: VSSNode):
         if not private_element.is_private():
             continue
 
-        element_name = "/" + private_element.qualified_name()
+        element_name = "/" + private_element.qualified_name("/")
         candidate_name = element_name.replace("Private/", "")
 
         if not VSSNode.node_exists(tree_root, candidate_name):
-            new_parent_name = "/" + private_element.parent.qualified_name().replace("/Private", "")
+            new_parent_name = "/" + private_element.parent.qualified_name("/").replace("/Private", "")
             new_parent = r.get(tree_root, new_parent_name)
             private_element.parent = new_parent
 


gunnarx (approved):

SebastianSchildt (approved):
I am fine with it

danielwilms:
@gunnarx, please merge

PR #84, opened 2021-07-05
Adding the units for illuminance, frequency, time and pressure

### LINKED ISSUE #76: Change the Unit and VSSDataType to support custom types and units

oliveirarleo:
Adding the units for illuminance, frequency, and time, following the SI notation as described in [International_System_of_Units wiki page](https://en.wikipedia.org/wiki/International_System_of_Units#Units_and_prefixes.)

Adding the following units:
 - Lux as 'lx'
 - Hertz as 'Hz'
 - Millisecond as 'ms'
 - Millibar as 'mbar'
resolves #76

### DIFF
--- a/tests/model/test_contants.py
+++ b/tests/model/test_contants.py
@@ -47,6 +47,7 @@ class TestConstantsNode(unittest.TestCase):
         self.assertEqual(Unit.VOLT, Unit.from_str("V"))
         self.assertEqual(Unit.AMPERE, Unit.from_str("A"))
         self.assertEqual(Unit.SECOND, Unit.from_str("s"))
+        self.assertEqual(Unit.MILLISECOND, Unit.from_str("ms"))
         self.assertEqual(Unit.MINUTE, Unit.from_str("min"))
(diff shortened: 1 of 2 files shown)


SebastianSchildt:
If theres is consensus to merge I will not block, but I feel this is the "most wrong" way to add units. We create a randomly curated list of "allowed" units, some of which are in VSS and some not, and probably the first thing everybody needs to to, once he uses VSS besides the standard catalogue is hacking in more units in his fork  (which I already did, and which also ultimately leads to this PR)

Anyway, my thoughts on this are here already: https://github.com/GENIVI/vss-tools/issues/76#issuecomment-872043539

erikbosch:
Any particular reason to add Millibar as we already have Pascal (Pa) and kiloPascal (kPa)

danielwilms:
Configuration file (e.g. YAML) for units in the VSS repository.

PR #86, opened 2021-07-05
Renaming exclude_private to merge_private

### LINKED ISSUE #85: Variable exclude_private with wrong behavior
In the function `load_tree` 

https://github.com/GENIVI/vss-tools/blob/14d044699bc358590009aaf1740864c149d3c9be/vspec/__init__.py#L158-L166

the variable named `exclude_private` actually merges Private if it's set to `True` since it is passed to `render_tree` later as `merge_private`. This behavior indicates the exact opposite of its meaning, which should not merge Private if it's set to `True`.

oliveirarleo:
The parameter exclude_private in load_tree() is being passed to the
render_tree() function as merge_private. This behavior can confuse
the programmer since both names mean opposite ideas. The real behavior
is that the variable merges Private if it's true, so this commit
corrects this bug just by changing the variable name.

Resolves #85

### DIFF
--- a/vspec/__init__.py
+++ b/vspec/__init__.py
@@ -155,7 +155,7 @@ def convert_yaml_to_list(raw_yaml):
 
 
 
-def load_tree(file_name, include_paths, exclude_private=False, break_on_noncore_attribute=False):
+def load_tree(file_name, include_paths, merge_private=False, break_on_noncore_attribute=False):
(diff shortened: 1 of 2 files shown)


SebastianSchildt (approved):
Works. Output identical. lgtm

gunnarx (approved):

PR #87, opened 2021-07-06
Use of EnumMeta to refactor and improve performance

oliveirarleo:
Use of a new EnumMeta subclass to offer values() and an efficient from_str() in every enumeration that uses it.

### DIFF
--- a/vspec/model/constants.py
+++ b/vspec/model/constants.py
@@ -12,10 +12,36 @@
 #
 
 # noinspection PyPackageRequirements
-from enum import Enum
+from enum import Enum, EnumMeta
+from typing import Sequence, Type, TypeVar
 
 
-class StringStyle(Enum):
+T = TypeVar("T")
+
+
+class EnumMetaWithReverseLookup(EnumMeta):
+    """This class extends EnumMeta and adds:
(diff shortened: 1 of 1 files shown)


barbieri (approved):
thanks!

@gunnarx, @danielwilms , @SebastianSchildt this is the first commit of my #81, it just saves the code duplication and offers an optimized reverse lookup.

SebastianSchildt (approved):
The Python-Fu is strong with this one. Not sure we really _need_ to bring out the metaclass gun for this problem, but in total we have

 - less lines
 - generated output similar to master branch
 
Therefore, I approve if CI run is successful

barbieri:
well, `EnumMeta` is already a metaclass... just adding some more meat to that one :-D

barbieri:
also, not just the 2 points, but it's also more efficient. We're replacing a linear search (on a small subset of items, but still... it's running for every node in the tree) with a constant-time lookup

danielwilms:
@gunnarx, please start build job and merge if it runs through

gunnarx:
Meaning and purpose was explained by @barbieri and it seems OK for me.

Issue #95, opened 2021-08-20
Why there is no vspec2yaml?


crea7or:
Although vspec is almost yaml, it looks like it will be useful to supply vss tree as yaml. It's cleaner than json and better in some small usage patterns regarding c++ libraries that process json/yaml. I'm thinking about developing such script but maybe there some limitation of yaml that I don't know or not understand.

SebastianSchildt:
If you think about a way to generate a "flattened" single-file YAML version out of the existing vspec, that is indeed missing currently. I do think that would not be too hard to create based on the tooling available (check vspec2json or vspec2csv maybe it is as simple as just rewriting the export_node https://github.com/GENIVI/vss-tools/blob/865ea7c06c98f73f5ea724aa19a39ee6e12ae1ba/vspec2json.py#L33 function?)

If you try, pls make a PR :)

gunnarx:
I propose the discussion continues on the now existing PR (only)

PR #96, opened 2021-08-20
vspec2yaml builder

crea7or:
As discussed in the issue #95 I have created YAML from vspec builder.

### DIFF
--- a/vspec2yaml.py
+++ b/vspec2yaml.py
@@ -1,0 +1,123 @@
+#!/usr/bin/env python3
+
+#
+# (c) 2016 Jaguar Land Rover
+# (c) 2021 Robert Bosch GmbH
+# (c) 2021 Pavel Sokolov (pavel.sokolov@gmail.com)
+#
+# All files and artifacts in this repository are licensed under the
+# provisions of the license provided by the LICENSE file in this repository.
+#
+#
+# Convert all vspec input files to a single flat YAML file.
+#
+
+import sys
+import vspec
+import yaml
+import getopt
+from vspec.model.vsstree import VSSNode, VSSType
+
+
+def usage():
+    print(
+        "Usage:", sys.argv[0], "[-I include_dir] ... vspec_file [-s] yaml_file")
+    print("  -s                   Use strict checking: Terminate when non-core attribute is found")
+    print("  -I include_dir       Add include directory to search for included vspec")
+    print("                       files. Can be used multiple timees.")
+    print()
+    print(" vspec_file            The vehicle specification file to parse.")
+    print(" yaml_file             The file to output the YAML objects to.")
+    sys.exit(255)
+
+
+def export_node(yaml_dict, node):
+
+    node_path = node.qualified_name()
+
+    yaml_dict[node_path] = {}
+
+    yaml_dict[node_path]["type"] = str(node.type.value)
+
+    if node.type == VSSType.SENSOR or node.type == VSSType.ACTUATOR or node.type == VSSType.ATTRIBUTE:
+        yaml_dict[node_path]["datatype"] = str(node.data_type.value)
+
+    # many optional attributes are initilized to "" in vsstree.py
+    if node.min != "":
+        yaml_dict[node_path]["min"] = node.min
+    if node.max != "":
+        yaml_dict[node_path]["max"] = node.max
+    if node.enum != "":
+        yaml_dict[node_path]["enum"] = node.enum
+    if node.default_value != "":
+        yaml_dict[node_path]["default"] = node.default_value
+    if node.deprecation != "":
+        yaml_dict[node_path]["deprecation"] = node.deprecation
+
+    # in case of unit or aggregate, the attribute will be missing
+    try:
+        yaml_dict[node_path]["unit"] = str(node.unit.value)
+    except AttributeError:
+        pass
+    try:
+        yaml_dict[node_path]["aggregate"] = node.aggregate
+    except AttributeError:
+        pass
+
+    yaml_dict[node_path]["description"] = node.description
+    yaml_dict[node_path]["uuid"] = node.uuid
+
+    for child in node.children:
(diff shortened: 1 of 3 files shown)


erikbosch:
I would suggest to add it as a subtarget for travis_targets (yes it is still called so) in the Makefile (both here and in vss-repo) so that we it can be run automatically be Continuous Integration

SebastianSchildt:
I tested it manually and it ran, generated a YAML file that is syntactically valid. It should be low risk, as it is using 90% of the same code as the other tools, so I would be fine with merging.

For the curious, this is how a snippet looks like

```yaml

Vehicle.Speed:
  type: sensor
  datatype: float
  unit: km/h
  description: Vehicle speed
  uuid: efe50798638d55fab18ab7d43cc490e9

Vehicle.TravelledDistance:
  type: sensor
  datatype: float
  unit: km
  description: Odometer reading, total distance travelled during the lifetime of the vehicle.
  uuid: 90be9d7b0ac15b75a83027ea3b73b65b

Vehicle.TripMeterReading:
  type: sensor
  datatype: float
  unit: km
  description: Current trip meter reading
  uuid: 81f51ebfe29c591190171d7b96e1c948

Vehicle.AmbientAirTemperature:
  type: sensor
  datatype: float
  unit: celsius
  description: Ambient air temperature outside the vehicle.
  uuid: 2ffcbc2e6ea75dd991e3ae80b29a1d85

Vehicle.IsMoving:
  type: sensor
  datatype: boolean
  description: Indicates whether the vehicle is stationary or moving
  uuid: db69549cc7375e919c2a2883b41cd19c

Vehicle.AverageSpeed:
  type: sensor
  datatype: float
  unit: km/h
  description: Average speed for the current trip
  uuid: 43a489636a665c3abb99b63174eb552b

Vehicle.Acceleration:
  type: branch
  description: Spatial acceleration
  uuid: 6c490e6a798c5abc8f0178ed6deae0a8

Vehicle.Acceleration.Longitudinal:
  type: sensor
  datatype: float
  unit: m/s^2
  description: Vehicle acceleration in X (longitudinal acceleration).
  uuid: 3d511fe7232b5841be311b37f322de5a


```

SebastianSchildt:
Discussed 08/24: Please check and comment if you have an opinion
Open question: Is "vspec2yaml" a good name, as it might be confusing as vspec is also "yaml", maybe this should be called vspecflatten?

Also should it into contrib or as main tool. Please think about it & comment so we can discuss next week

gunnarx (changes requested):
- The two commits can be squashed into one before we merge.
(and resolve comments from @SebastianSchildt above)

gunnarx (changes requested):
- x (execution) flag missing on the vspec2yaml.py file

crea7or:
@gunnarx it looks like these changes are not on my side, yes?

danielwilms:
Summerizing:
1. let's try to find a different name
2. decide on if it goes to contrib or not

gunnarx:
@crea7or wrote:
> @gunnarx it looks like these changes are not on my side, yes?

You need to at minimum set the -x flag and then recommit.
Squashing is theoretically possible to do by the person who is merging, but it is easier if you update the PR.

crea7or:
> 1. let's try to find a different name (`vspecflatten`, `vspec2singleFile`, `vspec2yaml`, `vspec2singleYaml`)
>    favorite in the call was `vspec2singleYaml`. more suggestions?

Imho vspec2yaml is good. If vspec will be changes internally it will loose meaning. All other tools is named with this pattern and users should not think about why this file is named in the different way.
Whatever, I'll change it to any you choose.

SebastianSchildt (approved):
Looks good to me, so my vote is merge.

wrt. the name discussion I am fine with vspec2yaml (argument: vspec is not pure yaml that is pareseable in any "useful" way out of the with standard libs)) as well as with  any of the alternatives (argument: vspecs are written using yaml so "converting" to yaml  might feel confusing for some)

vspec2yaml.py
```diff
@@ -0,0 +1,123 @@
...
+#
+# (c) 2016 Jaguar Land Rover
+# (c) 2021 Robert Bosch GmbH
+# (c) 2021 Pavel Sokolov (pavel.sokolov@gmail.com)
+#
+# All files and artifacts in this repository are licensed under the
+# provisions of the license provided by the LICENSE file in this repository.
+#
+#
+# Concert vspec file into flat YAML
```
gunnarx, line 12: Spelling: Convert

To clarify what it does how about expanding into this:

Convert all vspec input files to a single flat YAML file.

gunnarx (changes requested):

vspec2yaml.py
danielwilms: @crea7or: vspec2yaml is fine, after we voted in the call today. But please check the spelling as mentioned by @gunnarx. Thanks so much for the work and the change!

crea7or: @gunnarx @danielwilms done!

SebastianSchildt (approved):
Tested. (Still) works. Vote merge

vspec2yaml.py
danielwilms: Thanks @crea7or!

gunnarx:
Fixed conflicts resulting from changes that were merged to README before this one (see #97)
@crea7or Warning - since it is possible to force-push to update a PR, it means I now updated @crea7or/master on GitHub (because that is the "from" branch you used).

gunnarx (approved):

gunnarx:
(fyi @erikbosch )
I messed this up slightly because I had a last minute local change that was not committed before merging:

For vspec2yaml the [table in the README](https://github.com/GENIVI/vss-tools/blob/aacda162ef21b0bea055585d8de57598fcf50e9f/README.md) now says "(Community Supported)" instead of simply "Community Supported".   I suppose it is something we can clean up in a later README update.

PR #103, opened 2021-09-23
Adapting datatype support with official datatypes

### LINKED ISSUE #102: vspec2yaml.py：KeyError: 'bytebuffer'
hello，I define a node with the datatype of bytebuffer. But when I parse the vspec file by vspec2yaml.py，it shows the error ---》KeyError: 'bytebuffer'
I checked the ‘vspec/model/constants.py’ that there’s no datatype of ‘bytebuffer’ or ‘byteBuffer’...
But byteBuffer indeed exists in the specification.
How could use the vss tool to parse the vspec file？

erikbosch:
Fixes #102

Addresses two differences compared to documentation in:
https://genivi.github.io/vehicle_signal_specification/rule_set/data_entry/data_types/

- Adding support for byteBuffer - listed in documentation
- Removing "Unix Timestamp" as data type - not included in documentation
  and not used by standardized signals. Even if some signals reference time
  (like Charging.Timer.Time) they instead use exact representaion like uint32.


SebastianSchildt (approved):

danielwilms:
8-bit integer array is in principle equal to byte-buffer, but one could say that the int array supports a sequence of items with the same meaning, where a byte buffer is just a 'blob'.

Any opinion on, if bytebuffer as a type is needed?

crea7or:
It looks like the byte buffer is a duplicate of the array.

erikbosch:
I have no problems removing byteBuffer from documentation instead, i.e. so that they need to use uint8[] instead. For now most people seems to prefer that, so unless someone objects I will in a few days update the PRs by creating a new PR for VSS where byteBuffer is removed from documentation, and changing this PR so that only "Unix Timestamp" is removed

Issue #106, opened 2021-09-28
Github Actions: Optional target tests failing


slawr:
The optional target tests in the Github Actions appear to be failing.

For example:
https://github.com/GENIVI/vss-tools/pull/104/checks#step:7:33

Raising it here as I could not see an open issue for this. Have not investigated further at the time of writing.

erikbosch:
Yes, they do. It is well known and partially intentional. The optional tests (https://github.com/GENIVI/vehicle_signal_specification/blob/master/Makefile#L20) were added to make sure that all tools actually are run as part of CI. But we have quite some tools that are broken or not maintained, and some other that are in beta-stage, and we do not want to enforce that they are up to date with vss and common parts of vss-tools, i.e. it shall be possible to add an improvement to vss or vss-tools without the need to make sure that no regressions occur for contrib/optional tools. I tried to explain the difference between "contrib" and "community supported" in the readme of https://github.com/GENIVI/vss-tools/ recently.

I checked a build log of a recent build.
vspec2c and vspec2ocf are already listed as obsolete and not functional in https://github.com/GENIVI/vss-tools/, so it is expected that they will fail.

The other should to my knowledge be functional as of now, but I noticed in e.g. https://github.com/GENIVI/vss-tools/pull/103/checks?check_run_id=3688390641 that vspec2ttl complains on rdflib missing. @danielwilms - do we need to add rdflib to requirements.txt?

danielwilms:
> The other should to my knowledge be functional as of now, but I noticed in e.g. https://github.com/GENIVI/vss-tools/pull/103/checks?check_run_id=3688390641 that vspec2ttl complains on rdflib missing. @danielwilms - do we need to add rdflib to requirements.txt?

yes, indeed. I'll create a PR.

slawr:
> Yes, they do. It is well known and partially intentional. The optional tests (https://github.com/GENIVI/vehicle_signal_specification/blob/master/Makefile#L20) were added to make sure that all tools actually are run as part of CI. But we have quite some tools that are broken or not maintained, and some other that are in beta-stage, and we do not want to enforce that they are up to date with vss and common parts of vss-tools, i.e. it shall be possible to add an improvement to vss or vss-tools without the need to make sure that no regressions occur for contrib/optional tools. I tried to explain the difference between "contrib" and "community supported" in the readme of https://github.com/GENIVI/vss-tools/ recently.

Sorry for posting and running. Been busy. 
[...]

PR #107, opened 2021-09-30
[vspec2ttl] Add dependencies, align with newest rdflib version

danielwilms:
* added rdflib dependency to requirements.txt
* used newest rdflib version and fixed issues (decode)
* fixed one wrong pointer to owl topDataProperty

related to: #106
fixes: #108 

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -3,3 +3,4 @@ anytree>=2.8.0
 stringcase>=1.2.0
 deprecation>=2.1.0
 pytest>=2.7.2
+rdflib>=6.0.0
--- a/Pipfile
+++ b/Pipfile
@@ -9,6 +9,7 @@ stringcase = ">=1.2.0"
 deprecation = ">=2.1.0"
 pytest = ">=2.7.2"
 PyYAML = ">=5.1"
+rdflib = ">=6.0.0"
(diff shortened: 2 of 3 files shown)


erikbosch (approved):
Looks good to me, ttl build passes!

requirements.txt
```diff
@@ -3,3 +3,4 @@ anytree>=2.8.0
 stringcase>=1.2.0
 deprecation>=2.1.0
 pytest>=2.7.2
+rdflib>=6.0.0
```
barbieri, line 6: please add (and use `Pipfile`) instead, it's just a matter of getting used and remember that, but much more handy than `requirements.txt` as yo don't mess with your system.

You can keep this one, but please consider adding to `Pipfile` as well.

Hint: if you `export PIPENV_VENV_IN_PROJECT=1` in your shell, then it will create a `.venv` foder and you keep all that stuff per project :-) Then when you `pipenv shell` you load that environment, when you exit that, it's all clean. Also `git clean -xfd` will remove everything, including such dependencies.

danielwilms: Thanks for the pointer. Will fix it!

danielwilms:
please merge @erikbosch, @gunnarx

PR #110, opened 2021-10-12
Code generation (Android Vehicle HAL)

BitManu:


danielwilms:
@erikbosch, @gunnarx, please check, if we can declare it as draft

slawr (changes requested):
The typemap.yml parameter is missing from the test script

PR #111, opened 2021-10-13
vspec2yaml: Add option to suppress the uuid in the output file

gunnarx:
Add option to suppress the uuid in the output file


gunnarx:
Tested manually with and without the --no-uuid flag

PR #112, opened 2021-10-13
Add csv2pdf.sh in contrib

gunnarx:
This creates a somewhat readable signal list in PDF from the CSV format.
NOTE: It relies on the pandoc tool being installed.

### DIFF
--- a/README.md
+++ b/README.md
@@ -28,6 +28,7 @@ Examples on tool usage can be found in the [VSS Makefile](https://github.com/COV
 [vspec2proto.py](contrib/vspec2protobuf.py) | Parses and expands a VSS and generates a Protobuf specification | Contrib (Beta)  | - |
 [vspec2ttl.py](contrib/vspec2ttl/vspec2ttl.py) | Parses and expands a VSS and generates a TTL specification | Contrib  | - |
 [vspec2graphql.py](contrib/vspec2graphql.py) | Parses and expands a VSS and generates a GraphQL specification | Contrib  | - |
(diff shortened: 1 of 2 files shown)


erikbosch:
Would it be possible to get this running in our CI (possibly as a separate PR if only a change in the vss makefile is needed). But maybe https://github.com/GENIVI/vss-tools/blob/master/.github/workflows/buildcheck.yml needs to be changed as well to install pandoc first?

gunnarx:
> Would it be possible to get this running in our CI (possibly as a separate PR if only a change in the vss makefile is needed). But maybe https://github.com/GENIVI/vss-tools/blob/master/.github/workflows/buildcheck.yml needs to be changed as well to install pandoc first?

Should be possible by installing pandoc as part of the workflow.

I chose to put it in the optional group of the Makefile (-> https://github.com/GENIVI/vehicle_signal_specification/pull/344) but it is simple enough that it could be nice to require it to be successful, i.e. be part of a successful test.  (Next step would be to publish the file as well).  I am not sure but I am guessing optional group is still correct, since it needs special support anyway if it is going to run in CI.

gunnarx:
... so I interpret this as a request to add also the corresponding GH actions workflow before accepting the PR?

erikbosch:
Should we possibly move the csv2pdf.sh file to a separate directory, e.g. contrib/csv2pdf/csv2pdf.sh, and in that directory add a very short README.md (linked from main readme) on how you should install the required tool and an example on how you first must generate csv and then run this tool.

danielwilms:
@gunnarx, please check

gunnarx:
Closed for reasons presented in COVESA/vehicle_signal_specification#344

PR #118, opened 2021-11-16
Fix strict checking

SebastianSchildt:
We had a regression with strict attribute checking in the main parser code.

When using the -s option in any  vspec2x tool supporting the feature, we expect them to terminate when an attrinbute that is not defined in the standard catalogue is detected. However, currently only a warning is printed, as the desire to abort on an unknown attribute was not propagated when processing sub tress.

This changes fixes that. Just look at the diff, that is easier than explaining this :) It is used in VSSNode creation (see line 67 in vsstree.py)


erikbosch (approved):

danielwilms:
@erikbosch, please merge.

PR #120, opened 2021-11-17
Adding support for comments in node descriptions

erikbosch:
Related to: 

https://github.com/COVESA/vehicle_signal_specification/issues/364
https://github.com/COVESA/vehicle_signal_specification/pull/367

This PR cover two parts:

- General model support for `comment`. I.e. tool will recognize `comment` as a valid but optional item and save it in the internal model.
- Support to output `comment` for CSV and JSON tools. We can discuss how we want to handle the tools. Try to update all or none or just some?

### DIFF
--- a/vspec2csv.py
+++ b/vspec2csv.py
@@ -45,7 +45,7 @@ def format_csv_line(*csv_fields):
 
 #Write the header line
 def print_csv_header(file):
-    file.write(format_csv_line("Signal","Type","DataType","Deprecated","Complex","Unit","Min","Max","Desc","Enum","Id","Instance"))
+    file.write(format_csv_line("Signal","Type","DataType","Deprecated","Complex","Unit","Min","Max","Desc","Comment","Enum","Id","Instance"))
 
 #Write the data lines
 def print_csv_content(file, tree):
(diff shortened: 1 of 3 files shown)


SebastianSchildt:
For the vspec2XX tools, I would guess, we want a parameter (like uuid generation in yaml and json) defining whether you want to generate the comment or not

Other than that, it looks like a straightforward implementation.

vspec2csv.py
```diff
@@ -57,10 +57,10 @@ def print_csv_content(file, tree):
         if tree_node.instances:
             for instance in tree_node.instances:
                 file.write(format_csv_line(
-                    tree_node.qualified_name('.'),tree_node.type.value,data_type_str,tree_node.deprecation,"true",unit_str,tree_node.min,tree_node.max,tree_node.description,tree_node.enum,tree_node.uuid,instance))
+                    tree_node.qualified_name('.'),tree_node.type.value,data_type_str,tree_node.deprecation,"true",unit_str,tree_node.min,tree_node.max,tree_node.description,tree_node.comment,tree_node.enum,tree_node.uuid,instance))
```
erikbosch, line 60: Any opinion on where to add comment in *.csv? As of now this code add comment after description which could be logically correct, but it could cause backward compatibility issues if someone use *.csv and rely on absolute position for "enum"/"id". As ana lternative we could add "comment" last.

(By the way - I assume the "instance" code here is obsolete - we have already expanded instanced before we print *.csv, so this line will never be executed)

gunnarx (approved):

erikbosch:
Meeting notes:
- Ok to merge, later make generation optional

PR #122, opened 2021-11-19
Use argparse for binary  generation.

SebastianSchildt:
Using argparse for vspec2binary.

It should work, please (in case CI will be green) test @UlfBj  or @gunnarx , but I need to guess a little butt as the original "usage" was form frank tool, and the actual code contains a parameter "-v" for vss_version, that was then never used (this I did not recreate that parameter)

Also for the future: This code still uses the legacy parser "load" instead of load_tree, so once we remove that, this code breaks. But I understood to little to change the traverse_tree function.

I suggest checking whether this works "as before" first, than merge it and open another issue

### DIFF
--- a/vspec2binary.py
+++ b/vspec2binary.py
@@ -16,22 +16,16 @@ import sys
 import os
 import vspec
 import json
-import getopt
+import argparse
 import ctypes
-
-def usage():
-    print(("Usage:", sys.argv[0], "[-I include_dir] ... vspec_file franca_file"))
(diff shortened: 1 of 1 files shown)


SebastianSchildt:
I also found CI starts this without "-I" parameter, I think the tool "guesses" includes or just uses cwd.

erikbosch:
Meeting: @UlfBj to test. If test pass ready to merge

UlfBj:
@erikbosch , @SebastianSchildt I have tested it, looks good.

Issue #105, opened 2021-09-28
change argument parsing for other tools than json


danielwilms:
@SebastianSchildt, please have a look, like done in #100

SebastianSchildt:
Can be closed @erikbosch

PR #123, opened 2021-11-23
Use argparse for argument parsing in vspec2franca

SebastianSchildt:
Using argparse for arguments parsing, removing the outdated getopt code.

See also #105

### DIFF
--- a/vspec2franca.py
+++ b/vspec2franca.py
@@ -12,17 +12,7 @@
 
 import sys
 import vspec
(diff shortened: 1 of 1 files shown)


erikbosch:
Meeting: Ready to merge

gunnarx:
I ran it.  Output looks as expected (but that output is/was not so useful in my opinion, hence #64

PR #124, opened 2021-11-24
Removing obsolete entries from VSS-tools

erikbosch:
"stream", "bytebuffer", "element" and "rbranch" are obsolete entities.

### DIFF
--- a/contrib/vspec2c.py
+++ b/contrib/vspec2c.py
@@ -39,7 +39,6 @@ type_map =  {
     "bool": "VSS_BOOLEAN",
     "boolean": "VSS_BOOLEAN",
     "string": "VSS_STRING",
-    "stream": "VSS_STREAM",
     "na": "VSS_NA"
 }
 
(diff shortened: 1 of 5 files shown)


vspec2csv.py
```diff
@@ -40,13 +40,8 @@ def print_csv_content(file, tree):
         data_type_str = tree_node.data_type.value if tree_node.has_data_type() else ""
         unit_str = tree_node.unit.value if tree_node.has_unit() else ""
 
-        if tree_node.instances:
```
erikbosch, line 43: Tree is flattened before we write *.csv so there will never be any instances. I verified that we generate the same number of lines after this change as before

gunnarx: I like this change but would have preferred it in a separate commit.

gunnarx (approved):

vspec2csv.py
erikbosch: Will put the csv part in a separate commit

PR #127, opened 2021-12-14
vsso2ttl proposals at Bosch

user-cjw:
Main change points:

- Use instances instead of classes for different VehicleComponents
- Introduce Enumeration and Numeric VehicleProperties

### DIFF
--- a/contrib/vspec2ttl/vspec2ttl.py
+++ b/contrib/vspec2ttl/vspec2ttl.py
@@ -107,6 +107,41 @@ def setup_graph():
     g.add((partOfVehicleComponent,RDFS.range, VssoConcepts.VEHICLE_COMP.uri))
     g.add((partOfVehicleComponent,RDFS.label, Literal(VssoConcepts.PART_OF_VEH_COMP.value,lang="en")))
 
+    vehiclePropertyName = VssoConcepts.VEHICLE_PROP_NAME.uri
+    g.add((vehiclePropertyName, RDF.type, OWL.DatatypeProperty))
+    g.add((vehiclePropertyName, RDFS.subPropertyOf, OWL.topDataProperty))
+    g.add((vehiclePropertyName, RDFS.label, Literal(VssoConcepts.VEHICLE_PROP_NAME.value, lang="en")))
+    g.add((vehiclePropertyName, RDFS.domain, VssoConcepts.VEHICLE_PROP.uri))
+    g.add((vehiclePropertyName, RDFS.range, XSD.string))
+
+    vehicleComponentName = VssoConcepts.VEHICLE_COMP_NAME.uri
+    g.add((vehicleComponentName, RDF.type, OWL.DatatypeProperty))
+    g.add((vehicleComponentName, RDFS.subPropertyOf, OWL.topDataProperty))
+    g.add((vehicleComponentName, RDFS.label, Literal(VssoConcepts.VEHICLE_COMP_NAME.value, lang="en")))
+    g.add((vehicleComponentName, RDFS.domain, VssoConcepts.VEHICLE_COMP.uri))
+    g.add((vehicleComponentName, RDFS.range, XSD.string))
+
+    minValue = VssoConcepts.MIN.uri
+    g.add((minValue, RDF.type, OWL.DatatypeProperty))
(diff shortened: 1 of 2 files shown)


danielwilms:
@felixloesch, thanks for the update. I'll go into the details later today, but I'd have one conceptional question. 

I'd really like to keep the sub-class structure (at least for now), as we have now the proposal for the split in VSSo, like `vsso-core` and `vsso` for the generated sub-classes. could we keep that structure for now and take the discussion for/against the instances vs. subclasses as a first discussion point in the working group?

felix-loesch:
We can definitely discuss this. I will prepare some slides for discussion including our motivation for that change for the workshops in January.

contrib/vspec2ttl/vspec2ttl.py
```diff
@@ -55,168 +55,263 @@ def usage():
...
-    g.add((hasDynProp, RDFS.domain, VssoConcepts.VEHICLE.uri))
-    g.add((hasDynProp, RDFS.range, VssoConcepts.VEHICLE_PROP.uri))
-    g.add((hasDynProp, RDFS.label, Literal(VssoConcepts.HAS_SIGNAL.value,lang="en")))
-
+    g.add((belongsTo, RDFS.subPropertyOf, OWL.topObjectProperty))
+    g.add((belongsTo, RDFS.label, Literal(VssoConcepts.BELONGS_TO.value, lang="en")))
+    g.add((belongsTo, RDFS.range, VssoConcepts.VEHICLE_COMP.uri))
+    g.add((belongsTo, RDFS.domain, VssoConcepts.VEHICLE_PROP.uri))
+
+    # holdsValue = VssoConcepts.HOLDS_VALUE.uri
```
erikbosch, line 72: For commented out code I would prefer a comment explaining why it is commented (and not removed). 

(I assume that we in general do not want commented out code, but if it is just temporary it might be OK)

erikbosch:
Noticed a build problem.  Most likely because a signal using the unit "h" was recently added, but "h" is not included in vssotypes.py

Hour listed as valid unit in:
https://covesa.github.io/vehicle_signal_specification/rule_set/data_entry/data_unit_types/
https://github.com/COVESA/vss-tools/blob/master/vspec/config.yaml

```
  File "./vss-tools/contrib/vspec2ttl/vspec2ttl.py", line 419, in <module>
    print_ttl_content(ttl_out, tree)
  File "./vss-tools/contrib/vspec2ttl/vspec2ttl.py", line 349, in print_ttl_content
    graph.add((node, VssoConcepts.UNIT.uri, DataUnits[tree_node.unit.value]))
KeyError: 'h'
make: *** [Makefile:56: ttl] Error 1
```

felix-loesch:
OK. I think it was missing before our commits as well. Should be added to vssotypes.py.

erikbosch:
> OK. I think it was missing before our commits as well. Should be added to vssotypes.py.

Yes, it just surfaced up now. Could be fixed in a separate PR but then you would anyway get a merge conflict, so likely easier to fix as part of this PR, or wait with fixing it until this PR has been merged to avoid merge conflicts.

contrib/vspec2ttl/vspec2ttl.py
user-cjw: Those comments means need discussion. 

We are changing the model used for generating .ttl file, the part for abstraction (i.e. the content within VSS) are ready in our proposal, but the part for describing instance (the single produced vehicle component part installed in vehicle) is not ready yet.

The commented part is originally used for instances, and now they have conflicts with our new model design. If the new model design works well, we will further adapt the commented part to fit the new design.

danielwilms:
d258d1b : opened as an [issue](https://github.com/w3c/vsso/issues/22) in w3c/vsso
88ab4c4 : part of #141
4fc8709, 8065958, 870ba40, 417db9b are valid extensions to the core ontology. I will create PRs/issues on those.

danielwilms:
@danielwilms: check if all issues are there, so PR can be closed.

danielwilms:
@erikbosch: extended the issues in the vsso github and created a new one. All mapped now and PR can be closed.

erikbosch:
Closing on request from @danielwilms

Issue #114, opened 2021-11-09
Move to Pipfile


danielwilms:
* remove `requirements.txt`
* check that build still works

danielwilms:
@danielwilms will look into it

PR #131, opened 2022-01-06
[BUILD] Move from `requirements.txt` to `Pipfile`

danielwilms:
Until now dependencies were defined in two separate files
(requirements.txt and Pipfile). As decided in #114 we'll
move to `Pipfile`.

* removed `requirements.txt` to avoid confusion
* removed section from `README.md`
* changed `buildcheck.yml` to use `pipenv`

Fixes: #114

### DIFF
--- a/.github/workflows/buildcheck.yml
+++ b/.github/workflows/buildcheck.yml
@@ -14,7 +14,7 @@ jobs:
           # required in `Pipfile`.
           curl https://pyenv.run | bash
           export PATH="$HOME/.pyenv/bin:$PATH"
-          pyenv install 3.8.5
+          pyenv install 3.8.12
       - name: Install dependencies
         run: |
           python -V
--- a/.github/workflows/buildcheck.yml
+++ b/.github/workflows/buildcheck.yml
@@ -8,16 +8,22 @@ jobs:
     steps:
       - uses: actions/checkout@v2
 
-      - name: Setup Python
-        uses: actions/setup-python@v1
-        with:
-          python-version: 3.8
-
+      - name: Install pyenv
+        run: |
+          # Needed for using exact python version, which is
+          # required in `Pipfile`.
+          curl https://pyenv.run | bash
+          export PATH="$HOME/.pyenv/bin:$PATH"
+          pyenv install 3.8.5
       - name: Install dependencies
         run: |
           python -V
-          pip install -r requirements.txt
-          python setup.py -q install
+          python -m pip --quiet --no-input install --upgrade pip 
+          python -m pip --quiet --no-input install --upgrade pipenv wheel
+          pipenv install
+          pipenv run python setup.py -q install
+        env:
+          PIPENV_VENV_IN_PROJECT: 1
 
       - name: Prepare vss-tools inside VSS
         run: |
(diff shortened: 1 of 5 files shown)


.github/workflows/buildcheck.yml
```diff
@@ -12,12 +12,21 @@ jobs:
...
         with:
           python-version: 3.8
-
+      - name: Install pyenv
+        run: |
+          # Needed for using exact python version, which is
+          # required in `Pipfile`.
+          curl https://pyenv.run | bash
+          export PATH="$HOME/.pyenv/bin:$PATH"
+          pyenv install 3.8.5
```
barbieri, line 21: you may avoid this one since we already use `actions/setup-python` above, see https://github.com/actions/setup-python#specifying-a-python-version

barbieri: also, would be nice to update to latest security release 3.8.12 https://www.python.org/downloads/release/python-3812/ update that in the Pipfile and all, generate a new `Pipfile.lock` (could be in another PR).

.github/workflows/buildcheck.yml
```diff
@@ -12,12 +12,21 @@ jobs:
...
       - name: Install dependencies
         run: |
           python -V
-          pip install -r requirements.txt
-          python setup.py -q install
+          pip install pipenv wheel
+          pipenv install
```
barbieri, line 26: you may use
```sh
python -m pip --quiet --no-input install --upgrade pip 
python -m pip --quiet --no-input install --upgrade pipenv wheel
```

The important part would be `--upgrade` and also update `pip` itself. I've found some issues using outdated `pip` in other projects

barbieri (approved):
looks good, left some improvement suggestions

.github/workflows/buildcheck.yml
danielwilms: thought so, too. but then we have to stick to the official versions, which are installable in the distro. With pyenv we're more flexible. But maybe we can remove the setup-python action.

.github/workflows/buildcheck.yml
danielwilms: thanks for the pointer. will fix it!

.github/workflows/buildcheck.yml
barbieri: with setup-python you don't need the python version to be in the distro, it's pretty much like `pyenv` (not sure if it's used underneath).

danielwilms: In case of 3.8.5 it didn't find it and just listed the ones currently released and active as official packages for the distro.

But that's not a hard requirement to meet. Just thought that we'd be more flexible that way. Either way is fine for me.

As for changing to 3.8.12, it's on my list but thought I'd put that in a separate PR.

danielwilms: I left pyenv and removed setup-python action.

.github/workflows/buildcheck.yml
danielwilms: Fixed in updated commit

danielwilms:
@barbieri, I changed the python version in a separate commit, but in the same PR. Please have a look, if now everything is resolved.

Pipfile
```diff
@@ -15,4 +15,4 @@ vss-tools = {editable = true, path = "."}
 [dev-packages]
 
 [requires]
-python_version = "3.8.5"
+python_version = "3.8.12"
```
erikbosch, line 18: In the README.md file we still mention 3.7 and 3.8.5 in some places, should they be replaced with 3.8.12 as well?

danielwilms: Thanks @erikbosch. Fixed it with the latest update.

barbieri (approved):

SebastianSchildt (approved):
Looks good to me. And much nicer to run the same in CI we expect users to run 👍

danielwilms:
meeting 11/01/2022: approved to be merged @erikbosch

Issue #133, opened 2022-01-11
Millisecond data unit type is documented in VSS doc but missing in vss-tools tooling


slawr:
As requested in today's W3C VSS tooling call I am creating an issue to add milliseconds as a data unit type to config.yaml.

**Background**
My data model defined in vpsec requires a millisecond data unit type.

**Issue**
The VSS documentation lists [millisecond ](https://github.com/COVESA/vehicle_signal_specification/blob/965d035904a124c4b9f4a14c990dcd78e3b1b83d/docs-gen/content/rule_set/data_entry/data_unit_types.md?plain=1#L31) as an available data unit type, but it is not implemented in vss-tools [vspec/config.yaml](https://github.com/COVESA/vss-tools/blob/master/vspec/config.yaml)

**Request**
The my_config.yaml support added in #104 would be a valid workaround to the issue by allowing the vss-tools list to be extended in a private way. However this is the only data unit type I am missing and having to ship a private yaml just for it would complicate adopting the data model to the VSS eco-system. The code will also be made public as OSS. It is also a simple SI unit type. I would therefore like to add it to the standard list.

I am happy to create a PR to add it to the code, but I have been asked to raise this issue as a first step.

oliveirarleo:
Okay, so a good idea is to guarantee that all units specified in the VSS documentation are supported. Doing it would solve your problem, but also problems like this in the future.

erikbosch:
I have a PR open (https://github.com/COVESA/vss-tools/pull/134) to add Watt (listed in doc, currently not in config), I could add millisecond in that PR as well

slawr:
> I have a PR open (#134) to add Watt (listed in doc, currently not in config), I could add millisecond in that PR as well

Hi Erik, PR from you, me I really don't mind if it gets it in. Looking at related commits it looks like it was a straight forward change. I have not done the coding yet so happy for you to add it to your PR. Do you want to go ahead then? We are probably approaching the point of collectively spending more time discussing then it would take to add it :)

slawr:
@erikbosch I see you have incorporated milliseconds into #134 so I'll close this. Thanks for the change.

PR #134, opened 2022-01-13
Adding support for missing units

erikbosch:
Background:

Not all units in described in https://github.com/COVESA/vehicle_signal_specification/blob/master/docs-gen/content/rule_set/data_entry/data_unit_types.md are currently supported by tools.

Notes/Remarks:

- The documentation list "lat" and "lon". They are not included by this PR. Currenly in VSS we use "degrees" to represent lat/lon, and refer to WGS84 to specifiy how it shall be interpreted. So do we really see a benefit in having dedicated lat/lon units, which practically just would be a speciailization of degrees (degrees with well defined range and meaning)?
- There are still some inconsistencies  related to domain names between config.yaml and documentation. Suggestion to change here first, then update doc accordingly
- Config.yaml contain some units not included in documentation, like starts and unix timestamp. 

Related to:

https://github.com/COVESA/vss-tools/issues/133
https://github.com/COVESA/vehicle_signal_specification/issues/396

### DIFF
--- a/vspec/config.yaml
+++ b/vspec/config.yaml
@@ -23,13 +23,29 @@ units:
     label: kilometer per hour
     description: Speed measured in kilometers per hours
     domain: speed
+  m/s:
+    label: meters per second
+    description: Speed measured in meters per second
+    domain: speed
   m/s^2:
     label: meters per second squared
     description: Acceleration measured in meters per second squared
     domain: acceleration
+  cm/s^2:
+    label: centimeters per second squared
+    description: Acceleration measured in centimeters per second squared
+    domain: acceleration
+  ml:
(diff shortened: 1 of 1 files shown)


oliveirarleo (approved):

SebastianSchildt (approved):
I am generally not sure we should add units to the default distribution that are currently not used in the standard catalogue (because: Where to stop?), 

Considering our currently rather inconsistent approach, I still vote merge. At least one step towards consistency of docs/code, and improved descriptions in the config.yaml  :+1:

danielwilms (approved):
agreed to merge 2022/01/18

Issue #75, opened 2021-06-29
Package vss-tools and publish to PyPi index


gunnarx:
Discussion around the topic ^^

barbieri:
https://pypi.org/project/vss/ exists but is not this package, so I'd like to propose us to register the official project and publish the package (since bb593541cd35ef19c1b992bf84f7e2e5d29d0464 it's doable).

1. Register your user and project at https://pypi.org/
2. generate the packages (ie: `setup.py bdist_egg bdist_wheel`)
3. upload to pypi using [twine](https://pypi.org/project/twine/)

```sh
rm -fr dist/ build
./setup.py egg_info -Db '' sdist bdist_wheel bdist_egg

twine check dist/*.tar.gz || exit 1
twine upload dist/*
```

barbieri:
@danielwilms , @SebastianSchildt , @erikbosch care to comment? :-)

SebastianSchildt:
I can not do this, but I think it would be a nice thing, if people can just "pip install" VSS tools. 

We are doing this for the kuksa.val client libraries already : https://pypi.org/project/kuksa-viss-client/ and while it is not our "make or break" feature it is neat 😀 

With Github Actions, it should even be possible to automate this for tagged releases. So my position would be as @barbieri  is knowledgable in these things, please feel encouraged to do this 😁 

(I think @gunnarx  should register at pypi (so this remains in "Genivi's hands", and then just throw they keys over to @barbieri 🔑 )

barbieri:
I don't need/want the keys :-) For me, GitHub actions that automate the process, such as when we create a tag or some branch name (ie: release/<XXX>) then it runs the release process and uses twine to upload to pypi.

gunnarx:
OK. I don't know the pypi process, but I would be willing to find out.

barbieri:
@gunnarx it's this, based on https://packaging.python.org/tutorials/packaging-projects/#uploading-the-distribution-archives

1. install [twine](https://pypi.org/project/twine/) `python3 -m pip install --upgrade twine `
2. create your user (personal): https://pypi.org/account/register/ (recommended to add 2fa and all)
3. to automate stuff you may issue an API Token at https://pypi.org/manage/account/
4. vss-tool's `setup.py` requires a git tag, so create one that will be used as version (ex: `0.1.0`)
[...]

PR #138, opened 2022-01-20
Fixing versioning at setup.py

oliveirarleo:
This may be the error that was making it not possible to install without the editable part issue https://github.com/COVESA/vss-tools/issues/75.

Please look at versioning at

https://packaging.python.org/en/latest/guides/distributing-packages-using-setuptools/#version
https://packaging.python.org/en/latest/guides/distributing-packages-using-setuptools/#pre-release-versioning

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -9,7 +9,7 @@ version = r.stdout.rstrip().decode('UTF-8')
 # ...otherwise, use abbreviated git commit hash
 if version == '':
     r = subprocess.run(['git', 'rev-parse', '--short=8', 'HEAD'], stdout=subprocess.PIPE)
-    version = r.stdout.rstrip().decode('UTF-8')
+    version ='3.0.dev+' +  r.stdout.rstrip().decode('UTF-8')
 
 setup(
     name='vss-tools',
--- a/Pipfile.lock
+++ b/Pipfile.lock
@@ -80,11 +80,11 @@
         },
         "pyparsing": {
             "hashes": [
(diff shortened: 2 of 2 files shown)


setup.py
```diff
@@ -4,16 +4,16 @@
...
-if version == '':
+if git_version == '':
     r = subprocess.run(['git', 'rev-parse', '--short=8', 'HEAD'], stdout=subprocess.PIPE)
-    version = r.stdout.rstrip().decode('UTF-8')
+    git_version = r.stdout.rstrip().decode('UTF-8')
 
 setup(
     name='vss-tools',
-    version=version,
+    version='2.2+' + git_version,
```
erikbosch, line 16: I assume this is intended to correspond to VSS version. Think it would be go to document that it needs to be updated as part of release, possibly by mentioning it in https://github.com/COVESA/vehicle_signal_specification/wiki/Release-Checklist.

Another point, here we add "2.2" as prefix. Shall that be done even if we found a tag on line 7?

oliveirarleo: Yes, the intention is to match which vss version this tool is supporting.

> Another point, here we add "2.2" as prefix. Shall that be done even if we found a tag on line 7?

In the doc you can verify that the pattern is `<public version identifier>+<local version label>`, so this label is being kept as it were before, the only thing is that the `<public version identifier>` should be respected, so the package can be installed without the warning.

danielwilms:
please change to 3.0<something-dev>

oliveirarleo:
Done! Also I updated so if the tag exists, then the versioning is just the tag.

danielwilms (approved):
ready to merge, decided on 2022-02-01

PR #146, opened 2022-02-17
Unify vspec2X tooling

### LINKED ISSUE #140: combine compiler in one python file
1. first with anytree based tools (csv, json, yaml)
2. look at binary franca how to move to anytree

SebastianSchildt:
Removes redundant code, unifying codebases for common exporters

Will fix #140 

Everything is "vspec2x" now and you just need to add a `--format` parameter. 

There are simple wrappers for the old vspec2json, yaml, csv tools so the CI runs "as is", but in the long run I would vote to remove them, e.g. now you could do 

```
./vspec2json --format yaml ...
```
And the output would be yaml.


vspec2json.py
```diff
@@ -1,118 +1,17 @@
...
-        print("Loading vspec...")
-        tree = vspec.load_tree(
-            args.vspec_file, include_dirs, merge_private=False, break_on_noncore_attribute=abort_on_non_core_attribute, break_on_name_style_violation=abort_on_namestyle)
-        print("Recursing tree and creating JSON...")
-        export_json(json_out, tree, generate_uuids)
-        print("All done.")
-    except vspec.VSpecError as e:
-        print(f"Error: {e}")
-        sys.exit(255)
+    vspec2x.main(["--format", "json"]+sys.argv[1:])
```
erikbosch, line 17: "No newline at end of file" Does it matter?

vssexporters/vss2json.py
```diff
@@ -0,0 +1,85 @@
...
+
+    for child in node.children:
+        export_node(json_dict[node.name]["children"], child, generate_uuid)
+
+def export(config: argparse.Namespace, root: VSSNode):
+    print("Generating JSON output...")
+    json_dict = {}
+    export_node(json_dict, root, not config.no_uuid)
+    outfile=open(config.output_file,'w')
+    if config.json_pretty:
```
erikbosch, line 74: I get a behavioral difference when doing `make json` . Previously we got "pretty" by default, now we get "ugly"

vspec2json.py
SebastianSchildt: No, but it is bad style :) Fixed.

vssexporters/vss2json.py
SebastianSchildt: That is true and was sort of by intention: I assume most often JSON is generated to be consumed by another machine, so minimised format is more efficient.
I sort of was thinking about generating both versions in CI and maybe even ship both with release (like many JS libs, who have a version for "debugging" and a minimised version for production)

But before adapting CI, we should make a decision, whether we want to keep the 2csv/2yaml/2json aliases, or just use vspec2x directly

danielwilms:
This is a nice one! Works for me!

gunnarx:
My opinion:  For now at least, keep the 2csv/2yaml/2json aliases

PR #149, opened 2022-03-15
Graphql tooling

danielwilms:
Alternative graphql schema generation. Discussion is open, which one to use.


contrib/vspec2graphql/README.md
```diff
@@ -0,0 +1,57 @@
...
+
+Leaves look like this:
+```
+"""Vehicle brand or manufacturer"""
+type Vehicle_VehicleIdentification_Brand {
+  """Value: Vehicle brand or manufacturer"""
+  value: String
+
+  """Source system: Vehicle brand or manufacturer"""
+  source: String
```
erikbosch, line 41: A few lines describing the intended usage of `source` and `channel` would be helpful.

contrib/vspec2graphql/traverse_tree.py
```diff
@@ -0,0 +1,113 @@
...
+    VSSDataType.UINT16: GraphQLInt,
+    VSSDataType.UINT16_ARRAY: GraphQLList(GraphQLInt),
+    VSSDataType.INT32: GraphQLInt,
+    VSSDataType.INT32_ARRAY: GraphQLList(GraphQLInt),
+    VSSDataType.UINT32: GraphQLInt,
+    VSSDataType.UINT32_ARRAY: GraphQLList(GraphQLInt),
+    VSSDataType.INT64: GraphQLFloat,
+    VSSDataType.INT64_ARRAY: GraphQLList(GraphQLFloat),
+    VSSDataType.UINT64: GraphQLFloat,
+    VSSDataType.UINT64_ARRAY: GraphQLList(GraphQLFloat),
```
erikbosch, line 48: Is it so that a GraphQLInt only can handle up to int32, and that is the reason why we use float here?

contrib/vspec2graphql/traverse_tree.py
```diff
@@ -0,0 +1,113 @@
...
+    VSSDataType.FLOAT: GraphQLFloat,
+    VSSDataType.FLOAT_ARRAY: GraphQLList(GraphQLFloat),
+    VSSDataType.DOUBLE: GraphQLFloat,
+    VSSDataType.DOUBLE_ARRAY: GraphQLList(GraphQLFloat),
```
erikbosch, line 52: Does GraphQL say anything about how many significant digits a float contain. I.e. is there a risk that we will loose some precision using GraphQLFloat for VSS Double?

contrib/vspec2graphql/README.md
adobekan: Default scalar types in GraphQL:

    Int: A signed 32‐bit integer.
    Float: A signed double-precision floating-point value.
    String: A UTF‐8 character sequence.
    Boolean: true or false.
    ID: The ID scalar type represents a unique identifier, often used to refetch an object or as the key for a cache.

contrib/vspec2graphql/traverse_tree.py
adobekan: Default scalar types in GraphQL:

Int: A signed 32‐bit integer.
Float: A signed double-precision floating-point value.
String: A UTF‐8 character sequence.
Boolean: true or false.
ID: The ID scalar type represents a unique identifier, often used to refetch an object or as the key for a cache.

contrib/vspec2graphql/traverse_tree.py
adobekan: Double-precision, should be fine.

contrib/vspec2graphql/README.md
erikbosch: My original thought was to just have a few lines describing if the contents/use of the fields are implementation specific or preferably shall be standardized.  Similar for timestamp - shall a specific format (like ISO 8601) be used or is it that implementation specific.

Maybe just add something similar to below to the readme:


_`source` and `channel` are free text fields which can be used to include information on the origin of the data. Content is implementation specific. `timestamp` shall be the timestamp of the observation formatted according to ISO 8601 with UTC time zone._

contrib/vspec2graphql/traverse_tree.py
erikbosch: But if Int is signed 32-bit, then we need float for uint32, or?

contrib/vspec2graphql/README.md
adobekan: Thanks

contrib/vspec2graphql/traverse_tree.py
danielwilms: changed graphql type.

PR #150, opened 2022-03-17
Binary tool refactored to use Exporter class support.

UlfBj:


vspec2x.py
```diff
@@ -31,6 +31,7 @@ def export(config: argparse.Namespace, root: VSSNode):
     json = vss2json
     csv = vss2csv
     yaml = vss2yaml
+    binary = vss2binary
```
erikbosch, line 34: Is it this that gives build problem - that we now need the generated binary also when generating e.g. csv or json?

(Which we likely do not want)

SebastianSchildt: Disclaimer: Have not tested it, but it might be possible to move this assignments in vss2binary.py

```python
dllName = "../binary/binarytool.so"
 dllAbsPath = os.path.dirname(os.path.abspath(__file__)) + os.path.sep + dllName
 _cbinary = ctypes.CDLL(dllAbsPath)
```

inside the call to export()?

In that case it would only explode IF you want to convert 2binary and forgot to compile. (which seems fine)

danielwilms:
* fix tests and scripts 
* squash commits

binary/go_parser/go.mod
```diff
@@ -3,6 +3,6 @@ module github.com/COVESA/vss-tools/binary/go_parser
 go 1.14
 
 require (
-	github.com/COVESA/vss-tools/binary/go_parser/datamodel latest
-	github.com/COVESA/vss-tools/binary/go_parser/parserlib latest
+	github.com/COVESA/vss-tools/binary/go_parser/datamodel v0.0.0-20220112071239-487b135d7f74
```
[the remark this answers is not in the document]
UlfBj: It is not likely to change very often. The Go build system did not accept the parameter latest anyway.

danielwilms:
can be closed in favor #152

PR #151, opened 2022-03-18
Add Ampere hours as unit

erikbosch:
We have identified some use-cases that we might contribute to VSS where Ampere hours (Ah) is needed. `Ah` is e.g. a common unit to describe capacity of supply voltage batteries.

If accepted, I will create a corresponding PR for VSS to update documentation.

### DIFF
--- a/vspec/config.yaml
+++ b/vspec/config.yaml
@@ -95,6 +95,10 @@ units:
     label: ampere
     description: Electric current measured in amperes
     domain: electric current
+  Ah:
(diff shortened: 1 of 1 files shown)


danielwilms:
@ravichalanti, we will open an issue to work on the unit descriptions, but as it's in line with the other descriptions, we'll merge this one for now.

@erikbosch, could you merge please 

decision 29/03/2 to merge

PR #155, opened 2022-04-27
LAYERS: Ability to merge standard VSS models

SebastianSchildt:
This will enable layers in vss-tools

wave-1 will only support merging "standard" models, adding support for non-standard metadata will be another PR

In all `vspec2x`based tooling, you can add an arbitrary number of overlays that will be applied in the order they appear on the commandline.

The instance expansion needed to be rewritten (to operate on anytree, in the old code it was applied to early, so that overwriting instances would not have ben possible)

Without overlays, it is expected the output of vspec2x is similar to the previous version

## State after this is merged
 - VSS tree can be "extended" via overlays
 - no support for deletion
 - instances can be overwritten
 - no warnings yet hen overwriting core attributes, i.e. you can change a datatype in the VSS standard catalogue using an overlay, and get no warning that this might not be clever
 - no support for arbitrary metadata (accuracy, access control etc.) yet.
 

For sanity check

```bash
sebastian@armbuntu:~/vehicle_signal_specification/vss-tools$ python3 vspec2json.py -I ../spec/ --json-pretty  ../spec/VehicleSignalSpecification.vspec  single.json
Output to json format
Loading vspec from ../spec/VehicleSignalSpecification.vspec...
Calling exporter...
Generating JSON output...
Serializing pretty JSON...
All done.
sebastian@armbuntu:~/vehicle_signal_specification/vss-tools$ python3 vspec2json.py -I ../spec/ --json-pretty  ../spec/VehicleSignalSpecification.vspec  overlay.json
Output to json format
Loading vspec from ../spec/VehicleSignalSpecification.vspec...
Calling exporter...
Generating JSON output...
Serializing pretty JSON...
All done.
sebastian@armbuntu:~/vehicle_signal_specification/vss-tools$ md5sum single.json overlay.json 
15fb22ccb8dd0b19adea16ad44b569bb  single.json
15fb22ccb8dd0b19adea16ad44b569bb  overlay.json
sebastian@armbuntu:~/vehicle_signal_specification/vss-tools$ 
```

For an example with overlay

```
python3 vspec2json.py -I ../spec/ --json-pretty -o test.vspec ../spec/VehicleSignalSpecification.vspec  test.json
```

This is a valid overlay

```yaml
#
# Example overlay
#
Vehicle:
  type: branch



Vehicle.RandomSpeed:
  datatype: float
  type: sensor
  unit: km/h
  description: Average speed for the current trip.


Vehicle.AmbientAirTemperature:
  datatype: float
  type: sensor
  unit: celsius
  min: 2500
  description: Temperature at the surface of the sun



Vehicle.Cabin:
  type: branch
  description: Fancy description for cabin

Vehicle.Ashtray:
  type: branch
  description: Describes stuff in  the Ashtray
```

The failed CI is "fine" it should work after https://github.com/COVESA/vss-tools/pull/156 is merged and this is rebased (because Franka tolling is not optional  in CI even tough it used deprecated parser)

The overlay feature should work with __all__ vspec2x exporters (yaml, json, vs, binary)

Please test @erikbosch @danielwilms

### DIFF
--- a/.github/workflows/buildcheck.yml
+++ b/.github/workflows/buildcheck.yml
@@ -1,57 +1,61 @@
 name: Standard Build Check
 
-on: [push, pull_request]
+on:
+  push:
+    branches: [ master ]
+  pull_request:
+    types:
+      - opened
+      - synchronize
+      - reopened
+      - ready_for_review
+
+concurrency:
+  group: ci-check-${{ github.ref }}
+  cancel-in-progress: true
+
+env:
+  VSS_PATH: ${{ github.workspace }}/vehicle_signal_specification
+  VSS_TOOLS_PATH: ${{ github.workspace }}/vehicle_signal_specification/vss-tools
+  PYTHON_VERSION: "3.8.12"  # keep in sync with vss-tools/Pipfile!
+  CI: 1  # shall any script needs to know if it's running in the CI
 
 jobs:
   buildtest:
+    if: github.event.pull_request.draft != true
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v2
-
-      - name: Install pyenv
+      - name: Checkout VSS as parent
+        uses: actions/checkout@v3
+        with:
+          repository: COVESA/vehicle_signal_specification
+          path: ${{ env.VSS_PATH }}
+          submodules: false
+      - name: Remove VSS vss-tools from parent
         run: |
-          # Needed for using exact python version, which is
-          # required in `Pipfile`.
-          curl https://pyenv.run | bash
-          export PATH="$HOME/.pyenv/bin:$PATH"
-          pyenv install 3.8.12
-      - name: Install dependencies
+          rm -rf ${{ env.VSS_TOOLS_PATH }}
+      - name: Checkout vss-tools inside vehicle_signal_specification/
+        uses: actions/checkout@v3
+        with:
+          path: ${{ env.VSS_TOOLS_PATH }}
+      - uses: actions/setup-python@v3
+        with:
+          python-version: ${{ env.PYTHON_VERSION }}
+          cache: pipenv
+      - name: Install pipenv
         run: |
-          python -V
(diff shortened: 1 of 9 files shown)


erikbosch:
Great work. I think we need to describe somewhere exactly what the "merge" supports. I suspect it does not support all ideas in https://github.com/COVESA/vehicle_signal_specification/wiki/VSS-Layers-Concept, like removing nodes.

This also makes me think if we should try to come up with a better test framework for the tools to be run by CI to be able to spot regressions in the future. Like creating small test cases not relying on VSS but rather on a simple small model from which we generate e.g. JSON and compare with a pre-stored instance.

SebastianSchildt:
Current state

 - instances will not work as expected I assume (becasue we exapnd before merge) -> This needs to be fixed
 - no warnings yet hen overwriting core attributes
 - as I said, no support yet for new arbitrary metadata

The ifrst point needs to be solved (here) to have a first usable implementation, warnings are nice to have, and the last one is sort of "Independent" (can be a separate PR), but just needs some plumbing in our data model, and the exporters need to be aware

SebastianSchildt:
Not ready yet, but need to set to ready to test CI

SebastianSchildt:
Ready for review, please refer to **first comment**, I updated it with everything important

erikbosch:
Found a possible bug, sent it to Sebastian for analysis. I also think that it would be good to document an example. An idea is to create a new file `vspec2x.md` where one or more examples of overlays are shown as a "mini-tutorial", andreference it from the [the table in the main README file](https://github.com/COVESA/vss-tools#available-tools).

SebastianSchildt:
Please try to break it again @erikbosch  😸

erikbosch (approved):
Could not find any more bug now. Documentation can be handled in a later PR.

SebastianSchildt:
Rebased to https://github.com/COVESA/vss-tools/pull/156  to fix CI runs

danielwilms:
decision to merge 2022/05/16

PR #156, opened 2022-04-27
Moved vspec2franca to anytree

danielwilms:
- changed to anytree in creation
- adopted struct and generated output

@gunnarx, could you please check, if correct?

### DIFF
--- a/vspec2x.py
+++ b/vspec2x.py
@@ -1,5 +1,6 @@
 #!/usr/bin/env python3
 
+# (c) 2022 BMW Group
 # (c) 2022 Robert Bosch GmbH
 # (c) 2016 Jaguar Land Rover
 #
@@ -15,7 +16,7 @@ from enum import Enum
 import sys
 import vspec
 
-from vssexporters import vss2json, vss2csv, vss2yaml, vss2binary
+from vssexporters import vss2json, vss2csv, vss2yaml, vss2binary, vss2franca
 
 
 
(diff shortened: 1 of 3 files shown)


erikbosch:
@danielwilms @SebastianSchildt noticed that both your builds fail. Do not know if it concerns some form of circular dependency, that we specify that we require vss-tools in Pipfile but there is no (recent) one.

vspec2x.py
```diff
@@ -32,6 +33,7 @@ def export(config: argparse.Namespace, root: VSSNode):
     csv = vss2csv
     yaml = vss2yaml
     binary = vss2binary
+	franca = vss2franca
```
erikbosch, line 36: Odd indent, tab?

danielwilms: true, will change that

danielwilms:
> @danielwilms @SebastianSchildt noticed that both your builds fail. Do not know if it concerns some form of circular dependency, that we specify that we require vss-tools in Pipfile but there is no (recent) one.

I should have mentioned it in the PR. I took away the version flag, which was specific for franca. Either we leave it, and I'll implement it or we take it away from the makefile as well. Let's discuss tonight.

erikbosch:
Possibly need to rebase to get the new buildcheck changes

erikbosch:
@danielwilms - do you think you have time to fix the review comments before the meeting tomorrow (and possibly rebase, if that helps to make the build pass)

erikbosch (approved):
Looks good for me

SebastianSchildt:
Looking good to me "structure-wise", I can not  say anything about the output

danielwilms:
decision to merge 2022/05/16

PR #157, opened 2022-04-28
Trying to fix build problem

erikbosch:
Tries to the build problems experienced in #155 and #156.

I am by no means a pip expert, so proposals on better solutions are welcome

### DIFF
--- a/.github/workflows/buildcheck.yml
+++ b/.github/workflows/buildcheck.yml
@@ -21,7 +21,6 @@ jobs:
           python -m pip --quiet --no-input install --upgrade pip 
           python -m pip --quiet --no-input install --upgrade pipenv wheel
           pipenv install
-          pipenv run python setup.py -q install
         env:
           PIPENV_VENV_IN_PROJECT: 1
 
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,3 +1,0 @@
-[build-system]
-requires = ["setuptools", "wheel"]
-build-backend = "setuptools.build_meta"


erikbosch:
FYI - @barbieri @SebastianSchildt @danielwilms - removing the vss-tools reference (and doing `pipenv lock`) seems to solve the build problem. But is it the "right" way to solve it?

Pipfile
```diff
@@ -10,7 +10,6 @@ deprecation = ">=2.1.0"
 pytest = ">=2.7.2"
 PyYAML = ">=5.1"
 rdflib = ">=6.0.0"
-vss-tools = {editable = true, path = "."}
```
barbieri, line 13: this turns the pipfile usage useless... since the current project wouldn't be reached from inside pipenv.

To me the problem seems to be https://github.com/COVESA/vss-tools/blob/master/.github/workflows/buildcheck.yml#L24 remove that line.

I'd also suggest fixing the workflow to use https://github.com/actions/setup-python and remove all `pipenv install` handling, also the `PIPENV_VENV_IN_PROJECT` and `PIPENV_PIPFILE`, just say `cache: pip`.

erikbosch: Thanks for the ideas. Does not seem to be as easy as just removing line 24.

barbieri: tomorrow I'll try to give it a try and propose a PR fixing this issue and using GitHub's best practices (I'm doing similar stuff for some internal projects, so I got to understand some pitfalls and pain points...)

barbieri:
please take a look at #158, the CI is working and much faster since it's using cached Python stuff...

erikbosch:
Replaced by #158

PR #158, opened 2022-04-29
build: improve continuous integration using GitHub Actions

barbieri:
- use actions/checkout to do multiple, nested checkouts of both vehicle_signal_specification and vss-tools

- use actions/setup-python to more easily install the correct python-version and also cache pipenv modules

- Cancel previous execution of the workflow and only run if the PR is not a draft

Closes: #157

### DIFF
--- a/.github/workflows/buildcheck.yml
+++ b/.github/workflows/buildcheck.yml
@@ -1,57 +1,61 @@
 name: Standard Build Check
 
-on: [push, pull_request]
+on:
+  push:
+    branches: [ master ]
+  pull_request:
+    types:
+      - opened
+      - synchronize
+      - reopened
+      - ready_for_review
+
+concurrency:
+  group: ci-check-${{ github.ref }}
+  cancel-in-progress: true
+
+env:
+  VSS_PATH: ${{ github.workspace }}/vehicle_signal_specification
+  VSS_TOOLS_PATH: ${{ github.workspace }}/vehicle_signal_specification/vss-tools
+  PYTHON_VERSION: "3.8.12"  # keep in sync with vss-tools/Pipfile!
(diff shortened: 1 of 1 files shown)


barbieri:
@erikbosch it's now working 😅

these CI stuff is always a pain, takes a while to test, hard to test it locally... but now seems to be fully working, takes around 36s to execute (versus 4minutes in the previous workflow).

.github/workflows/buildcheck.yml
```diff
@@ -1,57 +1,61 @@
...
-          # "Our" version of vss-tools was prepared inside VSS => go there and run tests
-          cd ../vehicle_signal_specification
-          pipenv install
-          pipenv run make travis_targets
-        env:
-          PIPENV_PIPFILE: ./vss-tools/Pipfile
-          PIPENV_VENV_IN_PROJECT: 1
-      - name: Test optional targets. NOTE - always succeeds
+          cd ${{ env.VSS_TOOLS_PATH }}
+          pipenv run make -C "${{ env.VSS_PATH }}" travis_targets
```
barbieri, line 56: this one was a PITA to get right... usually I just have one repo, but with the nested approach it's cumbersome: you **MUST** run the `pipenv run` from **INSIDE VSS-TOOLS**, otherwise it will create **another** virtualenv...

Then we must add `-C DIR` to have make to chdir before it executes... If you have some other command, you'd have to go with `pipenv run sh -c "cd DIR && execute your stuff"`

barbieri:
btw, python 3.8.13 is out, I'd suggest to change the Pipfile and re-lock with all new dependencies.

erikbosch (approved):
Looks good for me, thanks for your effort @barbieri

danielwilms:
decision to merge: 22/05/03

PR #159, opened 2022-05-04
Adding JSON diff test cases

erikbosch:
This is an idea on how we could improve vss-tools to catch regressions that we might introduce. The idea is to allow us to add an arbitrary number of models as test cases. The initial PR contains three test cases, relies on generating JSON and comparing output with reference output, but the framework could be used for other tests as well. Most likely our JSON output shall be quite stable, so the reference output shall likely be quite stable over time.

I think a framework like this could be useful as tests  (and reference examples) for e.g. #155, where we could add a number of test cases to cover various corner cases of layers/overlay. I am by no way a script/test expert, the solution can likely be improved but this seems to work.


erikbosch:
This is how result is presented in the build:
image

erikbosch:
In discussion with Daniel regarding if a Python-based test framework would be better. Converting it to a draft for now.

danielwilms:
@danielwilms push PR

SebastianSchildt:
Please check @danielwilms

erikbosch:
Meeting decision to merge

PR #167, opened 2022-06-07
Adding first tests for checking instances

danielwilms:
Using pytest as testing framework to check some good cases of instantiation. 
Meant to be extended.

### DIFF
--- a/tests/instances/resources/instance_complex_1.vspec
+++ b/tests/instances/resources/instance_complex_1.vspec
@@ -1,0 +1,14 @@
+Vehicle:
+  type: branch
+  instances: 
+    - ["Test1"]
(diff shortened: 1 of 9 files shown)


erikbosch:
Uses a different approach than my PR in #159. This one is more classic unit-testing, e.g. parse a vspec and then traverse model to check that it contain what is expected. The #159 is more black box, does not care about the internal model, only compares output with expected output. Could possibly be discussed if we think that the concept in this PR suffice, or if it makes sense to have both.

danielwilms:
will fix the failed build. It's about the path to the testfiles

danielwilms:
> will fix the failed build. It's about the path to the testfiles

fixed.

PR #168, opened 2022-06-07
Exclude marked nodes from instantiation

danielwilms:
If a node is marked with `instantiate: False` it
won't use the instances as parent, but the original one
instead.

Related to: https://github.com/COVESA/vehicle_signal_specification/issues/448

### DIFF
--- a/vspec/model/vsstree.py
+++ b/vspec/model/vsstree.py
@@ -55,7 +55,7 @@ class VSSNode(Node):
     type: VSSType
     datatype: VSSDataType
 
-    core_attributes = ["type", "children", "datatype", "description", "unit", "uuid", "min", "max", "allowed",
+    core_attributes = ["type", "children", "datatype", "description", "unit", "uuid", "min", "max", "allowed", "instantiate",
                        "aggregate", "default", "instances", "deprecation", "arraysize", "comment", "$file_name$"]
 
     # List of accepted extended attributes. In strict terminate if an attribute is
@@ -66,6 +66,7 @@ class VSSNode(Node):
     min = ""
     max = ""
     allowed = ""
+    instantiate = True
 
     ttl_name = ""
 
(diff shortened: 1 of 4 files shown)


UlfBj:
The comment by Sebastian
"It seems current implementation always needs to spell out a complete leaf, instead of being able to just override one attribute"
I think should be contemplated together with how removal of attributes, or entire nodes, shall be expressed.

danielwilms:
@danielwilms: search the ticket with removal of nodes and attributes.

danielwilms:
@danielwilms rebase pending, but ready for review

vspec/model/vsstree.py
```diff
@@ -42,6 +42,7 @@ class VSSNode(Node):
     min = ""
     max = ""
     allowed = ""
+    instantiate = True
```
tobiask84, line 45: A lot of programmers prefix booleans by convention either with "is" or "has" like for example `user.isAdmin`. This can be helpful to identify a var as boolean immediately when reading the name. It can help too with the autocomplete feature of the IDE. Following this standard the value here would normally be something like `isInstance`.

danielwilms: @tobiask84, valid point. Just thinking that `isInstance` is not correct, as it's not an instance yet but will be instantiated. Then to be consequent we would have to change `allow` to `isAllowed`, `max` to `hasMax` etc. I don't know if we wanna go there.

But don't get me wrong, I'm all in for making it easier to understand.

danielwilms:
wait after overlay merge

erikbosch:
TODO:
@danielwilms to fix conflict
@erikbosch  to do quick review
Decision to merge when rebased and Erik has approved, aim to include for 3.0

vspec/model/vsstree.py
```diff
@@ -118,6 +119,9 @@ def __init__(self, name, source_dict: dict, parent=None, children=None, break_on
             
         if "comment" in source_dict.keys():
             self.comment = source_dict["comment"]
+        
+        if "instantiate" in source_dict.keys():
```
erikbosch, line 123: After rebase I do not think that the two lines here are needed as they are handled by the generic mechanism introduced by Sebastian. But I am thinking if it would make sense to have a check that verifies that the value after "instantiate" actually is a boolean value. Today you will not get an error if you use `instantiate: Falsch`, I think it will be treated as  `instantiate: True`

If we do not want to allow the flexible Python (or is yaml?) approach to allow a lot of synonyms like `On` and `Off`, we should maybe check that it is exactly `true`or `false` (doc PR currently use only lower case letters)

danielwilms: true, will fix that

PR #171, opened 2022-06-13
VSSo: VehicleComponents as instances

danielwilms:
Currently the VehicleComponents were modelled as
subclasses with `rdfs:subclass`. This change introduces
a variable to choose either way until it's decided.

Additionally introduced documentation improvements and
code cleanups.

Mentioned in: https://github.com/w3c/vsso/issues/22

### DIFF
--- a/contrib/vspec2ttl/vspec2ttl.py
+++ b/contrib/vspec2ttl/vspec2ttl.py
@@ -31,8 +31,21 @@ from rdflib.namespace import RDFS, OWL, RDF,  SKOS
 from enum import Enum
 
 
+# There are two options of handling branches:
+# 1. branches are handled as _subclasses_ of vehicle components
+# 2. branches are handled as _instances_ of vehicle components
+#
+# As it's not decided for now, we can generate both by setting this
+# variable to True or False. True means that branches are handled as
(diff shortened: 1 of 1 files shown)


contrib/vspec2ttl/vspec2ttl.py
```diff
@@ -31,8 +31,21 @@
...
+# There are two options of handling branches:
+# 1. branches are handled as _subclasses_ of vehicle components
+# 2. branches are handled as _instances_ of vehicle components
+#
+# As it's not decided for now, we can generate both by setting this
+# variable to True or False. True means that branches are handled as
+# subclasses of vehicle components. False means that branches are handled
+# as instances of vehicle components.
+
+COMPONENTS_AS_CLASSES = False
```
erikbosch, line 43: What do we see as possible time range for keeping both options. If we intend to keep both for a long period, then we should maybe think about adapting CI so that we test both options, to make sure that we do not introduce regressions in any of them.

danielwilms: It's just for now, to avoid jumping from one commit to the other. It should be decided in the next couple of weeks.

danielwilms:
decision to merge 2022/06/14

PR #172, opened 2022-06-13
Introducing new contrib structure

danielwilms:
Current contrib exporters live without major tool-updates.
 Proposal to do the following:
    1. create exporters for them in `vssexporters/contrib`
    2. dont create shortcut, but call them in Makefile with vspec2x --format <format>

### DIFF
--- a/contrib/vspec2ttl/vspec2ttl.py
+++ b/contrib/vspec2ttl/vspec2ttl.py
@@ -31,8 +31,21 @@ from rdflib.namespace import RDFS, OWL, RDF,  SKOS
 from enum import Enum
 
 
+# There are two options of handling branches:
+# 1. branches are handled as _subclasses_ of vehicle components
+# 2. branches are handled as _instances_ of vehicle components
+#
+# As it's not decided for now, we can generate both by setting this
+# variable to True or False. True means that branches are handled as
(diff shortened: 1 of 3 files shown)


SebastianSchildt:
Would we add them as "optional" targets in CI?

danielwilms:
> Would we add them as "optional" targets in CI?

Yes, that would be the idea

danielwilms:
question: how to handle dependencies of contrib?

danielwilms:
> question: how to handle dependencies of contrib?

@SebastianSchildt, can be done see https://github.com/COVESA/vss-tools/pull/172/commits/cd777735224075b95a51b928ad21bbb4f4247fd5. At the end try except would have to be done for each exporter in order to be sure that all, which can be imported are so. But that shouldn't be a huge problem.

SebastianSchildt:
Hi, I like this pattern, Reasonably easy without the need to write a full blown "plugin loader". As this draft (due to the 2ttl?) relies on the extended attribute parameter, I suggest we get that nailed down first

erikbosch:
@danielwilms - please finish this one before you leave for good.

adobekan:
@danielwilms will create new PR.

Issue #139, opened 2022-01-25
Add tools support for private config.yaml


danielwilms:

erikbosch:
Has been fixed by #200

PR #174, opened 2022-07-04
add more constants

tobiask84:
This MR adds one new node type (Unix Timestamp) and 3 new units.

These units are used in the BMW private spec. The COVESA units currently are not customizable/extandable. Since a unit can be used by anyone, it does not "hurt" to support additional units in COVESA, even if they are not used within the COVESA spec.

The new node type Unix Timestamp is self explanatory. Since timestamps can be mapped to different datatypes depending on the target system, it does not make sense to already define a primitive type in the VSpec. Defining a node as a UNIX Timestamp leaves more flexibilty to tools that process the spec to decide on a different primitive type depending on the tool/usecase/environment at a later stage of the processing. Since unix timestamps are used in many systems, it does make sense to allow it as a node type.


SebastianSchildt:
Hi,
maybe split it? I assume the units are probably (let's see :) ) fine with everyone, because as you said "not hurting".
 (Actually the niche/"correct" way would be to use the official API https://github.com/COVESA/vss-tools/blob/master/FAQ.md and finally closing this issue: https://github.com/COVESA/vss-tools/issues/139 

For the timestamp I feel it may need more discussion: It was decided a year ago or so to remove it ( It hink very early VSS had it), and this has implications for downstream tooling, i.e. if you have tooling that can deal with all VSS datatypes, it would need to be extended to support the "new" type to be VSS compliant. (in an ideal world would be easier, if such things could also be configured during runtime, but opposed to units I think for types there is not even infrastructure for it currently)

SebastianSchildt:
Found it. For reference: https://github.com/COVESA/vss-tools/pull/103

tobiask84:
@SebastianSchildt 
Ok, I overwrote my commit and removed the timestamp.

vspec/config.yaml
```diff
@@ -203,3 +203,16 @@ units:
...
+    label: nano meter per kilometer
+    description: nm/km
+    domain: nm/km
+  dBm:
+    label: decibel milliwatt
+    description: Wear of tire tread depth
+    domain: relation
+  kN:
+    label: kilo newton
+    description: force measured in kilo newton
```
adobekan, line 216: Formatting.

tobiask84: Fixed

vspec/config.yaml
```diff
@@ -203,3 +203,16 @@ units:
     label: percent
     description: Relation measured in percent
     domain: relation
+  nm/km:
+    label: nano meter per kilometer
+    description: nm/km
+    domain: nm/km
+  dBm:
+    label: decibel milliwatt
+    description: Wear of tire tread depth
```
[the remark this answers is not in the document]
tobiask84: Fixed

adobekan (approved):
kk

danielwilms:
decision to merge 12/07/22

PR #177, opened 2022-08-02
Fixes bug in binary exporter when node has no description

erikbosch:
Previously binary exporter crashed if description was missing for nodes.


erikbosch:
This fixes the build error in https://github.com/COVESA/vehicle_signal_specification/pull/465 

(But the bug shall be no problem if the missing "description" is added in the PR, so I do not see this PR as mandatory for 3.0 release)

@UlfBj - could you check if you have any comments

UlfBj:
@erikbosch I tested your fix, and it worked fine.

Issue #178, opened 2022-08-06
Fix DDS IDL converter


SebastianSchildt:
See also https://github.com/COVESA/vehicle_signal_specification/issues/475

Guess wie ddi not catch it, as CI only checks the tool ran, and not output

erikbosch:
Fixed by #179, closing it.

PR #179, opened 2022-08-07
fixed issue w.r.t. value attribute generation in DDS-IDL file

zam1kor:
fixed **vss2ddsidl.py** file, based on the previous change in **VSSNode** -> **data_type** attribute is changed to **datatype**. This change of the attribute name is considered in this commit


erikbosch:
Fixes #178

erikbosch:
Looks good

Issue #185, opened 2022-08-23
DDS export of vss generated IDL fails


AkhilTThomas:
Hi,

I'm trying to generate the python idl types for the vss data points.

IDL creation
``python vspec2ddsidl.py -I ..\spec ..\spec\VehicleSignalSpecification.vspec vss.idl``

IDL to python conversion
``/opt/cycloneDDS/bin/idlc  -l py vss.idl``

Error:
``
/tmp/dds/vss.idl:597:27: Declaration 'UNKNOWN' collides with earlier an declaration of 'UNKNOWN'
/tmp/dds/vss.idl:693: fatal error: File write error
    };
1 error in preprocessor.
``

The vss IDL generated enums with same value:

line 597 ``enum AspirationTypeValues{UNKNOWN,NATURAL,SUPERCHARGER,TURBOCHARGER};``

line 535 ``enum ConfigurationValues{UNKNOWN,STRAIGHT,V,BOXER,W,ROTARY,RADIAL,SQUARE,H,U,OPPOSED,X};``

vss = _release/3.0_
vss-tools =_release/3.0_
idlc (Eclipse Cyclone DDS)  = _0.10.0_

Is this an error in the VSS or the IDLc from cyclonedds ?

SebastianSchildt:
@zam1kor  can you check? I am not sure, does DDS IDL not allow to choose same symbolic names in different enums (seems strange). Does FastDDS IDL compiler behave differently?

If we are doing something illegal in DDS-IDL we need to rethink that part of the mapping, otherwise maybe open a report upstream in Cyclone.

AkhilTThomas:
FastDDS throws even more errors!

interesting ones:
``vss.idl:5122:0: error: Illegal identifier: Switch is a keyword, use escape character if you want to use it as identifier (_Switch)
``
``vss.idl:671:0: error: Illegal identifier: MAP is a keyword, use escape character if you want to use it as identifier (_MAP)
``
https://github.com/COVESA/vehicle_signal_specification/blob/2b24d0d9ac6b2d357e1d97ca3064cc2ca5bfa089/spec/Cabin/SingleSliderSwitch.vspec

Switch:
  datatype: string
  type: actuator
  allowed: ['INACTIVE', 'CLOSE', 'OPEN', 'ONE_SHOT_CLOSE', 'ONE_SHOT_OPEN']
  description: Switch controlling sliding action such as window, sunroof, or blind.

I've attached the errors as well as the idl generated
[...]

PR #186, opened 2022-08-24
DDS export of vss generated IDL fails Fix

AkhilTThomas:
Added C,Python and IDL keyword crosss checking and appended underscore in front of keywords
Ex: MAP becomes _MAP
Also enums now exist in their own namespace under a module wrapper, the modules name is the enum name+_M
Ex:
module LowVoltageSystemState_M` { enum LowVoltageSystemStateValues{UNDEFINED,LOCK,OFF,ACC,ON,START}; }; struct LowVoltageSystemState { LowVoltageSystemState_M::LowVoltageSystemStateValues value; //const string type ="sensor"; //const string  #description="State of the supply voltage of the control units (usually 12V)."; };
#185

### DIFF
--- a/docs/VSS2DDSIDL.md
+++ b/docs/VSS2DDSIDL.md
@@ -50,7 +50,7 @@ Below elements are considered only if the switch ***--all-idl-features*** is sup
 
 | VSS    | DDS-IDL         |
 |--------|----------------|
-| <pre>Direction:<br> datatype:string<br> type: actuator<br> allowed: ['FORWARD','BACKWARD']<br> description: Driving direction of the vehicle</pre>  | <pre>enum DirectionValues{FORWARD,BACKWARD};<br>struct Direction<br>{<br>string uuid;<br>DirectionValues value;<br>};</pre>   
+| <pre>Direction:<br> datatype:string<br> type: actuator<br> allowed: ['FORWARD','BACKWARD']<br> description: Driving direction of the vehicle</pre>  | <pre>module Direction_M {<br>enum DirectionValues{FORWARD,BACKWARD};<br>};<br>struct Direction<br>{<br>string uuid;<br>DirectionValues value;<br>};</pre>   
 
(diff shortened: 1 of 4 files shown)


SebastianSchildt:
(note: CI test failure is to be expected, because here the fix is changing the idl output, and the CI simply tests an example file against an expected known output)

SebastianSchildt:
Not an DDS-IDL expert, but just with my VSS head on, I appreciate the progress of having a valid/legal DDS output from VSS 😁 . @nayakned since I happen to know you are using DDS with VSS, can you check whether this variant of DDS-IDL mapping would work with your use cases/models, and give feedback?

erikbosch (approved):

erikbosch:
@AkhilTThomas did you close it by mistake? It is a PR so to be included in master it must be merged. I marked this PR as approved, but wanted input from others before clicking the merge button.

If it was closed by mistake you or me can reopen it.

AkhilTThomas:
@erikbosch I have opened it.

nayakned:
> Not an DDS-IDL expert, but just with my VSS head on, I appreciate the progress of having a valid/legal DDS output from VSS grin . @nayakned since I happen to know you are using DDS with VSS, can you check whether this variant of DDS-IDL mapping would work with your use cases/models, and give feedback?

Looks ok to me. I think our usecases should (still) work. However, this doesnt completely solve #185 I think. Since the problem with redefined enums would remains.

AkhilTThomas:
> Looks ok to me. I think our usecases should (still) work. However, this doesnt completely solve #185 I think. Since the problem with redefined enums would remains.

The enums are now encased in their own module block and accessed via this module scope. This avoids the conflict

nayakned:
> The enums are now encased in their own module block and accessed via this module scope. This avoids the conflict

Ah. had missed that part. Thanks for the clarification.

erikbosch:
Merging it to allow some other PRs to be updated.

PR #188, opened 2022-08-25
Remove stringcase dependency

erikbosch:
We do not seem to use the functionality, and stringcase might (in some way) trigger the build problems we sometimes see. This PR tests removing it.

**NOTE:**

As can be seen in this PR `stringcase` can be easily removed as we actually do not use the functionality, we always use `NONE` as formatting. The question to be discussed is if we want to remove it, or rather refactor to use another library in case any tool would need it in the future.

### DIFF
--- a/Pipfile
+++ b/Pipfile
@@ -5,7 +5,6 @@ name = "pypi"
 
 [packages]
 anytree = ">=2.8.0"
-stringcase = ">=1.2.0"
 deprecation = ">=2.1.0"
 pytest = ">=2.7.2"
 PyYAML = ">=5.1"
--- a/Pipfile.lock
+++ b/Pipfile.lock
@@ -1,7 +1,7 @@
 {
     "_meta": {
         "hash": {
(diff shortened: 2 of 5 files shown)


erikbosch:
I would like to get comments on:

- Now when hopefully the build problem have been solved (by https://github.com/COVESA/vehicle_signal_specification/pull/483), do we still want to do any changes regarding stringcase (as it is old and not that maintained)
- If so, is removing (like in this PR) the preferred approach (as we do not use the functionality anyway)
- Or pay the effort do migrate to something else (but the ones that can be found from https://github.com/okunishinishi/python-stringcase/issues/34 seems to be mostly one-man-shows anyway)

erikbosch:
Meeting notes: Seems OK to remove. Please review change until next meeting

SebastianSchildt (approved):
Tested and works (fresh install in new pipenv, and ran all the non-contrib tools)
I "ugly" with the Git webui fixed the merge conflict in setup.py (py 3.7 vs. 3.8). Feel free to niceify via force-push

PR #189, opened 2022-08-25
Add new parameter --uuid

erikbosch:
This is the first step towards changing uuid generation.

A new parameter --uuid is introduced

Behavior with this PR:

- Neither --uuid or --no-uuid given: UUIDs are generated (like today), but warning generated
- Only --uuid given: UUIDs generated (like today when no argument given)
- Only --no-uuid given: No UUIDs given (like today)
- Both --uuid and --no-uuid given: Error

Intention is to change default behavior in VSS 4.0 so that no UUID becomes default.

Proposed migration strategy:

- Users of VSS 3.X, X> 0 will get a warning if they use the tools without specifying either `--uuid` or `--no_uuid`. If it is important whether UUIDs are generated or not they shall add  either `--uuid` or `--no_uuid` as preparation for VSS 4.0
- From VSS 4.0 generation of no UUIDs will be default, and `--no_uuid` will be deprecated and a warning shall be given if used. The warning if neither `--uuid` nor `--no_uuid` is used shall be removed.
- From VSS 5.0 support for `--no_uuid` shall be removed

Signed-off-by: Erik Jaegervall <erik.jaegervall@se.bosch.com>


erikbosch:
Meeting notes:

- At least for minor releases we should not remove UUID, but rather possibly generate both, and rmeove the old variant with UUID first in next major release (v4.0)

adobekan:
Discussion in the call 06/09/22:
- Keep old command for input --no-uuid. Introduce deprecation  with date and LOG message.
- Introduce new command input --uuid. 
- If parameter is not provided, create message for the default behavior.

erikbosch:
Meeting notes: Possibly lightweight list of upcoming/planned changes (wiki/md) so we keep track of deprecations and when it will be removed.

erikbosch:
Added a CHANGELOG file similar to https://github.com/COVESA/vehicle_signal_specification/pull/491

erikbosch:
Meeting decision: merge

PR #199, opened 2022-11-02
refactor dir structure

tobiask84:
Pip allows to install a dependancy like this:

pip install git+https://github.com/COVESA/vss-tools

or

pip install git+https://github.com/COVESA/vss-tools@commit-id

That is cleaner and easier than to check out a submodule, install the dependancies of that submodule and then access the directory structure of that submodule from within build pipelines.

However it is currently not possible to install COVESA like that because the COVESA python module exports only the vspec dir and that vspec dir is missing vital code like the vssexporters directory which is located parallel to the vspec dir. This breaks importing the vss-tools as a normal dependancy. All code that is necessary to run the tools needs to be in the directory that is exported by the module (ie the vspec dir). 

This will then allow to install the COVESA tools as a normal dependancy. And that is easier than to checkout the repo and manually install the directory.


tobiask84:
Yeah. you can install the dependancy and then run the tools to check that. After you install the dependancy you can execute any tool by just writing its name on the console. Which actually is easier then to call python and pass the path of the other repo. Your VSS code should not be concerned with the dir structure of the tools. These things should be decoupled. Different modules should interact through the module system. Not by referencing each other through directory of submodules.

erikbosch:
Thanks @tobiask84 for the explanations. The python-packaging has not got much attention lately, so your contributions and ideas are most welcome

erikbosch (approved):

tobiask84:
@erikbosch How is the process to get things merged?

PR #200, opened 2022-11-02
Refactor unit handling

erikbosch:
We have an old decision to move units to VSS, to separate the metamodel from VSS catalog. This is an attempt of a first step to achieving that by allowing unit files to be given as parameters. As the decision is quite old I would like to have a discussion on if this still is the right way to go.

This is a change that would need to be done in multiple steps, and this PR is not ready for merge yet as there are some things like tests that better shall be added, but the path I see for moving units to VSS completely is as follows:

- First add support in vss-tools to give unit files as parameters (this PR)
- Then copy the current config.yaml to VSS and let all testcases/docs in VSS give it as parameter
- When done and verified, remove the default handling in vss-tools, i.e. request that `-u` always is given if you want to specify units.

Fixes [#396](https://github.com/COVESA/vehicle_signal_specification/issues/396)
Fixes [#346](https://github.com/COVESA/vehicle_signal_specification/issues/346)

### DIFF
--- a/FAQ.md
+++ b/FAQ.md
@@ -1,38 +1,24 @@
 # Frequently asked questions
 
-### How to include my own units?
+## How to include my own units?
 
-Standard units are located at the vehicle_signal_specification at the
-(data_unit_types)[https://github.com/COVESA/vehicle_signal_specification/blob/master/docs-gen/content/rule_set/data_entry/data_unit_types.md]
-file. It's not possible to overwrite current units, but if you want to add extra
-units you can create a new file using the same pattern as the current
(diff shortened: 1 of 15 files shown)


erikbosch:
Meeting notes: Introducing this feature seems to be the right decision. Erik to add test cases and update PR

erikbosch:
Test cases added. Verifies that it works to specify unit file(s) explicitly, but also that default file still works.

erikbosch:
Meeting notes:

- Please review until next meeting

erikbosch:
Meeting notes:

- Decision to merge

PR #202, opened 2022-12-15
Unit support, part 2

erikbosch:
Extending the unit support work:

- Adapting contrib tools, they were not included in the previous PR
- Minor change in vspec2x to make it more clear that you need to give `-u`multiple times to include additional unit files
- A minor fix to support Python 3.10

### DIFF
--- a/binary/README.md
+++ b/binary/README.md
@@ -14,13 +14,13 @@ or, by invoking all tools:
 $ make all
 ```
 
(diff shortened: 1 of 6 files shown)


SebastianSchildt (approved):
Ahh user error. 

I learned that binary seem to be sneakily adding to an output file instead of recreating it. Starting with non-existing outputs also creates the same result.

So this seems to work ⚙️

erikbosch:
Meeting decision: merge

PR #203, opened 2022-12-16
Add support for overlays on instances

### LINKED ISSUE #201: Overlays on expanded tree?
There might for deployments be reasons to give additional data for individual signals, also for branches with instances. As of today we support overlays, but they are merged before we do instance expansion, so we cannot add information to individual instances. This could hypothetically be solved if we also add support for overlays to be merged after expansion. A hypothetical example below where we want to add a signal to the Row1.Pos1 seat, and in addition set min/max values for all seats. This is not possible with current tooling.

**Idea:**

_Shall we expand tooling so we can support two types of overlay insertion mechanisms, e.g. use `-o`for insertion before expansion, and `-o2` to insert after expansions?_ 

It would from a tooling perspective be a very easy solution.

Example:

```

#Something that only exist for one seat

Vehicle.Cabin.Seat.Row1.Pos1.IsHairDryerOn:
  type: actuator
  datatype: boolean

# Giving values that differ

Vehicle.Cabin.Seat.Row1.Pos1.Position:
  type: actuator
  datatype: uint16
  min: 0
  max: 125
  
Vehicle.Cabin.Seat.Row1.Pos2.Position:
  type: actuator
  datatype: uint16
  min: 0
  max: 125
  
Vehicle.Cabin.Seat.Row2.Pos1.Position:
  type: actuator
  datatype: uint16
  min: 0
  max: 60
  
Vehicle.Cabin.Seat.Row2.Pos2.Position:
  type: actuator
  datatype: uint16
  min: 0
  max: 60
```

erikbosch:
Fixes #201


erikbosch:
Changing to draft due to discussion in issue. In short - could we possibly make tool smarter so that different arguments are not needed.

erikbosch:
Meeting notes: Please review until next meeting

SebastianSchildt (commented):
There is definitely a curiosity that needs addressing, when compiling vss master with this PR or with vss-tools master the results are different.

I am puzzled as to what change caused it, the main difference is, that with this PR suddenly the trunks got a comment

```
(vss-tools) sebastian@sebastian-ThinkPad-S1-Yoga-12:~/vehicle_signal_specification/vss-tools$ diff outmaster.json outeo.json 
794a795
>                 "comment": "A trunk is a luggage compartment in a vehicle. Depending on vehicle, it can be either in the front or back of the vehicle. Some vehicles may have trunks both at the front and at the rear of the vehicle.",
813a815
>                 "comment": "A trunk is a luggage compartment in a vehicle. Depending on vehicle, it can be either in the front or back of the vehicle. Some vehicles may have trunks both at the front and at the rear of the vehicle.",
(vss-tools) sebastian@sebastian-ThinkPad-S1-Yoga-12:~/vehicle_signal_specification/vss-tools$ 
```

image

That may be correct, but I think then we are definitely missing test cases

SebastianSchildt:
Apart from the comments above, I played around and it did what I expected.

(side note, when testing  `instantiate: false` , it seems it is currently not used in the standard catalogue, because the example from docs "DoorCount" is meanwhile modelled as "normal" signal in Cabin)

erikbosch:
> There is definitely a curiosity that needs addressing, when compiling vss master with this PR or with vss-tools master the results are different.
> 
> I am puzzled as to what change caused it, the main difference is, that with this PR suddenly the trunks got a comment
> 

This is related to this [line](https://github.com/COVESA/vss-tools/blob/master/vspec/__init__.py#L395) in the old code. Previously when creating instance nodes we copied description but not comment from the "parent" branch. One could of course discuss if one wants to duplicate both description and comment on created children, or just keep them on the level above. I.e. does it make sense that "Front" has the same description and comment as "Trunk".

I will add a test case to show how it works

erikbosch:
Meeting decision: Merge

PR #207, opened 2023-01-18
Add megabyte and megabytes per second as units

CorvetteCole:
This is in anticipation for further work I am doing refining the Vehicle.Connectivity model. I chose megabytes as a good middle-ground between too-big or too-small units

### DIFF
--- a/vspec/config.yaml
+++ b/vspec/config.yaml
@@ -215,11 +215,11 @@ units:
     label: kilo newton
     description: Force measured in kilo newton
     domain: force
-  MB:
-    label: megabyte
-    description: Data measured in megabytes
-    domain: data
-  MB/s:
-    label: megabytes per second
-    description: Data per time measured in megabytes per second
-    domain: data per time
+  MiB:
+    label: mebibyte
+    description: Digital information measured in mebibyte (base 2)
+    domain: digital information
+  MiB/s:
+    label: mebibytes per second
+    description: Digital information per time measured in mebibytes per second (base 2)
+    domain: digital information per time
--- a/vspec/config.yaml
+++ b/vspec/config.yaml
@@ -215,4 +215,11 @@ units:
     label: kilo newton
     description: Force measured in kilo newton
     domain: force
-  
+  MB:
+    label: megabyte
+    description: Data measured in megabytes
+    domain: data
+  MB/s:
+    label: megabytes per second
+    description: Data per time measured in megabytes per second
+    domain: data per time


SebastianSchildt:
> This is in anticipation for further work I am doing refining the Vehicle.Connectivity model. I chose megabytes as a good middle-ground between too-big or too-small units

Cool, connectivity is a gap. Maybe we also need dBm, asu or similar things?

I think megaBYTES/s for bandwith is seldomly used, wouldn't Mbit or kbit be more common?

For megabytes at we should decide and document whether we mean MB (base 10, e.g. 1000 byte is 1 kByte etc., basically only used for persisitent storage) or MiB (base 2, 1000 byte is 1024 kByte, more common for verything else).

(and potentially discuss if in case of MiB still want to use the "wrong" (in SI sense) unit name "MB", as that is also quite common)

CorvetteCole:
@SebastianSchildt dBm seemingly is already implemented, interestingly enough! https://github.com/COVESA/vss-tools/blob/de88c8aec035bfb66a49f4c334a7ffb34ce69d76/vspec/config.yaml#L210

I chose to use MB/s vs Mb/s (or Mbps) because it felt like a simpler more natural choice to match the existing MB data unit. I'm not set on that though, and did debate with myself what choice was better. Happy to change if you think it is better as megabits.

Regarding MB and MiB... kill me. The whole data unit stuff is a mess. I actually think we should use MiB after thinking about it. Be accurate for the standard and then developers can present that data with the MB "wrong" unit if they want on their end. I suppose if we stick with MB/s it should become MiB/s then haha!

Let me know what you think and I'll revise this.

CorvetteCole:
We do also have the option of including both megabyte and mebibyte, if desired

vspec/config.yaml
```diff
@@ -215,4 +215,11 @@ units:
     label: kilo newton
     description: Force measured in kilo newton
     domain: force
-  
+  MB:
+    label: megabyte
+    description: Data measured in megabytes
```
erikbosch, line 220: Regarding the MB/MiB discussion. I think we shall select unit names so we can support both variants (if needed) in the future. And then it could be better to use `MiB` instead of `MB` if that is what we need. Regardless, to avoid misunderstanding  I think it could be good in the description to state what we refer to, by adding e.g. `(base 2)` or `(base 10)` at the end of the description

erikbosch:
For info, there is work in progress to move the unit file from vss-tools to vss, see https://github.com/COVESA/vehicle_signal_specification/pull/521. The rationale is that different signal catalogs may support different units, and it is then not the tooling that shall dictate what units to use. Intention is that the unit file in vss-tools by time shall be deleted.

But that work should not block this PR, we just need to make sure that we later merge the changes to the unit file in vss.

vspec/config.yaml
CorvetteCole: fixed this in the latest commit, thanks for the feedback

CorvetteCole:
I edited the units to be MiB and MiB/s respectively and added "(base 2)" for clarification. I also changed their domain to be "digital information" instead of "data" because I determined that to be more semantically correct. 

Let me know what else you might see problems with, I know you mentioned the MiB/s vs somewhat more common Mbps earlier. In my opinion stuff that is user facing is more often in MiB/s (e.g. file transfers, network speeds in an OS), but Mbps is used often for bandwidth which is confusing already. I think there is a case for MiB/s, MB/s, AND Mbps all existing under the same roof, but that isn't worth doing right now.

Maybe we start with this and expand on it if desired?

Thanks again for your feedback! I'm looking forward to opening a PR for the connectivity improvements once we've got this in.

SebastianSchildt:
I am all for starting small, and improving over time.

With regard to "data transferred/time unit", I think it depends a little bit what we want to model with it. 

I agree for "Netflix or OTA download speed" it might be presented as M(i)B/sec to users. I was thinking more about modelling something like "Your cellular/WiFi is connected to base station/AP and gross link speed is", that would be MBit/s, kBit/s, or similarly when monitoring/measuring speed on the network level, i.e. "Internet speed test".


But unit-wise I am fine with having one or both now. The nore interesting discussion comes when modelling specific VSS points and we do need to choose :)

adobekan:
> I am all for starting small, and improving over time.
> 
> With regard to "data transferred/time unit", I think it depends a little bit what we want to model with it.
> 
> I agree for "Netflix or OTA download speed" it might be presented as M(i)B/sec to users. I was thinking more about modelling something like "Your cellular/WiFi is connected to base station/AP and gross link speed is", that would be MBit/s, kBit/s, or similarly when monitoring/measuring speed on the network level, i.e. "Internet speed test".
> 
> But unit-wise I am fine with having one or both now. The nore interesting discussion comes when modelling specific VSS points and we do need to choose :)

As you said it would be more logical to go with Mb or Kb for bandwidth and throughput,  and then for reporting actual consumption in MB.

erikbosch:
Two thoughts.

First - Concerning what actual units to add, we could wait merging this pull request until we have had agreed on the new signals in https://github.com/COVESA/vehicle_signal_specification/pull/526 and just add the units needed for the new signals in the PR.

Second - we have a process where each commit should be signed off (although this has not always been followed strictly in the past). Please see the [Contribution Guidelines](https://www.covesa.global/contribute).:

Every contribution must carry the following sign-off line with your real name and email address: "Signed-off-by: Firstname Lastname <[firstname.lastname@example.com](mailto:firstname.lastname@example.com)>"

I assume you do not work for an organization with a CLA with COVESA, so then it is Code Contribution Option 1: Using the Project License that is relevant for you. Please read the contribution guidelines, and if you find them acceptable add a signed-off line to (each of) your commits and do a force push to your branch.

CorvetteCole:
sorry about that @erikbosch. I read the contributor guidelines but totally missed that - I'll fix it in a few (on the other PR too).

About the first point, I agree. Hopefully we should be able to discuss that a little tomorrow and we'll know then.

erikbosch:
Suggestion - lets wait with merging this one until #248 is merged. Then this PR can be closed and instead the file https://github.com/COVESA/vehicle_signal_specification/blob/master/spec/units.yaml shall be changed, either as part of https://github.com/COVESA/vehicle_signal_specification/pull/526 or as a separate PR

CorvetteCole:
yes I agree, I need a little time for the PR depending on this anyways.

PR #211, opened 2023-01-20
Generalize cleanup solution

erikbosch:
The previous limitation to only clean branches will give problems when supporting structs, but also if using structured extended attributes.

The problem can be seen by running the included overlay example without the fix, then some unwanted metadata will be printed.

### DIFF
--- a/tests/vspec/test_overlay/expected.json
+++ b/tests/vspec/test_overlay/expected.json
@@ -26,7 +26,24 @@
       },
       "SignalA": {
         "datatype": "int8",
-        "description": "Changing unit to celsius",
+        "dbc": {
+          "changetype": "ON_CHANGE",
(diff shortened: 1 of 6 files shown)


adobekan:
Looks good. I see as well extension for the deployment configuration. I guess that was needed in order to test updates of cleanup improvements.

erikbosch:
> Looks good. I see as well extension for the deployment configuration. I guess that was needed in order to test updates of cleanup improvements.

Yes, the problem that some metadata was kept was detected when we experimented with adding deployment metadata. The metadata in the test files shall just be seen as an example on users can add metadata and how the tool will handle it.

erikbosch:
Meeting notes: Please review until next week

erikbosch:
Meeting decision: Merge

PR #209, opened 2023-01-19
Add py.typed to support mypy

kse3hi:
Fix issue https://github.com/COVESA/vss-tools/issues/208

PR #212, opened 2023-01-24
Run mypy and remove some warnings on skipped methods

erikbosch:
This change results in that mypy is run as part of CI.
Mypy reports as of today some errors. Some of them shall better be fixed, other suppressed.
Currently the build succeeds even if mypy reports errors.
We can later when all errors have been fixed/suppressed change it so the build fails if mypy complains


adobekan:
Interesting report.
"Found 40 errors in 10 files (checked 31 source files)"

adobekan (approved):
Looks good.

erikbosch:
Meeting decision: Merge

erikbosch:
Integrated the change from #209 into this PR as a workaround for the signoff problem. That change is as I see it trivial from an IP perspective so should not be a problem to contribute in my name

PR #216, opened 2023-01-30
Improve error handling on units and datatypes

### LINKED ISSUE #215: Improve error messages related to unit usage
This is related to #214 

I see two needed improvements.

- Complain if the provided unit file does not contain any unit definitions in expected format. If you for instance five a normal *.vspec file as input you should get an error as no types are loaded at all
- If a unit cannot be found, give a more user friendly error (today you get a Python exception)

erikbosch:
Fixes #215

This PR emits understandable errors for 3 scenarios:

- If there is no units found in provided (or default) unit files an error will be given and execution terminate
- If unit lookup fails a reasonable error (unknown unit for signal X) instead of just a KeyError exception
- If datatype lookup fails a reasonable error (unknown datatype for signal X) instead of just a KeyError exception

It also add informational printout on how many units that have been collected. With these changes it should be easier to identify that the unit files your loaded had errors.

### DIFF
--- a/vspec/__init__.py
+++ b/vspec/__init__.py
@@ -24,7 +24,7 @@ from anytree import (Resolver,
 
 from .model.vsstree import VSSNode
 from .model.exceptions import ImpossibleMergeException, IncompleteElementException
-from .model.constants import VSSTreeType
+from .model.constants import VSSTreeType, Unit
 
 nestable_types = set(["branch", "struct"])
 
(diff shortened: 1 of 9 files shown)


erikbosch:
Meeting notes: please review until next week

vspec/__init__.py
```diff
@@ -900,13 +930,20 @@ def check_data_type_references_across_trees(
     nonexistant_types = []
     for _, _, node in RenderTree(signal_root):
         # signals with user-defined data types
+        append_node = False
         if node.is_signal() and node.datatype is None:
-            if not node.does_attribute_exist(
+            if data_type_root is None:
+                # No user-defined types exist
```
erikbosch, line 936: @kkoppolu1 - I updated this PR, rebasing it on your changes and changes unit behavior somewhat according to the discussion in #217.

I then found what seems to be a bug, if you do not specify any datatype file but use something that is not a primitive type (like `int7`) you will get an exception as `data_type_root` is None. In this PR I added a check here, but we could alternatively make sure that data_type_root never is None.

vspec/__init__.py
```diff
@@ -858,6 +858,36 @@ def create_tree_uuids(root: VSSNode):
             namespace_uuid, vss_element.qualified_name()).hex
 
 
+def load_units(vspec_file: str, unit_files : List[str]):
+
+    total_nbr_units = 0
+    if not unit_files:
+        # Search for a file units.yaml in same branch as vspec file
```
kkoppolu1, line 865: comment should read in the same directory?

vspec/__init__.py
kkoppolu1: Indeed, this method should not be called if there is no data type tree specified on the command line. I'll add in the check.

kkoppolu1: I have fixed this in https://github.com/COVESA/vss-tools/pull/232
The change is identical. So, you can either revert this particular change. 
Or, I can wait for this to merge and re-base.

vspec/__init__.py
```diff
@@ -900,13 +930,20 @@ def check_data_type_references_across_trees(
...
+            else:
+                append_node = not node.does_attribute_exist(
                     data_type_root,
                     lambda n: n.data_type_str,
                     lambda n: n.qualified_name(),
-                    lambda n: n.is_struct()):
-                nonexistant_types.append(f'{node.data_type_str}')
+                    lambda n: n.is_struct())
+        if append_node:
+            logging.info(f"Adding {node}")
```
kkoppolu1, line 945: The log statement is not accurate. We are not adding the nodes anywhere.

vspec/__init__.py
```diff
@@ -900,6 +930,7 @@ def check_data_type_references_across_trees(
     nonexistant_types = []
     for _, _, node in RenderTree(signal_root):
         # signals with user-defined data types
+        append_node = False
```
kkoppolu1, line 933: Unused variable. Can be removed.

kkoppolu1 (approved):

PR #217, opened 2023-01-30
Update changelog for -u

erikbosch:
This change highlights what has been changed and also propose way forward. I.e. that we when releasing VSS 4.0 will remove the default unit file from vss-tools. Instead users of tools must explicitly specify unit file, and for example provide the path to the unit file in VSS.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,7 +19,9 @@ It includes changes that are included in released version, but also changes plan
 Overlays introduced to allow customization of VSS. See [documentation](https://covesa.github.io/vehicle_signal_specification/rule_set/overlay/).
 See [vss-tools documentation](https://github.com/COVESA/vss-tools/blob/master/docs/vspec2x.md) on how to include overlays when transforming VSS.
 
-## Planned Changes VSS-Tools 3.1
+## Merged Changes for next release (VSS-Tools 3.1)
+
+*This is changes that has been merged on master but not yet released*
 
 ### Change in UUID handling.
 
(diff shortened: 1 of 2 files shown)


erikbosch:
Meeting notes: Please review - check if you think it is ok that `-u` gets mandatory for VSS 4.0 (planned for end of April if struct feature ready).

SebastianSchildt (approved):
I support this

Potential ux improvement might be that without any given unit file we expect it at a default location&name, directly besides the vspec given on the cmdline? Just a thought. Advantage is, it would still be tied to the model instead of the tools, but for users  (90%) "Just generating VSS standard catalogue" there would  be no need to explicitly type it

erikbosch:
That sounds like a reasonable improvement. I.e. logic like this:

1. If given, use unit file(s) specified by `-u`
2. If not check for  `units.yaml` same location as given *.vspec file (for default build that would mean [`units.yaml`](https://github.com/COVESA/vehicle_signal_specification/blob/master/spec/units.yaml)
3. If not found, for VSS 3.X, the tools will check [vss-tools default unit file](https://github.com/COVESA/vss-tools/blob/master/vspec/config.yaml). This alternative will be removed in VSS 4.0
4. If none of the above found an error will be given.

I will change this one to draft for now, until the changes to support the logic above has been integrated

CHANGELOG.md
```diff
@@ -39,6 +41,19 @@ For VSS-Tools 3.1 the following behavior is implemented:
 The tools vspec2c and vspec2ocf in the contrib folder has been moved to the obsolete folder.
 The background is that they have been broken for a long period and no one has volunteered to fix them.
 
+### Support for specifying unit files
+
+Add new parameter `-u` has been introduced, see [documentation](https://github.com/COVESA/vss-tools/blob/master/docs/vspec2x.md#handling-of-units).
+Use of default unit file deprecated.
+At the same time a unit file has been added to [VSS](https://github.com/COVESA/vehicle_signal_specification/blob/master/spec/units.yaml),
+allowing VSS tooling to control their own units rather than relying on units in VSS-tools.
+Default behavior for units have changed, if there are a file `units.yaml` in the same directory as the `*.vspec`
```
SebastianSchildt, line 50: typo: if there *is* a file

PR #218, opened 2023-02-02
Fix mypy errors and forbid mypy errors

erikbosch:
This PR is a step forward on making Linters and code analysis tools happy.

This one focus on mypy and flake8

**MyPy**

- Making zero mypy errors a criteria for the PR passing CI checks. I think that in general is reasonable.
- Fixing/Suppressing all mypy errors

**Flake8**

- Including flake8 in CI
- Let build still if flake8 report errors - we still have a quite a lot of linter errors

This PR also contains some additional changes:

- Allowing CI to run also for draft pull requests. I think that is reasonable - that allows you to check that your PR actually pass CI before creating a "real PR", i.e. making it ready for review.
- Some linter errors fixed

Signed-off-by: Erik Jaegervall <erik.jaegervall@se.bosch.com>

### DIFF
--- a/vspec/model/vsstree.py
+++ b/vspec/model/vsstree.py
@@ -7,8 +7,8 @@
 # provisions of the license provided by the LICENSE file in this repository.
 #
 
-from anytree import Node, Resolver, ChildResolverError, RenderTree
-from .constants import VSSType, VSSDataType, Unit
+from anytree import Node, Resolver, ChildResolverError, RenderTree  # type: ignore[import]
+from .constants import VSSType, VSSDataType, Unit, VSSConstant
 from .exceptions import UnknownAttributeException, NameStyleValidationException, \
     ImpossibleMergeException, IncompleteElementException
 from typing import Any, Optional, Set, List
@@ -26,16 +26,16 @@ class VSSNode(Node):
 
     type: VSSType
     description = None
-    comment = ""
-    uuid = None
+    comment: str = ""
+    uuid: str = ""
     # data type - string representation. For struct names, this is the fully
     # qualified struct name.
-    data_type_str = None
+    data_type_str: str = ""
     # data type - enum representation if available
     datatype: Optional[VSSDataType]
 
     # The node types that the nodes can take
-    available_types: Set[str] = None
+    available_types: Set[str] = set()
 
     core_attributes = ["type", "children", "datatype", "description", "unit", "uuid", "min", "max", "allowed",
                        "instantiate", "aggregate", "default", "instances", "deprecation", "arraysize",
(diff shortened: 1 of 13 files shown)


erikbosch:
Changing to draft for now, better fixing after refactoring for struct support

erikbosch:
@SebastianSchildt @kkoppolu1 - this is a continuation of what has been added earlier. I did a check on the mypy errors, fixed most but marked some as ignored as I see them as either false positives or deliberate code on our side. I also added flake8 to CI so we can track how many flake8 errors we have in builds, and possibly require new/changed code to not add new errors. I think this fits well into the hook you added Krishna. A few linter fixes added as well.

I was considering if we also shall have pylint, but I expect it to have quite big overlap to flake8. If you look carefully in the PR you can find one ignore-comment "solving" an error reported by pylint, originating from when I did a test with pylint.

Please let me know if you have any concerns/opinions. I have compared output with previous versions of vss-tools, this PR shall not introduce any regressions.

vspec/model/vsstree.py
```diff
@@ -259,6 +259,9 @@ def get_struct_qualified_name(self, struct_name) -> str:
             root = path
 
         # find the struct node under the branch
+        if root is None:
```
kkoppolu1, line 262: When would this trigger? properties are always defined under struct nodes and we have an orphan check for that.

erikbosch: This a typical checker "false positive" - it is not smart enough to know if `root` actually will be None here or not. In this case it it is mypy that complains: 

`error: Item "None" of "Optional[VSSNode]" has no attribute "children"  [union-attr]`

It seems that the reason for this error is that we assign `root = None` before the while loop. I guess that indicates to mypy that we actually not always may find a root. If we remove that line mypy will not complain.

```
        # find the ancestor branch
        root = None
        while not path.is_branch():
            path = path.parent
            root = path
```
But I guess it is also a question on how "bullet-proof" we want the methods to be in class  VSSNode. Do we want them to not crash regardless of how "self" looks like, or do we expect them to be used only in specific context.

Here I can see a few alternatives to make mypy happy.

- Keep the PR as is, i.e. let method return None if "self" has no parent (even if that never should happen as we use the method today)
- Remove the `root = None` assignment. That will mypy happy, but you will get a crash if "self" has no parent (even if that shall not happen as we use the method today)
- Add an `# type: ignore[union-attr]` on line 265. Will make mypy happy, but we will still get a crash if there is no parent. 

I prefer to either have it bullet-prof (i.e. keep as in PR) or remove the assignment and possibly add a comment that we expect there to always be a parent.

kkoppolu1: In another PR, we need to make the callers check if the returned value is `None`

kkoppolu1 (approved):

PR #219, opened 2023-02-03
File re-organization

kkoppolu1:
## Summary
Re-organize the files for better separation of concerns.

### Motivation
As we build more and more features such as struct support into the VSS tree, having clearer file organization helps with code re-usability.

- Exceptions are moved into their own module
- String formatting utilities are moved into their own module
-  Remove unused imports in the files touched

## Testing
- Ran pytest - All unit tests passed successfully
-  Ran GraphQL converter. Output generated successfully for the standard catalog

Signed-off-by: Krishna Koppolu <[krishnachaitanya.85@gmail.com](mailto:krishnachaitanya.85@gmail.com)>

### DIFF
--- a/tests/model/test_vsstree.py
+++ b/tests/model/test_vsstree.py
@@ -1,6 +1,6 @@
 import unittest
 
-from vspec.model.constants import VSSType, VSSDataType, Unit, StringStyle
+from vspec.model.constants import VSSType, VSSDataType, Unit
 from vspec.model.vsstree import VSSNode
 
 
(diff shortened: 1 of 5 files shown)


erikbosch (approved):
Looks good to me. I did a sanity check and did not spot any regressions.

@adobekan @SebastianSchildt - we need to agree on a lean way of handling struct-related PRs. It is tooling so not that important to give "everyone" at least one week for review. OK if only one of us approves?

PR #248, opened 2023-02-27
Remove unit file from vss-tools

erikbosch:
Remove unit file from vss-tools and adapt tooling so that it does not look in vss-tools
Adapt test cases previously relying on vss-tools file


erikbosch:
Meeting notes: merge

kkoppolu1 (approved):

PR #249, opened 2023-02-27
Include linebreak at end of JSON files

erikbosch:
Currently our JSON exporter does not add a newline. As our output is to a file I think ending with a new-line makes sense. That would also solve the problem that our commit-hook tries to add a newline in "expected" file if you try to check them in.

**Important part to discuss - does anyone object that we add a newline at end?**


kkoppolu1 (approved):

SebastianSchildt:
A newline in the standard minified json would be a bug wouldn't it? Otherwise I am fine with it

erikbosch:
Meeting notes:

Sebastian: a commit hook is no good reason

erikbosch:
> A newline in the standard minified json would be a bug wouldn't it? Otherwise I am fine with it

My understanding after some googling:

- If you just serialize minified json there shall be no new line, like if you are to send the JSON as a VISS command
- But if you shall store the minified json in a textfile then there shall be a newline as [POSIX](https://pubs.opengroup.org/onlinepubs/9699919799/basedefs/V1_chap03.html#tag_03_206) requires an end of line character, see https://stackoverflow.com/questions/729692/why-should-text-files-end-with-a-newline

SebastianSchildt:
Not wanting to fight a single byte here, but we did at one line of code to add it. Why should we consider what POSIX defines as a "line"? And the generated file is a .json file. Is that even a "text" file? Are there any practical tools that will break on not having `\n` on EOF?. Also, I think VSS as standard does not necessarily "target" POSIX, that may be the majority of uses, but I could just as easily putting my .json in a z/OS dataset, then it wouldn't even be a "file".

That being said, I can totally live with a newline, if the majority agrees. I just think the arguments for that are pretty weak

erikbosch:
I am on vacation next week, but feel free to discuss anyway. I can live with both with/without line-breaks. In some of the other PRs I have added an exception for the hook for files containing the word `expected` (used in test files), so we can handle both options.

But I noticed that also some other tools does not generate line breaks (at end), and vspec2csv generates two which also upsets the hook. So I think it would be quite good to have a wider discussion - for which tools do we consider a line break as "correct" and for which do we consider it as "incorrect". Then we should better try to improve our tool-documentation, as of today it is not always obvious how output will look like for various input.

adobekan:
07.03.2023 decision to wait one more week

erikbosch:
Meeting notes: Please add a comment with your vote if you want line breaks or not for JSOn files, do it before 16.00 CET Wednesday. Depending on outcome merge or close this one.

SebastianSchildt:
I vote - unsurprisingly -  *against* mandatory line breaks :)

kkoppolu1:
Vote: No (to explicit addition of newline character.)

There are reasons for not having a terminating character enumerated by Sebastian. To add to that, we can just rely on `json.dump` to do the right thing.

erikbosch:
Seems that we have agreed to NOT add line breaks to JSON. I will shortly close/reject this PR

erikbosch:
Result of vote: No linebreak in JSON, rejecting this PR

PR #252, opened 2023-03-02
Check that mandatory object description is present

erikbosch:
Also refactoring existing checks somewhat to check when tree is finished rather than on individual nodes.
That relaxes requirements on individual overlays, I think we say in documentation today that an overlay shall be a valid tree in itself. That is however not strictly needed as it is not intended to be usable by itself. 

With this change the "defect" in VSS 3.1 with a branch without description would had been detected

### DIFF
--- a/tests/vspec/test_description_error/correct.vspec
+++ b/tests/vspec/test_description_error/correct.vspec
@@ -1,0 +1,10 @@
+#
+A:
+  type: branch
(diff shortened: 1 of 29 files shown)


erikbosch:
Meeting notes: Please review, merge decision next week

erikbosch:
Meeting notes: Erik to update parametrization, then ok to merge

PR #254, opened 2023-03-13
Adding go.sum for binary/go_parser/parserlib

erikbosch:
Seems to be needed, and was missed when we added the same file for binary/go_parser


erikbosch:
Meeting notes: Please review until next week

PR #258, opened 2023-03-28
Provide the option to output structs in a separate file or consolidated

kkoppolu1:
## Summary
Allow the user to choose how they want to output structs:
1. Separate files - Current behavior
2. Consolidate with signal file - New behavior

### Consolidation Logic:
1. CSV : Concatenation with the first column using a generic heading called "Node"
2. YAML/JSON : The struct data types are nested under the key "ComplexDataTypes"

## Testing
- Updated unit tests

"Signed-off-by: Krishna Koppolu <krishnachaitanya.85@gmail.com>"


tests/vspec/test_structs/expected-signals-types.yaml
```diff
@@ -0,0 +1,68 @@
...
+  description: A rich sensor with user-defined data type.
+  type: sensor
+
+A.UInt8:
+  datatype: uint8
+  description: A uint8.
+  type: sensor
+  unit: km
+
+CustomDataTypes:
```
erikbosch, line 21: I see that you put the type declarations within "CustomDataTypes". I think putting types in a separate branch makes sense as we then avoid the risk of name conflicts (as we have said that having a type called `A.UInt8` is allowed even if you have a signal called `A.UInt8`). But I would like to get feedback from others and we should also better describe this design decision in the markdown files for vspec2x

vspec2x.py
```diff
@@ -199,11 +202,11 @@ def processDataTypeTree(parser: argparse.ArgumentParser, args, include_dirs,
...
     type information provided in vspec format
     """
-    # check that both required arguments are available to process the data
-    # type tree
-    if args.types_output_file is None or args.vspec_types_file is None:
+    if args.vspec_types_file is None:
         parser.error(
             "Please provide the vspec data types file and the output file")
+    if args.types_output_file is None:
+        logging.info("Sensors and custom data types will be consolidated into one file")
```
erikbosch, line 209: This needs to be reflected in documentation in some way: https://github.com/COVESA/vss-tools/blob/master/docs/vspec2x.md

kkoppolu1: ack. once we align on the usage of the custom keyword, I'll update the documentation in one go.

vspec2x.py
```diff
@@ -199,11 +202,11 @@ def processDataTypeTree(parser: argparse.ArgumentParser, args, include_dirs,
     Helper function to process command line arguments and invoke logic for processing data
     type information provided in vspec format
     """
-    # check that both required arguments are available to process the data
-    # type tree
-    if args.types_output_file is None or args.vspec_types_file is None:
+    if args.vspec_types_file is None:
```
erikbosch, line 205: This shall never evaluate to true as we check this before we call, right? But I do not mind having the check here as well, but does it make sense to continue if this evaluates to true? I.e. shall we add a return after the error below. We also need to rephrase the error message below and remove "and the output file"

kkoppolu1: You are right. The check is redundant. I'll remove it.

SebastianSchildt:
I tested this a little bit, and can confirm it does not break anything when not datatypes are used. I also was able to generate datatypes into the output file, or as a seperate file.

Usability-wise it is not quite clear to me, how this is "supposed" to work,I think this mainly implies missing documentation.

So I see, datatypes are treated differently than  "normal" VSS, i.e. giving a datatype definition as overlay errors out with type struct not known. I think that is fine, but maybe we might want to give an extra hint in the output saying "Is this a datatype definition? Use -vt to load it" or something along those lines.

is it possible to give several -vt? the tool accepts it, but at least in my experiment the output did not change (but Ihave not checked in detail, maybe one definition was included in the other).

In any case: If this is not supported, I think we should (becasue overlays might bring their own datatypes), if it is supported, I think the help text for the argument is not very clear about it.

tests/vspec/test_structs/expected-signals-types.yaml
erikbosch: Meeting notes:

- Ulf: Shall we always use "Custom" as prefix?
- Sebastian: Possible "Complex"?
- Decision: Use "ComplexDataTypes"

erikbosch:
Meeting notes: Ok to merge after krishna has updated name

PR #260, opened 2023-03-29
Testing pre-check on changed files

erikbosch:
We have support for pre-commit but that require that the user do `pre-commit install`.
This CI check will verify that all checks performed by the hook actually have passed

### DIFF
--- a/.github/workflows/pre-commit.yml
+++ b/.github/workflows/pre-commit.yml
@@ -1,0 +1,16 @@
+name: pre-commit
+
+on: [pull_request]
+
+jobs:
+  pre-commit:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+        with:
+          # required to grab the history of the PR
+          fetch-depth: 0
+      - uses: actions/setup-python@v3
+      - uses: pre-commit/action@v3.0.0
+        with:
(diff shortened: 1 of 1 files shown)


.github/workflows/pre-commit.yml
```diff
@@ -0,0 +1,16 @@
...
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+        with:
+          # required to grab the history of the PR
+          fetch-depth: 0
+      - uses: actions/setup-python@v3
+      - uses: pre-commit/action@v3.0.0
+        with:
+          extra_args: --color=always --from-ref ${{ github.event.pull_request.base.sha }} --to-ref ${{ github.event.pull_request.head.sha }}
```
kkoppolu1, line 16: Does this action use the pre-commit config file from the repo?

erikbosch: Yes, it shall basically do the same checks as the pre-commit hook, and only on changed files. As I understand it we cannot force users to actually install the hook. By this check we check that the code is compatible with the hook. In one of the earlier versions I included some dummy blanks (and disabled the hook) and then the check marked the PR as failed as wanted

SebastianSchildt (approved):

PR #265, opened 2023-04-06
Enforce case sensitivity and allow default

erikbosch:
Solves two issues:

- Enforces case sensitivity for type
- Supports default in more cases as specified in https://github.com/COVESA/vehicle_signal_specification/issues/377


erikbosch:
A lot of things to discuss on next VSS meeting, so merging some PRs that are not that critical

PR #277, opened 2023-05-08
Escaping allowed literals that are numbers

erikbosch:
Also a lot of linter fixes


AkhilTThomas:
@erikbosch I don't think this solves the problem completely.
Is the idea that the VSpec will define a set of allowed lets say uint8 values like 1,2,3.. ? If that's the case mapping the _allowed_ keyword in VSpec to IDL enum has to be rechecked.
Firstly  IDLs have to be generated with 
```c
enum SomeEnum{
@value(10) ENUM1,
@value(20) ENUM2,
@value(30) ENUM3
};
```
have to check in detail how this will impact generator for other languages also

erikbosch:
@AkhilTThomas 

It is the recently added signals below that causes problem. So even if they logically represent integer literals, from a strict VSS perspective they are just strings. And theoretically, there is nothing that prevents that there also would be other literals not starting with a digit in the list of allowed values. The use-case could potentially be different if this would be an int-type variable?

```
PidsA:
  datatype: string[]
  type: attribute
  allowed: ["01","02","03","04","05","06","07","08","09","0A","0B","0C","0D","0E","0F","10","11","12","13","14","15","16","17","18","19","1A","1B","1C","1D","1E","1F","20"]
  description: PID 00 - Array of the supported PIDs 01 to 20 in Hexadecimal.


PidsB:
  datatype: string[]
  type: attribute
  allowed: ["21","22","23","24","25","26","27","28","29","2A","2B","2C","2D","2E","2F","30","31","32","33","34","35","36","37","38","39","3A","3B","3C","3D","3E","3F","40"]
  description: PID 20 - Array of the supported PIDs 21 to 40 in Hexadecimal.

PidsC:
  datatype: string[]
  type: attribute
  allowed: ["41","42","43","44","45","46","47","48","49","4A","4B","4C","4D","4E","4F","50","51","52","53","54","55","56","57","58","59","5A","5B","5C","5D","5E","5F","60"]
  description: PID 40 - Array of the supported PIDs 41 to 60 in Hexadecimal.
```

AkhilTThomas:
> 

@erikbosch The problem is that current vspec2ddsidl converts anything with an _allowed_ field to enum, hence it cannot hold string values.

erikbosch:
@AkhilTThomas - I do not know if I understand you correctly. Most (all?) signals in VSS using `allowed` are today using `string`as type, like in the example below. 

```
Sunroof.Switch:
  datatype: string
  type: actuator
  allowed: ['INACTIVE', 'CLOSE', 'OPEN', 'ONE_SHOT_CLOSE', 'ONE_SHOT_OPEN', 'TILT_UP', 'TILT_DOWN']
  description: Switch controlling sliding action such as window, sunroof, or shade.
```

This is currently generated as:

```
{
enum _SwitchValues{INACTIVE,CLOSE,OPEN,ONE_SHOT_CLOSE,ONE_SHOT_OPEN,TILT_UP,TILT_DOWN};
};
struct _Switch
{
string uuid;
_Switch_M::_SwitchValues value;
//const string type ="actuator";
//const string description="Switch controlling sliding action such as window, sunroof, or shade.";
};
```

Do we consider that as a reasonable mapping? I am no expert in DDSIDL but I can see that the information that the base type is "string" is lost. ? I think that mapping has existed for quite some time, and at least idlc does not complain. 

The new signals introduced in VSS 4.0 are similar, just that they use strings that could be interpreted as number. It is quite similar to if we would add `'33'` as an allowed value for `Sunroof.Switch`, like this:

```
Sunroof.Switch:
  datatype: string
  type: actuator
  allowed: ['INACTIVE', 'CLOSE', 'OPEN', 'ONE_SHOT_CLOSE', 'ONE_SHOT_OPEN', 'TILT_UP', 'TILT_DOWN', '33']
  description: Switch controlling sliding action such as window, sunroof, or shade.
```

With current latest master of vss-tools this results in this, which is not accepted by idlc:

```
{
enum _SwitchValues{INACTIVE,CLOSE,OPEN,ONE_SHOT_CLOSE,ONE_SHOT_OPEN,TILT_UP,TILT_DOWN,33};
};
struct _Switch
{
string uuid;
_Switch_M::_SwitchValues value;
//const string type ="actuator";
//const string description="Switch controlling sliding action such as window, sunroof, or shade.";
};
```

How would you prefer the generated IDL to look like for a signal like:

```
Sunroof.Switch:
  datatype: string
  type: actuator
  allowed: ['INACTIVE', 'CLOSE', 'OPEN', 'ONE_SHOT_CLOSE', 'ONE_SHOT_OPEN', 'TILT_UP', 'TILT_DOWN', '33']
  description: Switch controlling sliding action such as window, sunroof, or shade.
```

Either we need to just prefix/escape/whatever the entry `33`, or we need to come up with a bigger refactoring of the allowed-solution for ddsidl. But I assume the latter alternative takes time, so then one need to decide what to with the upcoming VSS/VSS_tools 4.0. I see three alternatives:

* Accepting that the released `vss_rel_4.0.idl` is faulty (i.e. `idlc` complains)
* Do a "least effort" fix to get correct IDL, like in this PR
* Do a major refactoring in how `vspec2ddsidl.py` handle allowed.

I assume the third alternative takes time, so maybe we do not want to block VSS 4.0 until that is implemented.

If no one intends to use the new signals PidsA/PidsB/PidsC from VSS 4.0 together with DDS IDL it should not be that problematic if we generate them in one way in VSS 4.0, but possibly in a different way in a later release.

PR #284, opened 2023-05-22
Make sure vspec2ddsidl can handle repated names

erikbosch:
That is allowed in VSS
But not in DSSIDL, or at least not in idlc

### DIFF
--- a/Pipfile
+++ b/Pipfile
@@ -15,6 +15,7 @@ graphql-core = "*"
 [dev-packages]
 flake8 = ">=6.0.0"
 pre-commit = ">=3.0.4"
+cyclonedds = "*"
 
 
 [requires]
--- a/Pipfile.lock
+++ b/Pipfile.lock
@@ -1,7 +1,7 @@
 {
     "_meta": {
         "hash": {
-            "sha256": "1815aebfa029c9c3e78f74112bfd413c2a9df98162959068b1258521aec85a89"
+            "sha256": "5b85120b496df1317317486b9758513f880757388fe47b68f4043e8e9ce9e8d2"
         },
         "pipfile-spec": 6,
         "requires": {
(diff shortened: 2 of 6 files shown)


erikbosch:
Discussion:

- Sebastian: Isn´t it a problem that it is difficult to guess how DDSIDL will look like
- Sebastian: Is it forbidden according to DDSIDL, or a limitation/problem in CycloneDDS
- Pierre: Simply adding a numeric suffix makes it difficult to find it. Possibly use a character not allowed in VSS. Possibly exclamation mark.
- Erik: Please review and give comments on best way to solve it.

SebastianSchildt:
Maybe 7.5.2 has something say about whether it "should be legal" or not , but not sure, it odes or I understand it correctly 
https://www.omg.org/spec/IDL/4.2/PDF

erikbosch:
In general after reading 7.5.2 it seems that the naming/scoping rules are quite complex; reuse of a name is often considered illegal as it clashes with the name from the higher scope. In DDSIDL names are also case insensitive - in VSS they are case sensitive.

Maybe we need to agree on what approach to have in general.

**Change VSS syntax**

We could change VSS syntax and add  restrictions, like that VSS is case insensitive, that you are not allowed to reuse a name from an upper branch and so on. I.e. if someone define a signal A.A it is not a valid VSS name.

**Define "mandatory guidelines" for VSS standard catalog and VSS-tools**

We could also keep the VSS syntax the same, but define style guides that shall be followed by VSS standard catalog, and also state that it is OK that tools in vss-tools only support VSS following the style guide. That way we can say that `A.A` might be a valid VSS name, but it will not be an acceptable name for the VSS standard catalog and tools like vspec2ddsidl does not need to allow/support names like `A.A`. In some cases it might be easy to accept limitations, like that it is bad practice to have signals where the only difference is the case, like `A.B` and `A.b`. In other cases target-language limitations might be more difficult to accept.

**Require "mandatory" VSS-tools to support "all" allowed VSS**

This is the third approach - all tools must change names to give a valid output. Like in this PR.

erikbosch:
Meeting notes:

- Please review/discuss how to proceed
- Fulup: Main problem is that rules not always are well defined. We need to better describe what is forbidden. Example; if we support dash and dot, a translator may need to replace it with underscore. So then best not to allow two signals where only difference is dash/dot.
- Stephen: Possibly valid to compare with RDF work in VSSo - adapt stricter naming rules, see https://wiki.covesa.global/display/WIK4/Defining+the+COVESA+data+modeling+strategy+and+its+associated+artifacts

erikbosch:
Closing this one for now, general view seems to be that we rather shall have stricter rules, at least for standard catalog

PR #285, opened 2023-05-24
Issue235.Binary tool update for complex datatype support.

UlfBj:
The initial task n this PR, to update the binary tool to support complex data types, led to the question of how an automated verification of the output from the binary tool could be implemented. At least some initial preparation to support that is also done here.

### DIFF
--- a/binary/c_parser/cparserlib.c
+++ b/binary/c_parser/cparserlib.c
@@ -216,7 +216,7 @@ int saveMatchingNode(long thisNode, SearchContext_t* context, bool* done) {
 	if (VSSgetValidation(thisNode) > context->maxValidation) {
 		context->maxValidation = VSSgetValidation(thisNode);  // TODO handle speculative setting
 	}
-	if (VSSgetType(thisNode) != BRANCH &&VSSgetType(thisNode) != STRUCT || context->leafNodesOnly == false) {
+	if (VSSgetType(thisNode) != BRANCH && VSSgetType(thisNode) != STRUCT || context->leafNodesOnly == false) {
 		if ( isGetLeafNodeList == false && isGetUuidList == false) {
 			strcpy(context->searchData[context->numOfMatches].responsePaths, context->matchPath);
 			context->searchData[context->numOfMatches].foundNodeHandles = thisNode;
--- a/binary/c_parser/cparserlib.c
+++ b/binary/c_parser/cparserlib.c
@@ -73,120 +73,36 @@ void printReadMetadata() {
 }
 
 nodeTypes_t stringToNodeType(char* type) {
+    if (strcmp(type, "branch") == 0)
+        return BRANCH;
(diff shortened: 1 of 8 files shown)


binary/c_parser/cparserlib.c
```diff
@@ -300,7 +216,7 @@ int saveMatchingNode(long thisNode, SearchContext_t* context, bool* done) {
 	if (VSSgetValidation(thisNode) > context->maxValidation) {
 		context->maxValidation = VSSgetValidation(thisNode);  // TODO handle speculative setting
 	}
-	if (VSSgetType(thisNode) != BRANCH || context->leafNodesOnly == false) {
+	if (VSSgetType(thisNode) != BRANCH &&VSSgetType(thisNode) != STRUCT || context->leafNodesOnly == false) {
```
erikbosch, line 219: Shall there better be a blank after the `&& `?

erikbosch (approved):
I did a sanity check just using the code with "latest" VSS and no errors were reported. Unfortunately I have no deeper knowledge in the binary tool or in Go, so would appreciate if someone with more knowledge could review as well

From a long term perspective I think two areas could be discussed/improved.

- How does the Go version/tag-syntax in go.mod/go.sum actually work? Do we need to have some form of release handling when releasing new versions of VSS-tools?
- Would it be possible to create some form of tests for vspec2binary, i.e. going one step further than just verifying that the tool can be run without errors? For the struct case a "real example" where you consume a vspec-model with a struct and then verifies that you can find/read/write the struct in the binary representation would be great.

But if no-one else says that they want to review the change I am happy to approve it.

erikbosch:
Meeting notes:

- Ulf: there is a test parser that possible could be used for testing.
- Erik: if you write down some examples using testparser I can check how it can be integrated into CI.
- Erik: Is there anyone else using vspec2binary
- Ulf: Stephen/Renesas use it.
- Stephen: I could have a look
- please review

binary/c_parser/testparser.c
```diff
@@ -22,7 +22,7 @@ long rootNode;
 char* vspecfile;
 
 char* getTypeName(nodeTypes_t type) {
-    switch (type) { 
+    switch (type) {
```
kkoppolu1, line 25: Shouldn't we have a case for "STRUCT" type nodes and "PROPERTY" type nodes?

UlfBj: Good catch! Thanks.

erikbosch:
Meeting notes:

- @UlfBj - are there any remaining work
- OK to merge if approved by @kkoppolu1

UlfBj:
I am not aware of any remaining work.
BR
Ulf

erikbosch:
@kkoppolu1 Do you have objections/remarks?

kkoppolu1 (approved):

PR #287, opened 2023-05-29
Fix flake8 errors and make no errors mandatory

erikbosch:
This PR:

- Fixes remaining flake8 errors for files we supervise
- Changes so that flake8 errors cause build to fail (This is partially redundant as changed file already are checked by pre-commit check)
- Updates one action to reduce NodeJS 12 dependency (for dco-check there is no NodeJS 16 version)


erikbosch:
Meeting notes: please review

PR #293, opened 2023-06-09
PR for discussion - how to keep instance information

erikbosch:


erikbosch:
Meeting notes:

- Erik presented the ideas.
- Please review/discuss

Expanded-info.md
```diff
@@ -0,0 +1,48 @@
...
+
+```
+vss-tools/vspec2json.py -I ./spec --uuid -u ./spec/units.yaml --json-pretty ./spec/VehicleSignalSpecification.vspec expand.json
+```
+
+...gives an extra output `expanded:true` for branches that has been expanded.
+
+I.e. branches like `Vehicle.Cabin.Seat.Row2.PassengerSide` will have `expanded:true`, while regular branches like
+`Vehicle.Cabin` and `Vehicle.Cabin.Seat` will not have any `expanded` attribute.
+
```
BjoernAtBosch, line 23: Disadvantages would be:
 *  The detection of instances is possible, in principle. But it requires extra effort.
 *  The detection might face some limits, when it comes to instances that contain attributes or signals valid for that single instance (or more precisely: a limited set of instances)
 * Digital.Auto is using a json-based VSS catalogue which is "not aware" of instances. There, the "expanded" attribute would need to be maintained manually, which probably leads to additional error cases.

Expanded-info.md
```diff
@@ -0,0 +1,48 @@
...
+In this PR an example for JSOn is shown, running as today:
+
+```
+vss-tools/vspec2json.py -I ./spec --uuid -u ./spec/units.yaml --json-pretty ./spec/VehicleSignalSpecification.vspec expand.json
+```
+
+...gives an extra output `expanded:true` for branches that has been expanded.
+
+I.e. branches like `Vehicle.Cabin.Seat.Row2.PassengerSide` will have `expanded:true`, while regular branches like
+`Vehicle.Cabin` and `Vehicle.Cabin.Seat` will not have any `expanded` attribute.
```
BjoernAtBosch, line 22: Disadvantages would be:
 *  The detection of instances is possible, in principle. But it requires extra effort.
 *  The detection might face some limits, when it comes to instances that contain attributes or signals valid for that single instance (or more precisely: a limited set of instances), only
 * Digital.Auto is using a json-based VSS catalogue which is "not aware" of instances. There, the "expanded" attribute would need to be maintained manually, which probably leads to additional error cases.

Expanded-info.md
```diff
@@ -0,0 +1,48 @@
...
+
+## Alternative 1B: Keep original instance-information (not implemented)
+
+Instead of `expanded:true` we could keep the original expansion information like below, or format it in some other way.
+But that require a bit more work, as today the original expansion information is modfiied during expansion
```
BjoernAtBosch, line 27: This then would be a "mixed" solution?
Means, expanded instances plus original instance information?

Expanded-info.md
```diff
@@ -0,0 +1,48 @@
...
+```
+                "description": "Decorative coloured light inside the cabin, usually mounted on the door, ceiling, etc.",
+                "instances": [
+                  "Row[1,2]",
+                  [
+                    "DriverSide",
+                    "PassengerSide"
+                  ]
+                ],
+```
```
BjoernAtBosch, line 48: I think, this would be our (Eclipse Velocitas') favorite solution, because with that the generated files (json, ...) become semantically identical (except file structure) to the original vspec.

I also tested this with our code generator and it works fine.

BjoernAtBosch:
We (Velocitas) also need to discuss with Digital.Auto, what they think/prefer.

Expanded-info.md
erikbosch: Yes, some form of mixed solution.

erikbosch:
Meeting notes:

- Sebastian: How would overlays work, adding a signal for a specific instance
- Please review

erikbosch:
Added test examples on how this works with overlays. In short, using our overlay mechanism to add signals to expanded paths works, but then the consumer of the output need to do the logic understanding that `Vehicle.Door.Row1.Pos1.NewSignal` actually refer to the signal `Vehicle.Door.NewSignal` in the door instance `Row1.Pos1`.

That is likely Ok for now, but perhaps as part ongoing VSS discussions (this one, ontologies, ...) we possibly need to refactor how we define and use instances, like discussed in https://github.com/COVESA/vehicle_signal_specification/issues/642.

erikbosch:
Meeting notes:

- Erik: Proposal - document limitation "individual instances in overlays" and the merge
- ok

PR #302, opened 2023-09-20
Removing --no-uuid

erikbosch:
This addresses what we previously has written in changelog; that no UUID shall be default and thus no reason to have a --no-uuid argument.

**Shall not be cherry-picked to 4.X branch!**


erikbosch:
Meeting notes:

- Please review

SebastianSchildt (approved):
lgtm 👍

PR #304, opened 2023-10-26
Give error if unit is used for branch or struct

erikbosch:
We do not list `unit` as an allowed keyword for branches in our [documentation](https://covesa.github.io/vehicle_signal_specification/rule_set/branches/). However, as of today we do not check that. This PR addresses that.


erikbosch:
Found some more unit-related parts that better should be improved as part of this PR

PR #305, opened 2023-10-26
Static id generator

erikbosch:
Only for comparison


nwesem:
Hi everyone, if you want to run or test this

To run 
```bash
./vspecID.py ../vehicle_signal_specification/spec/VehicleSignalSpecification.vspec ../output_id_temp.vspec --gen-layer-ID-offset 99 --validate-static-uid ../output_id_v2.vspec
```
For automatic mode add CLA `--validate-automatic-mode`

If you want to see how the pipeline works you can change one entry of the validation file here `../output_id_v2.vspec` like this
```
Vehicle.Chassis.Wheelbase:
  datatype: unit8
  staticUID: '0x0003F764'
  type: attribute
  unit: m
```
Manually changed datatype, staticUID and unit so most cases are triggered

To run the new tests:
```bash
cd path/to/your/vss-tools
export PYTHONPATH=${PWD}
# to tun all tests
pytest tests/
# to run only new tests
pytest tests/vspec/test_static_uids/
```

erikbosch:
Some thoughts - it would be good to have some documentation that describes common use cases, like:

- How to run it first time
- How to maintain the "generated" vspec/yaml file.
- How to integrate a new VSS version with added or renamed signals.

adobekan:
> * Thin name-based APIs, like KUKSA gRPC or VISS where access is by name and one client may use the old name, and another use the new name when doing set/get/subscribe. Then the server must know that these two names mean the same thing

Where do we store that information so implementation can know how to handle the data? 

> * Thin id-based APIs. If the client (or producer) converts name to "static id" before sending to server then server only needs to know static id, name does not matter.

I think that this is quite similar. It is just matter of where do you keep mapping details.
 
I see two open points.
1. How do we generate UID, hash or just sequence value? I would prefer short version of 24-bits, max 32-bits and anyhow keep the information in the generated file. Hash value is also fine as long as implementation is checking for collision and validating existing IDs.
2. Where do we keep the mappings details of version changes? I prefer separated file outside of VSS that is not related to specific implementation.

docs/vspecID.md
```diff
@@ -0,0 +1,103 @@
...
+They will be used as unique identifiers to transmit data between nodes. The static UIDs are
+implemented to replace long strings like `Vehicle.Body.Lights.DirectionIndicator.Right.IsSignaling`
+with a 4-byte identifier.
+
+ToDo what kind of data does each byte represent using latex
+
+## General usage
+
+```bash
+usage: vspecID.py [-h] [-I dir] [-e EXTENDED_ATTRIBUTES] [-s] [--abort-on-unknown-attribute] [--abort-on-name-style] [--format format] [--uuid] [--no-expand] [-o overlays]
```
erikbosch, line 13: Two thoughts:

- Shall we use `vspec2...` naming pattern like for other tools, i.e. `vspec2id`?
- Here we can remove a lot of arguments that does not make sense to use, like `--format format` and `--json-pretty`

vspec/vssexporters/vssidgen.py
```diff
@@ -0,0 +1,153 @@
...
+import logging
+import os
+from typing import Dict
+from vspec import load_tree
+from vspec.model.constants import VSSTreeType
+from vspec.loggingconfig import initLogging
+from vspec.model.vsstree import VSSNode
+from vspec.utils import vssidval
+import yaml
+
```
erikbosch, line 23: Add the snippet below. That indicates that this "tool" works with `--no-expand` which it does. (Whether it is useful is a different topic). If you do not think `--no-expand` shall be supported just return false

```

def feature_supported(feature_name: str):
    """Return true for supported arguments/features"""
    if feature_name in ['no_expand']:
        return True
    return False
```

vspec/utils/vssidval.py
```diff
@@ -0,0 +1,449 @@
...
+                    logging.warning(
+                        f"UNIT MISMATCH also, but a new id for '{k}' has already "
+                        "been assigned."
+                    )
+                else:
+                    logging.warning(
+                        "[Validation] "
+                        f"UNIT MISMATCH: Units of '{k}' in unit: '{v['unit']}' "
+                        f"in the current specification and the validation file "
+                        f"'{k}' in unit: '{validation_node.unit}' don't "
```
erikbosch, line 156: Using `f"'{k}'`means that the new name is written, not the old, see other comment

erikbosch:
There seems to be a limitation - if you both add a new signal and change name, then the checks seems to bail out. The same actually seems to happen if you just add a new signal.

Original:

```
Wheelbase:
  datatype: uint16
  type: attribute
  default: 0
  unit: mm
  description: Overall wheelbase, in mm.

#
# Axle definition
#
Axle:
  instances:
    - Row[1,2]
  type: branch
  datatype: uint8
  description: Axle signals
```

Changed to:

```
# Adding a new signal, all fields differ

NewSignal:
  datatype: float
  type: sensor
  unit: inch
  description: Neues Signal.

# Changing to "Wheelbaze", all other things equal
Wheelbaze:
  datatype: uint16
  type: attribute
  default: 0
  unit: mm
  description: Overall wheelbase, in mm.

# No change to axle
Axle:
  instances:
    - Row[1,2]
  type: branch
  datatype: uint8
  description: Axle signals
```

Gives the following output:

```
WARNING  [Validation] NAME CHANGE: The name of the node in the current vspec was changed from 'Vehicle.Chassis.Wheelbase' to 'Vehicle.Chassis.NewSignal'. This is a non-breaking change if the static UID stays the same. Continuing...
WARNING  DESCRIPTION MISMATCH: The description of Vehicle.Chassis.NewSignal has changed from 
	   Validation: 'Overall wheelbase, in mm.' to 
	   Current vspec: 'Neues Signal.'
WARNING  [Validation] UNIT MISMATCH: Units of 'Vehicle.Chassis.NewSignal' in unit: 'inch' in the current specification and the validation file 'Vehicle.Chassis.NewSignal' in unit: 'mm' don't match which causes a breaking change, so we need to assign a new id!
INFO     Assigned new ID '0x0004F263' for Vehicle.Chassis.NewSignal
WARNING  DATATYPE MISMATCH also, but new id for 'Vehicle.Chassis.NewSignal' has already been assigned
WARNING  [Validation] NAME CHANGE: The name of the node in the current vspec was changed from 'Vehicle.Chassis.Axle' to 'Vehicle.Chassis.Wheelbaze'. This is a non-breaking change if the static UID stays the same. Continuing...
WARNING  DESCRIPTION MISMATCH: The description of Vehicle.Chassis.Wheelbaze has changed from 
	   Validation: 'Axle signals' to 
	   Current vspec: 'Overall wheelbase, in mm.'
WARNING  [Validation] UID MISMATCH: IDs don't match. Current tree's node 'Vehicle.Chassis.Axle' has static UID '0x0003FC63' and validation tree's node 'Vehicle.Chassis.Axle' has static UID '0x0003FB63'!
WARNING  What would you like to do?
1) Assign new ID
2) Overwrite ID with validation ID
3) Skip


```

Also removing a signal seems to make the check bail out. Then you get warnings on all subsequent signals.

```
WARNING  [Validation] UID MISMATCH: IDs don't match. Current tree's node 'Vehicle.Chassis.Axle' has static UID '0x0003FA63' and validation tree's node 'Vehicle.Chassis.Axle' has static UID '0x0003FB63'!
```

nwesem:
Hi @erikbosch thanks a lot for your comments, they were a great help! I am currently working on a hashed pipeline, that we can compare to get current sequential numbering pipeline. I pushed it to a different branch until I'm done with the implementation. I will let you know when it's ready

nwesem:
Hi @adobekan and @erikbosch 
after some discussion I finished implementing a first draft of the hashing for the static UIDs. We decided to expect the user to write a new attribute called `fka` (formerly known as) in case the user wants to keep the same UID (semantic name change or path change). So as an example if we have a node called `OldName` and we update it to `NewName` we would extend the vspec like this:
```
A.B.NewName:
  datatype: uint32
  type: sensor
  unit: mm
  description: A.B.NewName's old name is 'OldName'. And its even older name is 'OlderName'.
  fka: ['A.B.OldName', 'A.B.OlderName']
```
 `fka` is a list of strings so it can be renamed more than once.

I used the test.vspec in the tests to verify my implementation. It would be very interesting to hear your opinion on the current state. If you like this approach I would go ahead and add full functionality to this. Currently we are checking for breaking changes in `name`, `unit`, `datatype` and `type`, so if you change one of the these you should receive a breaking change warning log.

For non-breaking changes we currently only track the description which also gives a corresponding log but it doesn't affect the current UID.

If you want to test the semantic name changes you can use this test tree as an input for the first generation of and `output_id_hash.vspec`

To ease testing I added a `temp_test.vspec` in the test directory of the current implementation, so you can follow these commands to trigger a `SEMANTIC NAME CHANGE`.

first generate the UID file (which will also be the validation later) with `temp_test.vpsec` which contains the node `A.B.OldName`.
```bash
./vspec2id.py tests/vspec/test_static_uids/test_vspecs/temp_test.vspec ../output_hash_test.vspec --gen-no-layer --use-fnv1-hash
```
and then validate using the generated ID file and the `test.vspec` that contains the renamed node `A.B.NewName`
```bash
./vspec2id.py tests/vspec/test_static_uids/test_vspecs/test.vspec ../output_hash_test_v2.vspec --gen-no-layer --use-fnv1-hash --validate-static-uid ../output_hash_test.vspec
```
You should now see the normal logs from vspec2x plus a `SEMANTIC NAME CHANGE` warning.

If you have any other question just let me know! Thank you

nwesem:
UPDATE: I added the remaining attributes that cause breaking changes, so you should now see multiple `WARNING` logs when running the id generator and validator as described above. You would expect the following warnings:

```
...
WARNING  [Validation] SEMANTIC NAME CHANGE or PATH CHANGE for 'A.B.NewName'.
WARNING  [Validation] BREAKING CHANGE: There was a breaking change for 'A.B.IsLeaf' which means it is a new node or name, unit, datatype, enum or min/max has changed.
WARNING  [Validation] BREAKING CHANGE: There was a breaking change for 'A.B.Min' which means it is a new node or name, unit, datatype, enum or min/max has changed.
WARNING  [Validation] BREAKING CHANGE: There was a breaking change for 'A.B.Max' which means it is a new node or name, unit, datatype, enum or min/max has changed.
...
```

erikbosch:
Meeting notes:

Adnan informed:
- Now Supporting 4 byte hash sync. Adnan to check on collision checks
- Sequence id removed from public PR, now only hash
- Generates static id before expansion
- Documenting what is considered breaking changes
- Documenting handling on files
- use FNv-1
- Shall be ready for review.

Erik: Please review, wants to make a merge decision next week to be able to handle 4.1 pre-release and release before Christmas.

nwesem:
Hi everyone, the latest changes have not been merged to this branch. I was waiting on input to merge my changes to here. Before reviewing please give me at least until tonight in the evening to merge it. Thank you

nwesem:
Merged the most recent changes to this branch now, last thing that we are missing is a `NON-BREAKING CHANGE` log for adding and removing attributes which I will add to the implementation asap.

erikbosch:
@adobekan @nwesem - there are many commits in this PR, so I think a squash before merging makes sense. Do you want to do the squash or should I do it when I merge it?

erikbosch:
Considered good enough for now, some additional tests will be performed as part of 4.1 release preparations

erikbosch:
Merged and squashed through #313 
Closing this one

PR #311, opened 2023-11-22
Tool restructuring

erikbosch:
This is an idea I have for restructuring of vss-tools. Ass of today we have [vspec2x.py](https://github.com/COVESA/vss-tools/blob/master/vspec2x.py) as main entry point. It explicitly lists all exporters and import all of them as well. This has some drawbacks:

- If you want to create your own tool you must customize `vspec2x.py`
- The vss-tools pypi package has dependencies for all exporters. You need graphql dependencies even if you not intend to use graphql.

An alternative solution would be to instead have `vspec2x` as an abstract tool that require an exporter as input. That would make it easier to create you own proprietary tool/exporter, you do not need to modify `vspec2x.py`, instead you could just create you own exporter and give it as an argument when calling vspec2x. This PR shows how that could look like for vspec2json. The file `vspec2x_abstract.py` is intended to replace `vspec2x.py`

If we go this path we could as a second step consider refactoring our [vss-tools PyPI package](https://pypi.org/project/vss-tools/) so that exporters with "heavy" dependencies (big size, or dependencies that only are supported on a limited set of platforms, or dependencies to less user friendly licenses) are put in a separate PyPI package. A possible example could be graphql. We could also possibly add generic exporter functionality/hooks so that exporters can add or control additional semantic checks or processing steps if needed. A typical example could be the recent discussion to expand signal tree based on allowed units. Maybe that is a feature that is possible or even mandatory for one exporter, but totally useless for others.


SebastianSchildt:
Like the idea, I guess @adobekan  would agree, as we have discussed a similar idea.

One main advantage would also be that you can "pip install " vss-tools, import sutff and create your custom exporter without touching COVESA upstream code. Today users are basically "forced" to fork, and I have seen such internals forks flying/rotting around. Resistance to pick up new features upstream would likely be lower, if you just need to upgrade a dependency

vspec2x_abstract.py
```diff
@@ -0,0 +1,172 @@
...
+class Vspec2XAbstract:
+
+    def __init__(self, exporter):
+
+        self.exporter = exporter
+
+    def main(self, arguments):
+        initLogging()
+
+        parser.add_argument('-I', '--include-dir', action='append', metavar='dir', type=str, default=[],
```
SebastianSchildt, line 35: I think it would be good to change logic, that somehthing like vspec2json "owns" the parser for command line arguments and can include/instanctiate vpsec2x with a feature array (do I know instances/structs/uuid/extended attributes), and then vspec2x would "add" generic options , and leave options out that a tool does not support anyway

erikbosch: It would be quite easy to do some changes of the current solution. Today we have if , if looking for instance at [vss2json.py](https://github.com/COVESA/vss-tools/blob/master/vspec/vssexporters/vss2json.py#L21) two methods:

```
def feature_supported(feature_name: str)
def add_arguments(parser: argparse.ArgumentParser):
````

Where `add_arguments` is for adding exporter-specific arguments,  and ` feature_supported` to answer yes/no on features like "no expand" (for struct we have a different mechanism), but I agree that passing in some form of config array would be a nicer solution, as we then could exclude non-supported arguments already when you do `-h`.

erikbosch:
Meeting notes:

- Adnan: We have some tools that are not really exporters, but would be good if they are "close" to vss-tools. How to handle them
- Adnan: Separate exporters from other tools
- Erik: we could have contributed tools (not only exporters) in repo.
- Adnan: Guidelines on how to add contributed tools
-

erikbosch:
Ready for review new.

erikbosch:
@adobekan - could you take a look and check if you have any objections to the content of this PR. If not I will merge it.

erikbosch:
Meeting notes:

- Please review
- Merge decision next meeting unless any objections

erikbosch:
MoM: Ok to merge, after the static id PR has been merged

PR #321, opened 2023-12-20
to lowercase before hash

adobekan:


erikbosch (approved):
LGTM

erikbosch:
@adobekan - is the only fix you want to have merged for 4.1, or is there more to expect?

PR #325, opened 2024-01-19
Adding datatype checks

erikbosch:
If a unit specifies allowed datatypes we will check that only those types are used


erikbosch:
MoM: Merge

PR #327, opened 2024-01-30
Add manual dispatch for build and update action versions

erikbosch:
Old actions use deprecated Node.js version


SebastianSchildt (approved):
very good, some housekeeping 👍

PR #335, opened 2024-03-14
Fix hard-coded username in pipenv installation instructions

ppb2020:
Copy/pasting the instructions for installing pipenv fail because of a hard-coded /home/user, forcing user to modify the instructions before executing the command. This can be addressed by using the ${USER} environment variable.

Signed-off-by: Pierre Pierre Blais <ppblais@blackberry.com>

### DIFF
--- a/README.md
+++ b/README.md
@@ -118,7 +118,7 @@ Install this project and its dependencies in the local `.venv` folder in this pr
 ```sh
 export PIPENV_VENV_IN_PROJECT=1 # will create a local `.venv` in the project, otherwise uses global location
 pip install pipenv
-export PATH=/home/user/.local/bin:$PATH
+export PATH=/home/${USER}/.local/bin:${PATH}
 pipenv install --dev # install the development dependencies as well
 ```
 


erikbosch:
The failed DCO check likely depends on that the "author field" of the commit does not match sign-off. Have there possibly been any rebase/edit in Github UI. I anyway do not see this as a real blocker, so no reason to spend too much time on fixing it.

image

erikbosch (approved):

ppb2020:
Weird; I used the same sign-off as I had used n my last commit...

Looks like if "ppb2020 [66322482+ppb2020@users.noreply.github.com]" was expected... Perhaps I should use this in the future?

erikbosch:
> Weird; I used the same sign-off as I had used in my last commit...
> 
> Looks like if "ppb2020 [[66322482+ppb2020@users.noreply.github.com](mailto:66322482+ppb2020@users.noreply.github.com)]" was expected... Perhaps I should use this in the future?

Nah, I think it just expects them to match. But I have seen "authors" like that previously when someone has done anything in the UI, like explicit/implicit rebase/squash and similar. But no big deal as I see it - there is a signoff and the signoff shows your real name so I consider the reported error more like a false positive

ppb2020:
Ah, yes. As this was a very simple patch and only affecting documentation, I used the GitHub Web GUI to make the change as opposed to making the change locally on my development host. That was certainly a difference from my previous submissions.

erikbosch:
MoM: Merge

Issue #339, opened 2024-04-02
Possible naming conflict -o and -ot


erikbosch:
As of today we support `-o` for overlays for `--ot` for type output when using structs. When working on enabling `-ot` only for cases when actually relevant I noticed that `-ot` then can get intepreted as `-o t` and you get a totally irrelevant error message.

I see some options for vss-tools 5.0:

- Use a different "short name" for `-ot`with no conflicts, like `-t`
- Remove the "short name", i.e. always request `--types-output-file`
- Accept that error message becomes odd if using `-ot` on a tool that does not support it

```
erik@debian4:~/vss-tools/tests/vspec/test_structs$ ../../../vspec2ddsidl.py -vt VehicleDataTypes.vspec -u ../test_units.yaml -ot expt.json test.vspec exp.json
usage: vspec2ddsidl.py [-h] [-I dir] [-e EXTENDED_ATTRIBUTES] [-s] [--abort-on-unknown-attribute] [--abort-on-name-style] [--uuid] [-o overlays] [-q quantity_file] [-u unit_file]
                       [-vt vspec_types_file] [--all-idl-features]
                       <vspec_file> <output_file>
vspec2ddsidl.py: error: unrecognized arguments: exp.json
erik@debian4:~/vss-tools/tests/vspec/test_structs$ 
```

ppb2020:
I note that there is a -vt options that may have the same issue. It's nice to have the short options, but you can run out of letters...

Does the option parsing code we use allow for non-conflicting shortening of long options? That is, would --abort be allowed for --abort-on-unknown-attribute if there is no other option that starts with --abort? If so, not having short options for optional options that are not not used often would work well.

Using -t instead of -ot is one way around if short options are necessary, but may be harder or confusing to remember. Would -O be another option.

erikbosch:
I think it is recommended to only use single characters for short options, so both `-vt` and `-ot` are "not recommended". Reason is that arg parser consider for example `-x3` and `-x 3` to be equal. `-vt` is currently not a problem as we have no `-v`

erikbosch:
MoM:

- Erik presented the problem
- PP: OK to go for only long name here
- Erik: But if so backward incompatible change, so only for 5.0

Issue #238, opened 2023-02-16
Struct support - vspec2ddsidl


erikbosch:
This issues concerns struct support for vspec2ddsidl.

Required work:

- Define/propose how it shall be handled (possibly in this PR)
- Implement and test
- Make sure that expected output format is described in documentation
- If feasible, add at least one test case checking output 

_VSS Struct support is expected to be officially released for VSS 4.0, late April. To identify and manage remaining a work number of issues have been created to facilitate more detailed discussion and to identify volunteers for implementing/reviewing/testing._

@SebastianSchildt  will check if the ones originally contributing vspec2ddsidl are interested

happikin:
I am willing to take up this feature addition. Is there any document or diagram which may help me briefly understand the flow of code in the py files.

erikbosch:
> I am willing to take up this feature addition. Is there any document or diagram which may help me briefly understand the flow of code in the py files.

Great!

We do not have that much documentation, there is a little at:

- https://github.com/COVESA/vss-tools/blob/master/docs/vspec2x.md#writing-your-own-generator
- https://github.com/COVESA/vss-tools/blob/master/docs/vspec2x_arch.md

To get know a little about the general structure of exporters/tools you can look at the test case at https://github.com/COVESA/vss-tools/blob/master/tests/generators/example_generator.py

I did however notice that we already did some work on DDSIDL support in https://github.com/COVESA/vss-tools/pull/262, so this issue is not really up to date. Please check if the current level support fits your needs. I noticed that not all test cases have been updated to include DDSIDL struct support, I will do some edits and mention you so that you can get an overview of the current level of support.

erikbosch:
See #340 for an example on how DDSIDL struct support works as of today, that could help identifying improvement areas. If we think it is good enough we should close this issue

PR #340, opened 2024-04-02
Add ddsidl struct tests and make -ot configurable

erikbosch:
This is inspired by:

- https://github.com/COVESA/vss-tools/issues/238
- https://github.com/COVESA/vss-tools/issues/339

This PR does two things:

- It adds a testcase with struct output for DDSIDL
- It extends the tool config so that you can indicate that a tool does not support `-ot`


erikbosch:
MoM: Please review

erikbosch:
MoM:Merge

Issue #194, opened 2022-10-20
AMM Idea: Overlays -removal of nodes


erikbosch:
Idea collected at AMM. It would be valuable to be able to specify that a node/branch shall be removed by an overlay

ppb2020:
There is some inconsistency in the documentation on how to refer to components of a node such as "datatype", "unit", etc. In some cases, these are referred to as "members" and as "elements" in other cases. Perhaps this should be cleaned up. For the purposes of this discussion, I will be using "element".

Removing nodes by deleting them from VSPEC files has licensing implications as it requires the person or company making the deletions to publish the updated VSPEC files. A mechanism that allows for the deletion or removal to be done through overlays would remove this requirement as overlays were specifically designed to support changing the signal catalog.

A possible approach to support deletion is to add a formal element to the definition of nodes that indicates whether the node is active or not. The name of the element would reflect the state of the node. Something like active, deleted, etc. For the purposes of this discussion, let's call it "active". This element could be defined as part of the VSS or introduced through overlays.

Early investigation suggested this "active" element would be a boolean value that could be set through overlays to indicate that the node is inactive. In this manner, post-overlay application tooling (such as the vss-tools that convert VSPECs to other formats) could remove the nodes marked as inactive and only leave active nodes. A branch marked as inactive would cause all nodes below this branch to be considered inactive, whatever their "active" element setting would be. This would support very simple deletion of irrelevant nodes, such as combustion engine nodes for EVs. In this case, the element could also be removed from each node in the catalog that is used at runtime, reducing memory and file usage.

Further investigation revealed that deletion of nodes through tooling may in some cases be premature. That is, in the interest of supporting dynamically inactivating or activating nodes at runtime, it would be preferable for some inactive nodes to not actually be deleted, but remain present in the runtime, such that they can be dynamically activated (hence my choice of "active" as the name of the element instead of "deleted"). For example, this would facilitate the support of "pay-as-you-go" vehicle features for which signals could come and go dynamically at runtime. This adds capability at the cost of more memory or file system usage.

Eventually, a combination of deletion and activation may be needed or useful. In that case, perhaps a better name for the new element would be "status", with the following values: "active", "inactive", "deleted". With this approach, "deleted" signals could be removed by tooling, while still allowing support for dynamic activation or inactivation at runtime. This would achieve some memory and file system savings, while providing unique functionality for dynamic activation of vehicle features.

[...]

PR #342, opened 2024-04-03
add removal of nodes / branches

nwesem:
This PR tries to solve issue https://github.com/COVESA/vss-tools/issues/194, by adding a `delete` attribute. So in the base specification or from overlays, we can now set `delete: true` on a branch or node which will delete them from the created tree.

Please refer to [./docs/vspec2x_arch.md](/docs/vspec2x_arch.md) to see an example. I added tests for all exporters but still would appreciate it if you gave this a quick spin to check if all necessary functionality has been implemented.

Thanks! Looking forward to your comments


erikbosch:
There is a conflict, I would suggest that you squash and rebase

docs/vspec2x_arch.md
```diff
@@ -22,6 +19,28 @@ This will result in that e.g. `Vehicle.Cabin.Door` is expanded to the following
 For some exporters expansion can be suppressed by using the `--no_expand` option.
 Then instance information will be represented by other means in the resulting output.
 
+## Deletion / Node removal
+
+Nodes can be removed from the tree by using the `delete` keyword in the overlay.
```
ppb2020, line 24: There is a lot of variation in how node elements (type, datatype, min, max, etc.) are called. Various terms are used, such as "attribute" (confusing, since some nodes have a type of "attribute"), member, element, keyword, etc.

Refer to https://github.com/COVESA/vehicle_signal_specification/issues/733 for a description of the issue.

Given the current recommendation (yet to be approved) is to use the term "element", I would suggest that you use this for now at least.

docs/vspec2x_arch.md
```diff
@@ -22,6 +19,28 @@ This will result in that e.g. `Vehicle.Cabin.Door` is expanded to the following
...
+This is useful when a signal is not used in a next version of the specification or if you
+simply want to delete a node or branch from the specification, e.g. you are not using a signal of the base
+specification.
+Please note, that the deletion for branches will delete all nodes that are connected to that branch (which is
```
ppb2020, line 28: Should make it clear that adding the boolean delete element set to true to a branch node causes all sub-nodes to be ignored and thus their delete element is irrelevant.

docs/vspec2x_arch.md
```diff
@@ -22,6 +19,28 @@ This will result in that e.g. `Vehicle.Cabin.Door` is expanded to the following
...
+desired behavior).
+
+If you want to delete a node or branch simply add the following to the specification/overlay, let's say the vehicle only
+has two doors in the front. In this case we would like to delete the signals for the rear doors:
+
+```
+Vehicle.Cabin.Door.Row2:
```
ppb2020, line 35: This example is possibly incorrect as it refers to a Row2 instance, which does not exist, per se, in the catalog. You may need a better example. Or, is it intentional that the deletion would support deleting specific instances? If so, I am not sure why the instance would be created in the first place...

As well, this raises a very interesting issue for which I will raise a separate issue or discussion point related to the MPL-2.0 license. Section 1.10b states that the license covers: "any new file in Source Code Form that contains any Covered Software". Given that one must provide in the overlay file the node path name of the node to be deleted, the license requires that this overlay file be made available upon request and that the license terms apply to this new file. The same applies to an overlay that modifies an existing signal (say, to change the unit or set a different number of instances).

This defeats the goals of providing the overlay mechanism to allow third-party implementors of a catalog to create their own version independent of the VSS version and license. We probably need a broader discussion of the license.

Indeed, the text at https://covesa.github.io/vehicle_signal_specification/license/ mentions that deletions are not considered to run afoul of the license but these statements are external to the license itself and, even if COVESA has no intention to enforce this part of the license, it remains that lawyers are generally worried by such statements. It's generally preferable to address this in the license itself. Anyway, a separate discussion to this PR, to be continued once I open a separate issue or discussion.

docs/vspec2x_arch.md
```diff
@@ -22,6 +19,28 @@ This will result in that e.g. `Vehicle.Cabin.Door` is expanded to the following
...
+  description: All doors, including windows and switches.
+  type: branch
+  delete: true
+```
+
+By adding the `delete: true` to the branch `Vehicle.Cabin.Door.Row2` all nodes and branches connected
+to `Vehicle.Cabin.Door.Row2:` will be deleted.
```
ppb2020, line 42:
```suggestion
to `Vehicle.Cabin.Door.Row2:` are deleted by vss-tools when converting to a different format.
```

tests/vspec/test_node_removal/test_files/units.yaml
```diff
@@ -0,0 +1,22 @@
...
+units:
+  mm:
+    label: millimeter
+    description: Distance measured in millimeters
+    domain: distance
+  m:
+    label: meter
+    description: Distance measured in meters
+    domain: distance
+  degrees/s:
```
ppb2020, line 11:
```suggestion
  degree/s:
```

ppb2020 (commented):
Some of my comments lead to a broader discussion on licensing. Shall I open a discussion or issue to track it?

docs/vspec2x_arch.md
nwesem: > This example is possibly incorrect as it refers to a Row2 instance, which does not exist, per se, in the catalog. You may need a better example. Or, is it intentional that the deletion would support deleting specific instances? If so, I am not sure why the instance would be created in the first place...

Up for discussion, to me it seemed like a good example to show that even deleting instances would work if `delete` is used from overlay, but maybe you are right and I should split it to show multiple examples:
I would add **one example for deleting a node**, **one for deleting a branch** and should we add **one for the instances** or how does deleting instances work currently? Maybe I'm missing some information here, could you explain? Let's say a car only has only one row instead of two, would this be removed from the base `VehicleSignalSpecification.vspec`?

tests/vspec/test_node_removal/test_files/units.yaml
nwesem: This one is copied from the main units.yaml and I just confirmed that `degrees/s` is correct

nwesem:
> There is a conflict, I would suggest that you do a squash and rebase. But apart from that it looks good.

Hi @erikbosch, I'm having trouble to finish the rebase because of the obsolete and tests directories, is there a small chance that in [this merge](https://github.com/COVESA/vss-tools/commit/f3aa9532924767700260222d7be8d618a37b32c2) pre-commit wasn't run or that the upgrade of pre-commit is causing it? Same thing occurs when I run pre-commit hooks in the current master branch..

erikbosch:
The Android part merged isn't conformant with pre-commit requirements (it is old code we do not want to touch). I can do a try and check how it works for me.

erikbosch:
As there are merge-commits I think a rebase might trigger pre-commit checks, and they do not pass for the Android-related part. This could possibly be solved by first squashing. I tried this (upstream is official COVESA repo)

```
git fetch upstream pull/342/head:pr342
git fetch upstream
git checkout pr342
# Squash all by interactive rebase
git rebase -i 469bfa346b27e8
# Then rebase on latest master
# A few easily fixed conflicts
git rebase upstream/master
# After fixing them
git rebase --continue
```
I uploaded the result for reference and created a dummy-PR in my own fork to verify that build works

https://github.com/erikbosch/vss-tools/pull/4

nwesem:
thx @erikbosch, I tried it and it seems to be working fine. I found a bug though, currently deleting from overlay doesn't work as expected, will get back asap.

docs/vspec2x_arch.md
erikbosch: @ppb2020 - concerning the license topic - there have been some previous discussions on license for VSS as some seem to prefer for example Apache over MPL due to concerns similar to yours. It is anyhow a separate discussion, check with @paulboyes if you want to discuss it further.

ppb2020: > @ppb2020 - concerning the license topic - there have been some previous discussions on license for VSS as some seem to prefer for example Apache over MPL due to concerns similar to yours. It is anyhow a separate discussion, check with @paulboyes if you want to discuss it further.

Wilco.

nwesem:
Is there anyone that could help me solve this bug? 

I added two more test "overlay" files that are supposed to remove a test node or branch from a base test.vspec but even though the overlay is loaded correctly and the `delete` element is detected as `True`, the `delete` element doesn't get overwritten by the overlay and therefore the node or branch is not deleted. Am I using it wrong? Do you have any ideas? I temporarily pushed the code here https://github.com/nwesem/vss-tools/pull/5

erikbosch:
@nwesem - I took a look and I think I found the problem

The problem seems to be that you delete the node from the overlay before merge, not from the final tree after merge. If you modify your function like this so that it actually does not remove the node you will see that the removal check will be called 3 times. Once before the merge and twice during the merge (I think)

```
    def check_for_removal(self):
        """Caution uses base node as input."""
        for node in LevelOrderIter(self):
            if node.delete:
                logging.info(f"Node {node.qualified_name()} will now be deleted - HAHA NOT!.")
                # node.parent = None
                # node.children = []
```

I think you should run the deletion after the merge. I did a test adding it to the function below and together with disabling `check_for_removal` it seems to give the right result, and the check is only performed once. You likely do not want to add the deletion to that function, at least not without renaming it, so some refactoring might be needed. If you experience other "strange merge errors" consider applying #346 to make sure it is not related to that error.

```
def verify_mandatory_attributes(node, abort_on_unknown_attribute: bool):
    """
    Verify that mandatory attributes are present.
    Need to be checked first after overlays (if any) have been applied, as attributes are not
    mandatory in individual files but only in the final tree
    """
    if isinstance(node, VSSNode):
        if node.delete:
            logging.info(f"Node {node.qualified_name()} will now be deleted - BUT NOW!.")
            node.parent = None
            node.children = []
        node.verify_attributes(abort_on_unknown_attribute)
        for child in node.children:
            verify_mandatory_attributes(child, abort_on_unknown_attribute)
```

docs/vspec2x_arch.md
nwesem: thx done

nwesem:
thx a lot @erikbosch, I have to get used to the codebase a bit more to solve these things, obviously it was deleted before the merge :grinning:

nwesem:
created a squashed and rebased version of this PR here https://github.com/COVESA/vss-tools/pull/348

therefore closing this PR

Issue #341, opened 2024-04-02
update pre-commit hooks


nwesem:
Hi everyone, I realized that we are not using the full functionality of `pre-commit`, so I'm suggesting that we update it to use all standard hooks for python-based projects. I extended the current one with `pyupgrade`, `isort`, `black`, `ruff`, `markdownlint-cli2`, and `codespell`. Please see suggested change below.

Regarding `pyupgrade` we could use python 3.8+ for now eventhough I would recommend to go for at least 3.10+

As this changes many files of the project I created this issue for discussion. 

```
# See https://pre-commit.com for more information
# See https://pre-commit.com/hooks.html for more hooks
repos:
  - repo: https://github.com/pre-commit/pre-commit-hooks
    rev: v4.5.0
    hooks:
    -   id: trailing-whitespace
        exclude_types: ["csv", "proto"]
    -   id: end-of-file-fixer
        exclude_types: ["json", "proto"]
        exclude: ".*\\.[fidl|jsonschema]"
    -   id: check-yaml
    -   id: check-added-large-files
  - repo: https://github.com/pycqa/flake8
    rev: '7.0.0'
    hooks:
    -   id: flake8
  - repo: https://github.com/asottile/pyupgrade
    rev: v3.3.1
    hooks:
      - id: pyupgrade
        args: [ --py38-plus ]
  - repo: https://github.com/pycqa/isort
    rev: 5.12.0
    hooks:
      - id: isort
        args: ["--profile", "black", "--filter-files"]
  - repo: https://github.com/psf/black
    rev: 23.10.1
    hooks:
      - id: black
  - repo: https://github.com/astral-sh/ruff-pre-commit
    # Ruff version.
    rev: v0.1.1
    hooks:
      - id: ruff
  - repo: https://github.com/DavidAnson/markdownlint-cli2
    rev: v0.12.1
    hooks:
      - id: markdownlint-cli2
        args:
          - --fix
  - repo: https://github.com/codespell-project/codespell
    rev: v2.2.4
    hooks:
      - id: codespell
```

erikbosch:
Some thought from my side - In general I am positive but we need to make sure that Continuous integration continue to work. Maybe update to some 3.10 version in files like https://github.com/COVESA/vss-tools/blob/master/Pipfile and https://github.com/COVESA/vss-tools/blob/master/.github/workflows/buildcheck.yml

erikbosch:
MoM:

- Niclas: Will give it a try, but many files changes
[...]

PR #343, opened 2024-04-03
Updating Python version to 3.10.6

erikbosch:
This is related to #341 and the discussion there; updating CI to use 3.10.6, selected as that is the current version used by Ubuntu 22.04 which is used by ubuntu-latest in Github runners


erikbosch:
MoM:

- Please review
- Possible merge decision next meeting

PR #344, opened 2024-04-04
Protobuf use static uid

eriksven:
This PR extends the protobuf generator to use the static-uid which would be created in a previous step with vspec2id generator. The PR also adds the feature to make all fields in the generated protobuf file to optional. 

Using the static-uids has some drawbacks but the previously existing approach of using incremental field identifiers has some issue regarding backwards compatibility. Because of that I want to propose the alternative solution of using the static-uids. More details on the advantages and drawbacks of both approaches are available in the new vspec2proto.md Documentation.


vspec/vssexporters/vss2protobuf.py
```diff
@@ -127,18 +127,41 @@ def traverse_signal_tree(tree: VSSNode, proto_file):
...
+            if dt_val.endswith("[]"):
+                data_type = "repeated " + data_type
+            elif add_optional:
+                data_type = "optional " + data_type
+        else:
+            data_type = node.qualified_name("")
+            if add_optional:
+                data_type = "optional " + data_type
+        if static_uid:
+            fieldNumber = int(int(node.extended_attributes['staticUID'], 0) / 8)
```
erikbosch, line 149: Could we possibly try to add a more user friendly message if a node does not have a `staticUID`.

Like first checking if there is a `staticUID` and if not give a user friendly message, like "to use --static-uid option you must first ..." and then exit

```
erik@debian4:~/vehicle_signal_specification$ ./vss-tools/vspec2protobuf.py -u ./spec/units.yaml --strict --add-optional --static-uid ./spec/VehicleSignalSpecification.vspec new.proto
INFO     Calling exporter...
INFO     Generating protobuf output...
  File "/home/erik/vehicle_signal_specification/./vss-tools/vspec2protobuf.py", line 25, in <module>
    vspec2x.main(sys.argv[1:])
  File "/home/erik/vss-tools/vspec/vspec2x.py", line 155, in main
    self.generator.generate(args, tree, self.vspec2vss_config, data_type_tree)
  File "/home/erik/vss-tools/vspec/vssexporters/vss2protobuf.py", line 193, in generate
    traverse_signal_tree(signal_root, f, config.static_uid, config.add_optional)
  File "/home/erik/vss-tools/vspec/vssexporters/vss2protobuf.py", line 130, in traverse_signal_tree
    print_message_body(tree_node.children, proto_file, static_uid, add_optional)
  File "/home/erik/vss-tools/vspec/vssexporters/vss2protobuf.py", line 149, in print_message_body
    fieldNumber = int(int(node.extended_attributes['staticUID'], 0) / 8)
KeyError: 'staticUID'
```

docs/vspec2proto.md
```diff
@@ -0,0 +1,98 @@
+# vspec2proto - vspec Protocol Buffer generator
+
+The vspec2proto.py script generates [Protocol Buffer message definitions](https://protobuf.dev) for all nodes in the tree. You can use these proto definitions to serialize the data of a VSS tree, e.g., as part of a gRPC API.
+
+## Example
+
+```bash
+./vspec2protobuf.py outputIds_v2.vspec indentity.proto  -q ../vehicle_signal_specification/spec/quantities.yaml -u ../vehicle_signal_specification/spec/units.yaml  --static-uid --add-optional
```
erikbosch, line 8: Do you also get warnings like:

`WARNING  Attribute(s) staticUID in element Major not a core or known extended attribute.`

In https://github.com/COVESA/vss-tools/blob/master/docs/vspec2id.md#generate-eg-yaml-file-with-static-uids they include `-e staticUID` in examples to avoid warnings

I am thinking if we promote `staticUID` to a `core_attributes` so that you never get a warning. Thoughts @adobekan

eriksven: Yes, I do get these warnings too. I thought this would be "normal" since `staticUID` is not a `core_attributes`. So thanks for the hint of how to avoid the warnings. I will have a look into it.

erikbosch: Short term I think you could just add `-e staticUID` in the example. But I am thinking of the long term solution where I see multiple alternatives:

- Promote `staticUID` as an allowed/core attribute for all tools always
- Add it in some way as core-attribute for the specific tool, regardless of whether `--static-uid` is used or not
- Add it only if  `--static-uid` is used

For the last two alternatives we should find a common mechanism that can reused by all tools.  I anyway not see it as a topic for this PR.

adobekan:
@eriksven 
I'm not sure if directly comparing static IDs with field numbers at the end makes sense. If we're using static IDs, the message description would likely be different. Additionally, I'm uncertain if we should describe any aspect of VSS in that area.

erikbosch:
MoM:

- Please review, possible merge decision next week

vspec/vssexporters/vss2protobuf.py
eriksven: Good point, I added a check and logging statement.

docs/vspec2proto.md
eriksven: The documentation now contains an example with the parameter `-e staticUID`.

erikbosch:
MoM:

- Adnan - If I would use static id I would not this approach, it would be more useful if one would use generic message. Would be good to have SvenErik available for a discussion. 
- Erik: Could be useful if server use 4.2, client 4.1
- Adnan: But then we could use a thin API (id+value), no need for fat API, but both approach valid, but we need to document both options.
- Erik: Possibly discuss next week
- Gunnar: Could it help if @eriksven describe intention with change better

PR #346, opened 2024-04-05
Fix overlay merge error

erikbosch:
This PR solves a problem when a signal is added in an overlay and the signal is "deeper" down in an instantiated tree. A hypothetical example below, adding a temperature signal for front left brake in an overlay. Current result is that `Temperature` signal is not included in the final result.

```
Vehicle.Chassis.Axle.Row1.Wheel.Left.Brake.Temperature:
  datatype: float
  type: sensor
  description: Brake Temperature FL
  dbc2vss:
    interval_ms: 1000
    signal: BrakeTempFL3FE
    transform:
      math: "floor(x-40)"
```

### DIFF
--- a/tests/vspec/test_overlay_on_instance/expected.json
+++ b/tests/vspec/test_overlay_on_instance/expected.json
@@ -202,6 +202,91 @@
                 "description": "Changed description, shall inherit comment",
                 "type": "sensor",
                 "unit": "km"
+              },
+              "TT": {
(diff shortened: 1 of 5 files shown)


erikbosch:
MoM:

- Please review until Thursday EOB
- Merge Friday if no remarks

erikbosch:
MoM: Merge

PR #347, opened 2024-04-08
Add version printout

erikbosch:
We have sometimes experienced problems that you run the "wrong" VSS-tools version due to faulty paths. This should hopefully help that somewhat by printing version number.

The idea is to have a hardcoded version number similar to https://github.com/COVESA/vehicle_signal_specification/blob/master/VERSION, i.e. it needs to be manually updated.

- Advantage: Simple, no need to have something fancy based on git tag or similar. Will work out of the box also for pypi package and other packinging
- Disadvantage: Must be mentioned in the release instructions so we remember to update it 

```
erik@debian4:~/vehicle_signal_specification$ make csv
./vss-tools/vspec2csv.py -u ./spec/units.yaml --strict ./spec/VehicleSignalSpecification.vspec vss_rel_$(cat VERSION).csv
INFO     VSS-tools version 5.0-dev
INFO     Calling exporter...
INFO     Generating CSV output...
INFO     All done.
erik@debian4:~/vehicle_signal_specification$ ./vss-tools/vspec2csv.py -v
5.0-dev
erik@debian4:~/vehicle_signal_specification$ ./vss-tools/vspec2csv.py --version
5.0-dev
```


vspec/vspec2x.py
```diff
@@ -38,6 +40,7 @@ def main(self, arguments):
         initLogging()
         parser = argparse.ArgumentParser(description="Convert vspec to other formats.")
 
+        parser.add_argument('-v', '--version', action='version', version=VERSION)
```
ppb2020, line 43: -v is often used for --verbose

Consider using -V for --version or not defining a short-form option?

vspec/vspec2x.py
```diff
@@ -22,6 +22,8 @@
 import sys
 import vspec
 
+VERSION = '5.0-dev'
```
ppb2020, line 25: Consider putting the definition into a file that can be included or read in here as well as included automatically in documentation or release notes?

erikbosch: We could have it in a separate file like in the VSS repo, but that would require some extra work.
Possibly integrate it in some way with what we already have in https://github.com/COVESA/vss-tools/blob/master/setup.py. But that would require some extra work and verification to make sure it works as expected in all deployment alternatives with have (like venv and pypi)

erikbosch:
PR updated to use version from https://github.com/COVESA/vss-tools/blob/master/setup.py

erikbosch:
I saw a comment from previous VSS meeting on automatic versioning, we could possibly do that with setuptools. What we need to do then (for vss-tools) is to change tag format slightly as it need to comply with pypi rules. You will also if running locally built versions get somewhat odd results like in the example below showing that you use something based on tag 4.4.4, plus 0 extra commits, plus some dirty (changed) code¨. Versions are also sometimes cached

```
erik@debian6:~/vehicle_signal_specification$ make csv
./vss-tools/vspec2csv.py -u ./spec/units.yaml ./spec/VehicleSignalSpecification.vspec vss_rel_$(cat VERSION).csv
INFO     VSS-tools version 4.4.4.post0+git.c7fe8e4c.dirty
```

erikbosch:
Updated to use git version. I did a test locally by tagging my last commit with `5.0.0.dev0`and creating https://pypi.org/project/vss-tools/5.0.0.dev0/ . Then installing it from pypi and verifying that correct version is used. Works well.

If we are to use this method we should possibly after each release on master create a tag like `5.0.0.dev0` so that you on local builds see a tag based on 5.0....... rather than on last release.

erikbosch:
MoM:

- OK to merge

PR #352, opened 2024-05-08
add const UID attribute to allow constant IDs

nwesem:
Hi this PR adds `constUID` attribute to allow overwriting generated static UIDs, as for static UID this is a 4-byte hex identifier. Please see test `test_const_id` for more details


tests/vspec/test_static_uids/test_static_uids.py
```diff
@@ -27,14 +27,24 @@
...
+            + " ./out.vspec --validate-static-uid "
+            + "./validation_vspecs/validation.vspec "
+        )
+    else:
+        return (
+            "../../../vspec2id.py "
+            + test_file
+            + " ./out.vspec --validate-static-uid "
+            + "./validation_vspecs/validation.vspec "
+            + "--only-validate-no-export"
```
erikbosch, line 46: To me it is not obvious why you have `"--only-validate-no-export"` here but not for the overlay alternative. Possibly add a comment.

erikbosch:
Should usage possibly be described in https://github.com/COVESA/vss-tools/blob/master/docs/vspec2id.md

vspec/model/vsstree.py
```diff
@@ -68,6 +68,7 @@ class VSSNode(Node):
         "$file_name$",
         "fka",
         "delete",
+        "constUID",
```
erikbosch, line 71: We should discuss how integrated we want the id-attributes to be in vss-tools. By adding `constUID` (and previously `fka`) they will be considered as core attributes and accepted in all tools.

There are two alternatives, one is to consider them as extended attributes and give warning unless they are specified by `-e`, the other is to refactor vss-tools so that each tool can add their own core attributes. Or alternatively that we consider them as "global core attributes", but then we should possibly mention them in https://github.com/COVESA/vss-tools/blob/master/docs/vspec2x_arch.md

nwesem: @erikbosch can you discuss this during the public meeting today? I have been sick yesterday / today, but can finish this tomorrow if you decide on how you want it to work. I would appreciate it, thank you!!

nwesem:
> Should usage possibly be described in https://github.com/COVESA/vss-tools/blob/master/docs/vspec2id.md

Thanks for the hint! I will add documentation as soon as we decided on how this will be integrated

erikbosch:
MoM:

- Erik and Niklas to sort out remaining steps, then ok to merge

vspec/vssexporters/vss2id.py
```diff
@@ -68,7 +70,15 @@ def export_node(yaml_dict, node, id_counter, strict_mode: bool) -> Tuple[int, in
...
     """
-    node_id, id_counter = generate_split_id(node, id_counter, strict_mode)
+
+    if not node.constUID:
+        node_id, id_counter = generate_split_id(node, id_counter, strict_mode)
+    else:
+        assert node.constUID.startswith(
+            "0x"
+        ), f"constUID has to begin with '0x': {node.constUID}"
+        assert len(node.constUID) == 10, f"Invalid constUID: {node.constUID}"
```
nwesem, line 80: I am considering to move these two `assert` statements to a utlils function and also run them on the generation of UIDs as an additional safety measure..

nwesem: Found a better way to solve this here https://github.com/COVESA/vss-tools/pull/352/commits/f46dd0c2251d69c73a8d14869d528a4db9387196

erikbosch:
I assume there is still some work remaining, like documentation. Correct @nwesem?
If possible it would be great if you could finish during the day, as this is the last PR intended to be included in 4.2, so that I could prepare a release candidate tomorrow.

nwesem:
thx for letting me know @erikbosch, I'm currently working on the documentation. Will be done soon. Would it be possible to include my suggestion for this issue https://github.com/COVESA/vss-tools/issues/354 in the release candidate as well? I can do it for the shown `min` and `max` variables and send another PR by the end of the day. It's quite important to make static UID file iteration work.

nwesem:
@erikbosch quick question: to fix the bug mentioned in https://github.com/COVESA/vss-tools/issues/354 I will branch off from here and merge back into this branch `feat/const-id`? Otherwise we would probably run into merge conflicts as I would be changing the same files.. do you agree that makes sense?

erikbosch:
That works fine, no problem to use the same branch, or to use the "same" commit in multiple branches. Use what is easiest for you, as long as it is understandable what is the intended final result I have no problems fixing minor merge issues when a PR is to be merged.

nwesem:
There is an issue. I implemented the default values for `min` and `max` and updated everything regarding the static UIDs but setting the default value causes 43 other tests to fail (AssertionErrors) because now `min` and `max` values are `null` or `None` instead of the empty string `""`.

@erikbosch Do you have a second to take a look and help me on this? I pushed it to branch https://github.com/nwesem/vss-tools/tree/fix/var-init 
The other option is to find a somewhat hacky solution to make static UIDs work for 4.2 release and then properly solve it for 4.3 release. I gave a hacky solution a quick try, I could make it work but yeah that's definitely not good practice :smiley: 
What do you think? It could also be worth spending the second to update the tests according to the new default values for `min` and `max` but I cannot estimate how much work that would be..

Also fyi @adobekan

nwesem:
Hi @erikbosch I found a decent way to temporarily fix it here https://github.com/COVESA/vss-tools/pull/352/commits/ad0b093338f389dbb7bfb7697af0ceed50a526e9 which means this PR is ready for 4.2 release. 
You can disregard the comment above for the moment. However, I will look into the initialization issue asap, but that needs more time. Let me know if this works for you.. Thx!!

erikbosch:
Thanks @nwesem - that works for me. I will take a look at the PR and let you know if I have any concerns.

PR #355, opened 2024-05-17
Add min/max handling tests

erikbosch:


erikbosch:
@nwesem I added some more test cases on min/max for both int and float. If we would change min/max internal handling later this test case should at least catch any regressions we introduce to the most frequently used tools.

(_This test by the way use the generic setup where the python file one level up search for subdirectories with no python file inside and perform a generic generation and comparison_)

nwesem:
Great idea! Looks good to me

PR #356, opened 2024-05-17
allow floats for min/max in vspec2id

nwesem:
@erikbosch one last minute addition on the vspec2id, sry about this

This change allows integers and floats for min max values


erikbosch (approved):
LGTM

Issue #353, opened 2024-05-15
backwards compatibility test doesn't clean up generated files


nwesem:
Just a minor change but we could clean up the generated files in the compatibility tests after they ran using a fixture like this:

```
@pytest.fixture(scope="function", autouse=True)
def delete_files(change_test_dir):
    yield None
    os.system("rm -f out.json out.txt")
```

erikbosch:
Great proposal. Feel free to create a PR or add a commit to an existing one.

PR #358, opened 2024-05-22
Cleanup after compat test

erikbosch:
Previously out files were not removed. Also some minor refactoring

### DIFF
--- a/tests/backward_compatibility/test_compatibility.py
+++ b/tests/backward_compatibility/test_compatibility.py
@@ -18,6 +18,11 @@ def change_test_dir(request, monkeypatch):
     monkeypatch.chdir(request.fspath.dirname)
 
 
+@pytest.fixture(scope="function", autouse=True)
+def delete_files(change_test_dir):
+    yield None
+    os.system("rm -rf out.json out.txt vehicle_signal_specification")
+
 # Test all VSS versions we support
 #
 # Intended workflow:
@@ -33,6 +38,8 @@ def change_test_dir(request, monkeypatch):
 # * If not add limitation to compatibility section in README and remove '
 #   unsupported versions from the test case
 #
+
+
 @pytest.mark.parametrize("tag",
                          [
                           'v4',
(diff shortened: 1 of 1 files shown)


erikbosch:
@nwesem feel free to review

Fixes #353

nwesem (commented):
we haven't been doing this but could be type hinted like this

```
...
def change_test_dir(request: pytest.FixtureRequest, monkeypatch: pytest.MonkeyPatch):
...
```

tests/backward_compatibility/test_compatibility.py
```diff
@@ -43,7 +50,6 @@ def test_compatibility(tag, change_test_dir):
     Test that we still can analyze wanted versions without error
     """
 
-    os.system("rm -rf vehicle_signal_specification")
     os.system("git clone --depth 1 --branch " + tag +
```
nwesem, line 53: you could move this to the fixture above the yield so it is only cloned once in the beginning, no?

nwesem: but I just realized that that would mean you would have to change the scope of the `delete_files` fixture function to `"module"`..

nwesem: or would it be `"package"`, I'm currently not sure https://docs.pytest.org/en/6.2.x/fixture.html

nwesem: and obviously switch branch depending on tag, but please ignore if this creates any overhead

erikbosch: Would indeed be doable. Lets see if I get time. As we only have one test case here the advantage might be limited, but we could possibly use it as playground and later reuse it for other files

erikbosch: I did some tests, the `monkeypatch.chdir` command has function scope, but even if refactoring like below it is faster to do 3 shallow clones for a single branch than doing a full clone. So it does not seem to be worth the effort right now

```
@pytest.fixture(scope="module", autouse=True)
def initiate_test_env(request):
    os.chdir(request.fspath.dirname)
    os.system("git clone https://github.com/COVESA/vehicle_signal_specification")
    yield None
    os.system("rm -rf out.json out.txt vehicle_signal_specification")
    os.chdir(request.config.invocation_params.dir)
```

Issue #193, opened 2022-10-20
AMM Idea: Overlays - introduce easy solution to add keys to multiple signals


erikbosch:
Idea collected during AMM. In some cases you might want to add the same key/property to a large number of signals, e.g. add something like "included: true" to a large number of signals. A "shortcut" solution for doing this could be useful

nwesem:
hi @erikbosch can you elaborate a bit on this? Do you already have sth in mind how a signal like this would be defined? 
I could take some time to implement this

erikbosch:
Hi @nwesem - I will try to explain some ideas that I had at that time. The problem as of today is that you need quite many lines to change a signals, like using `delete` below.

```yaml
Vehicle.Service.TimeToService:
  type: sensor
  datatype: int32
  delete: true
```

(Parts of it is due to tool limitations, that we for instance require `type` even if we theoretically could manage without for overlays). Anyhow, if we would like to use `delete` for a large number of signals it would be great with a method where you only need a single line for each extra signal. One idea could be to support something like below where you define the wanted change (add/delete) and then just list all signals that shall be involved. Would likely be quite easy to implement **if** we support defining overlays without specifying type so that the tool just could expand it. But one could think of other ideas/solutions/syntax as well. The challenge is likely to find something that can be implemented without adding to much complexity 


```

delete_some:
  type: overlay
  add:
     delete: true
  signals:
    - dsdsd
    - dsssd
    - ds
    
define_safety:
  add:
     asil: qm
  signals:
    - a
    - d
    - f
    
remove_unit:
  remove:
     unit:
  signals:
    - xxx
    - yyy
    - zzz
```

nwesem:
I've been trying to come up with a way to solve this, but you are right the way the overlays are currently loaded (just like any vspec using `load_tree(...)`), we need to at least provide type and datatype.

To me this seems like a two step process:

1. Find a way to merge overlay that doesn't define datatype and type to a currently loaded tree and then
2. add the keys to all signals defined under the condition that the signals already exist in the currently loaded tree

What if we defined an additional way of integrating overlays in a currently loaded tree by checking if all signals have datatype and type or not. If they don't provide datatype and type we assert that all defined signals in the overlay already exist in the currently loaded tree. If all signals defined in the overlay already exist, we can change only the data defined in the overlay. But that also means that we would have to deal with _two types_ of overlays from now on.

Is there a better way to solve this? Please let me know what you think @erikbosch @adobekan

PR #360, opened 2024-05-29
Simplify overlay handling

erikbosch:
This is an idea for simplifying overlays. If a node is found in overlay without type (e.g. `branch`, `sensor`) or datatype (e.g. `uint8`) it will try to do a look-up if a node with the same name already exist, and then use that type/datatype value.

There are limitations, like if you add something to `Vehicle.Axle.Front.SomeSignal` in an overlay (i.e. only for a single instance) it will not find source signal `Vehicle.Axle.SomeSignal`.

Many of the changes are just adding license headers. As I got failure on some changed files I also updated some files that were not changed.

Next step if this is merged is to adapt documentation in VSS repo. Possibly saying something like: 

- _In most cases you do not need to specify `datatype` and `type` in overlay if you are modifying an existing signal._
- _Specifying `datatype` and `type` is needed if overlay addresses specific instances of a branch, like front left wheel._

### DIFF
--- a/tests/vspec/test_faulty_type/test_faulty_type.py
+++ b/tests/vspec/test_faulty_type/test_faulty_type.py
@@ -24,7 +24,7 @@ def test_error(change_test_dir):
     assert os.WIFEXITED(result)
     # failure expected
     assert os.WEXITSTATUS(result) != 0
-    test_str = 'grep \"Unknown type: bosch\" out.txt > /dev/null'
+    test_str = 'grep \"Type not allowed: bosch\" out.txt > /dev/null'
     result = os.system(test_str)
     os.system("cat out.txt")
     os.system("rm -f out.json out.txt")
--- a/tests/vspec/test_overlay/expected_no_datatype.json
+++ b/tests/vspec/test_overlay/expected_no_datatype.json
@@ -1,0 +1,26 @@
+{
+  "A": {
+    "children": {
+      "AA": {
+        "children": {
+          "SignalAA": {
+            "datatype": "int8",
+            "description": "This is yet another signal.",
+            "type": "sensor",
+            "unit": "km"
+          }
+        },
+        "description": "Branch A.AA.",
+        "type": "branch"
+      },
+      "SignalA": {
+        "datatype": "int8",
+        "description": "No datatype OK if signal exist since before",
+        "type": "sensor",
+        "unit": "celsius"
+      }
+    },
+    "description": "Branch A.",
+    "type": "branch"
+  }
+}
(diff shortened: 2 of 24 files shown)


nwesem:
thx @erikbosch this is a great addition to implement issue https://github.com/COVESA/vss-tools/issues/193

I tested it on a clean install of vss-tools and is working as expected on non-struct signals. I haven't looked into the mentioned limitations on structs yet

nwesem:
@erikbosch I tested the structs and can confirm that I get the same warnings and errors. Now I was wondering what your planned next steps are on this topic?

erikbosch:
I did not had time (or forgot) to mention the PRs in vss-tools yesterday, but as you have already looked at it I see no problems merging this one. I have no problems updating documentation, if you have time and interest in something for #193 feel free to try it out, or if you want to improve the limitations.

nwesem:
hi @erikbosch was there a specific reason why you didn't implement the limitations for structs? I would like to proceed with it to make https://github.com/COVESA/vss-tools/issues/193 work. An idea how you would overcome the limitations would be a valuable first step for me as I haven't worked much with the code that deals with structs.

erikbosch:
For structs (i.e. type files) I did not really did any tests so there I do not really know how much effort, if any, that is required.

For instances we can take the example in https://github.com/COVESA/vehicle_signal_specification/pull/747

```YAML
# Type and Datatype not needed
Vehicle.Speed:
  unit: m/s

# Type and Datatype needed as overlay affects a particular instance
Vehicle.Occupant.Row1.DriverSide.HeadPosition.Yaw:
  type: sensor
  datatype: float
  unit: mm
```

Here the problem with the second one is that when we read the overlay we have not yet analyzed/expanded instances, so you can easily find `Vehicle.Occupant.HeadPosition.Yaw` in tree but you would need a more advanced search algorithm to find `Vehicle.Occupant.Row1.DriverSide.HeadPosition.Yaw` as `Row1` and `DriverSide` comes from instance information. Like finding `Vehicle.Occupant` and then check if `Row1` refers to a real sub-branch or rather to an instance. This is a bit related to the discussion in [#749](https://github.com/COVESA/vehicle_signal_specification/issues/749),

nwesem:
hi @erikbosch, sry about this, I accidentally called instances structs in this whole conversation and just realized it now.. I was referring to instances the entire time, as structs are not being used (yet) as far as I know.
I'm currently trying to come up with a search algorithm to find the definition of a signal that was instantiated to link its datatype and/or type to it when it is being overwritten from an overlay..

nwesem:
hi @erikbosch would it be possible to have a short 15-30min call about finding the type (and datatype) for instances? I've been trying to find a way to search for them but as this data is cleaned from the model before rendering a tree, I can't find a slick solution to retrieve the type and datatype. I would really appreciate it!!

erikbosch:
@nwesem - we could try to align on a slot using the COVESA slack channel, if that works for you

PR #366, opened 2024-06-21
Using rich for logging format, added .envrc

sschleemilch:
# About

Using [rich](https://rich.readthedocs.io/en/stable/introduction.html) as logging formatter for readability.

Before:
image

After:
image

Also added an [direnv](https://github.com/direnv/direnv) config for python3. Added gitignore entry accordingly

### DIFF
--- a/.envrc
+++ b/.envrc
@@ -1,0 +1,1 @@
+layout python3
--- a/.gitignore
+++ b/.gitignore
@@ -80,6 +80,7 @@ celerybeat-schedule
 
 # virtualenv
 .venv/
+.direnv/
 venv/
 ENV/
 
--- a/Pipfile
+++ b/Pipfile
@@ -11,6 +11,7 @@ deprecation = ">=2.1.0"
 pytest = ">=2.7.2"
 PyYAML = ">=5.1"
 rdflib = ">=6.0.0"
+rich = ">=13.7.1"
 vss-tools = {editable = true, path = "."}
 graphql-core = "*"
 importlib-metadata = ">=7.0"
--- a/Pipfile.lock
+++ b/Pipfile.lock
@@ -1,7 +1,7 @@
 {
     "_meta": {
         "hash": {
-            "sha256": "7f4f92123b5aa8439474f87981d5b801d6c2aac053c721fef8555c2e1c9d8094"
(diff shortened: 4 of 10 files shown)


sschleemilch:
Fixing the tests was a bit ugly since they are checking for stdout which changed now with rich.
Rich does adapt the output to the width of the terminal. If there is no terminal (such as in CI) the default value is 80.
So manually setting `COLUMNS` to 200 for the tests.
Overall I think the tests could be improved a lot.

Thoughts about them:

- Using `os.system` should be avoided and be replaced with `subprocess.run`
- Calling a subprocess could be avoided completely by just importing the module and running its `main()`
- Using host tools like `grep` and `cat` strikes out Windows support by design and are really not necessary `subprocess.run` can capture output and then a string check is quite simple
- Writing tests that are checking against stdout is not reliable and should be replaced by patching and looking at the Python logging module to see what messages have been received. That of course implies that we are importing the modules and not calling a new interpreter
- Not using a context based tempdir for tests but always removing things by hand is quite cumberstone. `with tempfile.TemporaryDirectory() as tmp` could be used to do that automatically

erikbosch:
MoM: Please review

sschleemilch:
Any findings? If not, mergy merge? :octocat:

erikbosch:
> Any findings? If not, mergy merge? :octocat:

Unless a PR is trivial we typically give people a week for reviewing, i.e. announcing at one VSS Tuesday meeting, and then taking a merge decision at the next, or the day before if there is nothing reported. For larger ones possibly even longer review time. 

In this case (as it is quite technical) i do not think discussing it again at next meeting makes sense, but I will wait until Monday to see if anyone has any objections.

sschleemilch:
> > Any findings? If not, mergy merge? :octocat:
> 
> Unless a PR is trivial we typically give people a week for reviewing, i.e. announcing at one VSS Tuesday meeting, and then taking a merge decision at the next, or the day before if there is nothing reported. For larger ones possibly even longer review time.
> 
> In this case (as it is quite technical) i do not think discussing it again at next meeting makes sense, but I will wait until Monday to see if anyone has any objections.

Okay, I thought this ones pretty trivial but all good

erikbosch (approved):
Looks good to me

erikbosch:
The weekend has passed and no-one has filed any additional comments. Looks good to me, merging so that you can rebase the other PR.

PR #367, opened 2024-06-24
Feature/poetry

sschleemilch:
# About

Restructured project to have a clear src and test layout. Using poetry as venv, dependency and release helper.

Benefits:
- Clear separation of src and tests
- Tools are available as "binary" names, e.g. `vspec2json` instead of `vspec2json.py`
- Tests are simpler by having clean package imports `from vss_tools.something import something`. Tests do not need to know where tool sources are `../../../vspec2json.py` -> `vspec2json`

Benefits of poetry:
- Poetry simplifies the process of managing project dependencies. It automatically resolves dependencies, handles version conflicts, and ensures consistency across different environments.
- Reproducible Builds: Poetry generates a poetry.lock file that captures the exact versions of all dependencies used in your project. This ensures that your project can be reproduced consistently across different environments.
- Virtual Environments: Poetry creates and manages virtual environments for your project, ensuring that dependencies are isolated and do not interfere with other projects on your system. Of course this is also possible manually but poetry makes it easier.
- Packaging and Distribution: Poetry makes it easy to package your Python project and distribute it as a distributable artifact (e.g., a wheel or source distribution) that can be installed by others. Simple wrappers for that with `poetry build` and `poetry publish`
- Dependency Resolution: Poetry's dependency resolver is more robust and reliable than the default pip resolver, handling complex dependency graphs and resolving conflicts more effectively.
- Cross-Platform Compatibility: Poetry is designed to work consistently across different operating systems, ensuring that your project can be easily set up and run on various platforms.
- Compatibility with Existing Tools: Poetry can coexist with other Python tools like pip, virtualenv, and setuptools, allowing you to integrate it into your existing workflow.
- Simplified Dependency Upgrades: Poetry makes it easy to upgrade dependencies to their latest versions, helping you stay up-to-date with security patches and bug fixes.


erikbosch:
You get some CI errors on obsolete files (pre-commit/check-headers) - those errors you can ignore as you only move the files

pyproject.toml
```diff
@@ -1,7 +1,42 @@
...
+types-setuptools = "*"
+types-PyYAML = "*"
+flake8 = "^6.0.0"
+pre-commit = "*"
 
-[tool.setuptools-git-versioning]
-# Starting version is just a dummy value that never shall be used in CI
-starting_version = "0.0.0"
+[tool.poetry.scripts]
+vspec2binary = "vss_tools.vspec2binary:main"
```
sschleemilch, line 29: We could also add `....py` entries here if we want to keep "API" backwards compatibility

pyproject.toml
```diff
@@ -1,7 +1,42 @@
-[build-system]
-requires = ["setuptools", "setuptools-git-versioning", "wheel"]
-build-backend = "setuptools.build_meta"
+[tool.poetry]
+name = "vss-tools"
+homepage="https://github.com/COVESA/vss-tools"
+description='COVESA Vehicle Signal Specification tooling.'
+authors = ["Erik Jaegervall <erik.Jaegervall@bosch.de>"]
```
sschleemilch, line 5: Not sure who we should add here. Is there an official covesa email address?

tests/generators/test_generator.py
```diff
@@ -6,25 +6,23 @@
...
-    result = os.system(test_str)
-    assert os.WIFEXITED(result)
-    assert os.WEXITSTATUS(result) == 0
-
-    test_str = 'grep \"I found 2 comments with vss contributor\" out.txt > /dev/null'
-    result = os.system(test_str)
-    assert os.WIFEXITED(result)
-    assert os.WEXITSTATUS(result) == 0
-    os.system("rm -f  out.txt")
+def test_generator():
```
sschleemilch, line 14: That refactoring is a  bit out of scope, I could have just set the COLUMNS env var.
However, I plan to refactor all tests in that fashion anyway :haircut:

pyproject.toml
sschleemilch: Changed it to something generic

pyproject.toml
```diff
@@ -1,7 +1,42 @@
-[build-system]
-requires = ["setuptools", "setuptools-git-versioning", "wheel"]
-build-backend = "setuptools.build_meta"
+[tool.poetry]
+name = "vss-tools"
+homepage="https://github.com/COVESA/vss-tools"
+description='COVESA Vehicle Signal Specification tooling.'
+authors = ["COVESA VSS <covesa-dev+subscribe@covesa.global>"]
+version = "4.2.0"
```
erikbosch, line 6: Does poetry support automatic versioning based on git tags. We quite recently changed vss-tools so it tries to take version from git tag if present and also including it in pypi builds. Here we go back to hard-coded version, right?

tests/generators/test_generator.py
```diff
@@ -6,25 +6,23 @@
...
+        "python",
+        "example_generator.py",
+        "-k",
+        "VSS Contributor",
+        "-u",
+        "../vspec/test_units.yaml",
+        "test.vspec",
+    ]
+    env = os.environ.copy()
+    env["COLUMNS"] = "120"
```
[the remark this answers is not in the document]
sschleemilch: It is passed as an environment variable into the subprocess. It is needed in order to correctly grab for the test string. If COLUMNS is not wide enough (default is 80), the output could be torn apart between two lines

pyproject.toml
sschleemilch: No idea but I would go the other way around to be honest.
I had quite some good experience with [bump2version](https://pypi.org/project/bump2version/).

With that you can specify the version only once in the root of the project and define other files where it is used to be bumped automatically. So if we have the version only present in the pyproject.toml the .bumpversion.cfg would look like this:

```
[bumpversion]
current_version = 4.2.0

[bumpversion:file:pyproject.toml]
```

We can then use `bump2version --allow-dirty minor` to bump the version to `4.3.0`.

In our master workflow we can then just call our tool with `vspec2json --version` parse the output and tag the commit accordingly.

pyproject.toml
```diff
@@ -1,7 +1,42 @@
-[build-system]
-requires = ["setuptools", "setuptools-git-versioning", "wheel"]
-build-backend = "setuptools.build_meta"
+[tool.poetry]
+name = "vss-tools"
+homepage="https://github.com/COVESA/vss-tools"
+description='COVESA Vehicle Signal Specification tooling.'
+authors = ["COVESA VSS <covesa-dev+subscribe@covesa.global>"]
```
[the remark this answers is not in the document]
sschleemilch: Done

pyproject.toml
sschleemilch: Or with .bumpversion.cfg in place we could directly mention the tag in the workflow file aswell to bump it

erikbosch: I did some googling and there seems to be some attempts to use tag as source like https://github.com/tiangolo/poetry-version-plugin, but it seems to expect that last commit is tagged.

I also saw [this article](https://nadeauinnovations.com/post/2020/08/one-version-to-rule-them-all-keeping-your-python-package-version-number-in-sync-with-git-and-poetry/) where they give an example on how to update version based on git tag before building.

```
poetry version $(git describe --tags --abbrev=0)
poetry build
```

To give some background, in the past we relied on hard-coded versions. After we released a version (lets say `5.4`) we changed version on main-branch to something like `5.5.dev0` so that if anyone built/run from main it would be obvious that they did not use the official `5.4` version. With the current "automated" version based on tags it would instead report something like `5.4.post2.dirty-3478349834` . I see no real problem in going back to having a version hard-coded in `pyproject.toml`.

erikbosch: And as of today for vss-tools we have only version in one place, so updating it manually with "poetry version X.Y.Z" would work fine. After this PR is merged when preparing next release we would anyhow need to update the (informal) release instructions at https://github.com/COVESA/vehicle_signal_specification/wiki/Release-Instructions-and-Checklist

erikbosch (approved):
Looks good to me. Propose merge decision at meeting unless someone wants more time

PR #370, opened 2024-07-03
Use None as init value for node min/max

### LINKED ISSUE #354: VSSNode variable initializations
Hi everyone,

I ran into an issue with the way variables are initialized in the VSSNode class. In this case variables `min` and `max` are the problem. For vspec2id we are creating a 4-byte hash using multiple attributes (name/path, unit, type, datatype, enum values/allowed, or minimum/maximum). This is the way `min` and `max` (and other) are initialized (line 79/80). 

https://github.com/COVESA/vss-tools/blob/109b3edb5b2a2a0b0abec0852a23996da274e67a/vspec/model/vsstree.py#L35-L92

Now if the minimum value for e.g. a percentage is 0 and the `min` value is implicitly tansformed to an integer (which means it was set from a vehicle specification) and you want to export to a file using one of the exporters we mostly do something like `if node.min: ...`. 

https://github.com/COVESA/vss-tools/blob/109b3edb5b2a2a0b0abec0852a23996da274e67a/vspec/vssexporters/vss2id.py#L97-L100

These if-clauses will evaluate to False for an integer with value 0 which causes the 4-byte hash to be different. So in other words if we iterate on a static UID file (we impemented a validation to iterate static UID files), the `min` value would not be exported to file causing a different ID in the next iteration of the file.

What I suggest is we change all initializations of variables that are not strings e.g. `min` and `max` to use optional None so for python >= 3.10:

```
min: int | float | None = None
max: int | float | None = None
```

By doing so we know if the value has been set from a vehicle specification / overlay and if not it will stay None and therefore will not be exported. So on the export we would do sth like `if node.min is not None: ...`

I'm creating this issue because I'm wondering if any other exporters work the same way which could cause missing `min` and `max` values (or other?) on iterated vehicle specification files

sschleemilch:
Fixes #354 

We can see that the franca exporter was buggy I think.

I also think more values could be refactored from initializing them with `""` and go with `None` instead.

### DIFF
--- a/src/vss_tools/vspec/model/vsstree.py
+++ b/src/vss_tools/vspec/model/vsstree.py
@@ -79,8 +79,9 @@ class VSSNode(Node):
 
     unit: Optional[VSSUnit] = None
 
-    min = ""
-    max = ""
+    min: int | float | None = None
+    max: int | float | None = None
+
     allowed = ""
     instantiate = True
 
--- a/src/vss_tools/vspec/utils/idgen_utils.py
+++ b/src/vss_tools/vspec/utils/idgen_utils.py
@@ -13,8 +13,8 @@ def get_node_identifier_bytes(
     node_type: str,
     unit: str,
     allowed: str,
-    minimum: str,
-    maximum: str,
+    minimum: int | float | None,
+    maximum: int | float | None,
     strict_mode: bool,
 ) -> bytes:
     """Get a node identifier as bytes. Used as an input for hashing
@@ -36,8 +36,8 @@ def get_node_identifier_bytes(
         f"datatype: {data_type}, "
         f"type: {node_type}"
         f"allowed: {allowed}"
-        f"min: {minimum}"
-        f"max: {maximum}"
+        f"min: {minimum if minimum is not None else ''}"
+        f"max: {maximum if maximum is not None else ''}"
     ).encode()
 
     if strict_mode:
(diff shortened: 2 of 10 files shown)


erikbosch (approved):
Looks good to me

@nwesem - as you reported original issue please feel free to check if the solution here is OK for you.

I will not make it to next week VSS meeting. If anyone wants more time to review please write a note, otherwise we can merge after the meeting as I see it.

src/vss_tools/vspec/model/vsstree.py
```diff
@@ -79,8 +79,8 @@ class VSSNode(Node):
 
     unit: Optional[VSSUnit] = None
 
-    min = ""
-    max = ""
+    min = None
+    max = None
```
nwesem, line 83: I would suggest to be explicit here
```
min: int | float | None = None
max: int | float | None = None
```

src/vss_tools/vspec/utils/idgen_utils.py
```diff
@@ -13,8 +14,8 @@ def get_node_identifier_bytes(
     node_type: str,
     unit: str,
     allowed: str,
-    minimum: str,
-    maximum: str,
+    minimum: Optional[str],
+    maximum: Optional[str],
```
nwesem, line 18: same here it should probably be `minimum: int | float | None` and same for maximum, no?

src/vss_tools/vspec/model/vsstree.py
sschleemilch: Done

src/vss_tools/vspec/utils/idgen_utils.py
sschleemilch: Done

nwesem:
thx for taking care of this, other than that everything looks fine, ready to merge in my opinion :heavy_check_mark:

PR #371, opened 2024-07-04
Added logging args for level and file

sschleemilch:
#  About

Improved logging capabilities with:
- Added argument for controlling the log levels via `--log-level` with "INFO" as default
- Added argument for writing logs additionally into a log file with `--log-file`
- Using a `log` instance

Unrelated:
- Cleaned up docs to be up to date with changes regarding poetry
- Added code syntax highlighting for all code blocks in docs
- Using [alerts](https://docs.github.com/en/get-started/writing-on-github/getting-started-with-writing-and-formatting-on-github/basic-writing-and-formatting-syntax#alerts)

PR #375, opened 2024-07-05
Feature/cli redesign

### LINKED ISSUE #372: CLI Redesign
Hi *,

I just wanted to know how open the community is to a cli redesign.
Since pictures say more than a lot of words, here are the impressions of what I am thinking of:

image

image

It would use [click](https://click.palletsprojects.com/en/8.1.x/) that also has some nice features e.g. being able to set args via env variables (`VSS_TOOLS_LOG_LEVEL=DEBUG vspec2x .....`)

Before actually doing the functional change and hooking the logic into the cli I wanted to get feedback first.
Of course the design could also look different like not having sub-commands but specifying the output format in something like  `--format json`. The sub-command approach could have the benefit that it might be easier to model output specific options

sschleemilch:
#About (WIP - builds on top of #371) 

Fixes #372 

Complete CLI redesign.


erikbosch:
Some comments/background information to current design. Some time ago we had something similar, see `--format` in for example https://github.com/COVESA/vss-tools/blob/v3.1/docs/vspec2x.md and https://github.com/COVESA/vss-tools/blob/v3.1/vspec2x.py

Some reasons why we changed the setup was:

- We wanted to make it easy to customize tooling to use your own generator (only), like in https://github.com/COVESA/vss-tools/blob/master/tests/generators/example_generator.py, without getting any python requirements (like graphlql) from other tools. 
- We wanted for our Python builds (to upload to pypi) to exclude the "binary-tool", as it is difficult for us to build the c++ library for other platforms (building locally is fine), so we opted for not including the binary tool in the pypi package (as you then anyway need to download vss-tools and build the lib manually)
- Making it easy to remove tools from your own fork, like if you do not want franca (by any reason), just remove the `*2franca.py` files. Could be useful if you want to use vss-tools internally and you for license scanning reasons, vulnerability management or any other reason want to get rid of everything you do not need.

With this PR we go back a bit to how it was before, i.e. `vspec2x.py` explicitly containing/listing the tools. and it also becomes quite long. I do not know if it would be possible in some way to get a more dynamic design, where each method like `def yaml_(` is put in a separate file and not explicitly mentioned in vspec2x.py` to make it easy to add/remove tools.

sschleemilch:
Good Point, I will investigate a bit whether ITS possible. If not we could keep different entry points like before

sschleemilch:
Good Point, I will investigate a bit whether ITS possible. If not we could keep different entry points like before.
Click seems to support lazy subcommand loading: https://click.palletsprojects.com/en/8.1.x/complex/#lazily-loading-subcommands

Will try it out..

sschleemilch:
Changed it back to previous entry points and left the original positional options like the vspec file and the output file.
Changed a few arguments though to be more consistent.

sschleemilch:
image

adobekan:
MoM: follow up in a week.

erikbosch:
Conflicts

nwesem:
I really like this idea and I think even though it might break the current API, we should consider this for the next release (which would have to be 5.0? as it breaks a few things in terms of backwards compatibility, e.g. proplery renaming of the CLAs :grin:)

sschleemilch:
> I really like this idea and I think even though it might break the current API, we should consider this for the next release (which would have to be 5.0? as it breaks a few things in terms of backwards compatibility, e.g. proplery renaming of the CLAs 😁)
> 
> Edit: Also I would definitely vote for the lazy loading of subcommands

It is now here: #378

PR #377, opened 2024-07-10
Fix Table layout

erikbosch:
By some reason this table has been garbled in a commit some time ago. (Aggressive pre-commit?) This should make the table work again.


erikbosch:
@nwesem - could you possibly take a look at this one, if it seems to be ok

nwesem:
Many thx for fixing it, I wonder if some markdown tool I used broke it, I will give it a look, this one is good to go

PR #373, opened 2024-07-04
update exporters to export static UIDs if available

nwesem:
ToDo

PR #379, opened 2024-07-15
Test that ddsidl is correct

erikbosch:
This PR is related to a discussion in #373.

Previous instructions seemed to be a mashup of installation from source and installation from pypi, see https://github.com/eclipse-cyclonedds/cyclonedds-python

### DIFF
--- a/docs/VSS2DDSIDL.md
+++ b/docs/VSS2DDSIDL.md
@@ -79,19 +79,13 @@ IDL files can be supplied as input to one of the DDS implementation (e.g: Cycloe
 
 
 ```bash
-git clone https://github.com/eclipse-cyclonedds/cyclonedds.git
-pip3 install cyclonedds conan
-cd cyclonedds
-mkdir build
-cd build
-conan install .. --build missing
-cmake ..
-cmake --build . --target install
+pip3 install cyclonedds
 ```
 
 **Usage of CycloneDDS to take IDL file as input and generate Python file with types**
 
 > idlc **-l py**  ./results/res.idl
+
 ### FastDDS
 
 **Installation of FastDDS**
--- a/poetry.lock
+++ b/poetry.lock
@@ -26,6 +26,20 @@ files = [
 ]
 
 [[package]]
+name = "click"
+version = "8.1.7"
+description = "Composable command line interface toolkit"
+optional = false
+python-versions = ">=3.7"
+files = [
+    {file = "click-8.1.7-py3-none-any.whl", hash = "sha256:ae74fb96c20a0277a1d615f1e4d73c8414f5a98db8b799a7931d1582f3390c28"},
+    {file = "click-8.1.7.tar.gz", hash = "sha256:ca9853ad459e787e2192211578cc907e7594e294c7ccc834310722b41b9ca6de"},
+]
+
+[package.dependencies]
+colorama = {version = "*", markers = "platform_system == \"Windows\""}
+
+[[package]]
 name = "colorama"
 version = "0.4.6"
 description = "Cross-platform colored terminal text."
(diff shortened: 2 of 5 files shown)


sschleemilch (approved):

nwesem:
tested it, looks good to me

docs/VSS2DDSIDL.md
```diff
@@ -79,19 +79,13 @@ IDL files can be supplied as input to one of the DDS implementation (e.g: Cycloe
...
 ```bash
-git clone https://github.com/eclipse-cyclonedds/cyclonedds.git
-pip3 install cyclonedds conan
-cd cyclonedds
-mkdir build
-cd build
-conan install .. --build missing
-cmake ..
-cmake --build . --target install
+pip3 install cyclonedds
```
nwesem, line 82: maybe the only thing we could improve is making this a dev dependency using `poetry add --group dev cyclonedds`?

nwesem:
Actually I'm wondering if the output of `idlc -l py ./test.idl` is supposed to be 
```
No default extensibility provided. For one or more of the aggregated types in the IDL the extensibility is not explicitly set. Currently the default extensibility for these types is 'final', but this may change to 'appendable' in a future release because that is the default in the DDS XTypes specification.
```

docs/VSS2DDSIDL.md
erikbosch: I have already added it to pyproject.toml (hopefully in the right way). So installing it manually is maybe not needed if you have built vss-tools with poetry? (But needed if you just download vss-tools from pypi, right?)

erikbosch:
> Actually I'm wondering if the output of `[poetry run] idlc -l py ./test.idl` is supposed to be
> 
> ```
> No default extensibility provided. For one or more of the aggregated types in the IDL the extensibility is not explicitly set. Currently the default extensibility for these types is 'final', but this may change to 'appendable' in a future release because that is the default in the DDS XTypes specification.
> ```

Based on what we currently generate I think the warning is correct. I do not know if it can be easily suppressed. We could off course add something like shown in the link below, but as no-one of us seem to use ddsidl we can likely leave it as is.  

https://fast-dds.docs.eprosima.com/en/latest/fastddsgen/dataTypes/dataTypes.html#extensibility

sschleemilch:
Unfortunately this broke `poetry install` for me. That might be my setup only though (MacOs with Python from Nix)

```bash
  - Installing cyclonedds (0.10.5): Failed

  ChefBuildError

  Backend subprocess exited when trying to invoke get_requires_for_build_wheel

  Could not locate cyclonedds. Try to set CYCLONEDDS_HOME or CMAKE_PREFIX_PATH


  at /nix/store/zp80lg9gv25mp11gh4cpwf1lg6y2jkfz-python3.11-poetry-1.8.3/lib/python3.11/site-packages/poetry/installation/chef.py:164 in _prepare
      160│
      161│                 error = ChefBuildError("\n\n".join(message_parts))
      162│
      163│             if error is not None:
    → 164│                 raise error from None
      165│
      166│             return path
      167│
      168│     def _prepare_sdist(self, archive: Path, destination: Path | None = None) -> Path:

Note: This error originates from the build backend, and is likely not a problem with poetry but with cyclonedds (0.10.5) not supporting PEP 517 builds. You can verify this by running 'pip wheel --no-cache-dir --use-pep517 "cyclonedds (==0.10.5)"'
```

erikbosch:
Possibly some arm64 issue? If necessary we could remove this test "out of pytest" so that you easily could run all other pytests even on a platform where it is is diffiult/impossible to install cyclonedds

nwesem:
I just remember now that I ran into the same issue until I built cyclonedds lib from source via https://github.com/eclipse-cyclonedds/cyclonedds
Did you have to do that too @erikbosch?

erikbosch:
Are you running on Mac/Arm64 @nwesem ? I am running on Linux/amd64 and not experiencing any problem. The same is true for CI I think.

Issue #380, opened 2024-07-16
Improve core logic of vspec parsing and the data model


sschleemilch:
# About

Idea is to refactor how we parse the vspecs and more importantly how we validate the node data.

The current implementation using `core_attributes` as well as a lot of classmethods to set a reference tree looks a bit complicated and error prune.

[Pydantic](https://docs.pydantic.dev/latest/) could help us model our nodes and their restrictions.

sschleemilch:
My implementation proceeded and looks quite solid already.

Things that it detects that we are currently not are:

### Cross checking of `allowed-datatypes`

Taking this example config:
```yaml
kWh:
  definition: Energy consumption measured in kilowatt hours
  unit: kilowatt hours
  quantity: work
  allowed-datatypes: ['numerics']
```
Is detected as a misconfiguration since `numerics` does not exist but passes the current vss-tools model:
```bash
ERROR    kWh: 1 validation error for VSSUnit                                                        units_quantities.py:20
                    allowed-datatypes
                      Assertion failed, Invalid datatype: numerics [type=assertion_error,
                    input_value=['numerics'], input_type=list]
                        For further information visit https://errors.pydantic.dev/2.8/v/assertion_erro
```

### Cross checking of `quantity` being defined

Pretty similar config but using a quantity that is not defined:
```yaml
kWh:
  definition: Energy consumption measured in kilowatt hours
  unit: kilowatt hours
  quantity: workie
  allowed-datatypes: ['numeric']
```
Is detected by tool but not by vss-tools:
```bash
ERROR    kWh: 1 validation error for VSSUnit                                                        units_quantities.py:20
                    quantity
                      Assertion failed, Invalid quantity: workie [type=assertion_error, input_value='workie',
                    input_type=str]
                        For further information visit https://errors.pydantic.dev/2.8/v/assertion_error
```

### Strict units yaml structure

Protecting against wrong structure in units.yaml and quantities.yaml such as using `description` instead of `definition`, `units` instead of `unit`
We actually have already violations in the current vss:
```diff
diff --git a/spec/units.yaml b/spec/units.yaml
index 5f32a026..ee0c9ab3 100644
--- a/spec/units.yaml
+++ b/spec/units.yaml
@@ -58,7 +58,7 @@ cm/s^2:
 # Volume

 ml:
-  description: Volume measured in milliliters
+  definition: Volume measured in milliliters
   unit: milliliter
   quantity: volume
   allowed-datatypes: ['numeric']
[...]

PR #382, opened 2024-07-19
Core Rework ⚠️ BREAKING CHANGES

sschleemilch:
Fixes: #380 
Fixes: #390
Fixes the issue in discussion of #360 (removed the test in test_overlay_on_instance)

Rewrote the core logic using Pydantic.
- Tree nodes contain a Pydantic class instance in their 'data' attribute
- Pydantic handles the model validation
- Model is initialized in 'relaxed' mode so that we can support
  overlaying of instance nodes without the need to specify nodes
  completely
- Added a lot more validations to the model regarding consistency of
  datatypes, units, allowed attributes etc
- Also added tree validation regarding what node types can have what types of parents (all nodes need to have branches as parents except properties). Enforcing it
- Rewrote include vspec logic and more
- Rewrote instance expansion logic and more
- Adapted exporters and improved slightly (e.g. csv uses the csv lib)
- Restricting `-e/--extended-attributes` to not be able to pass core attributes

> [!TIP]
> **Overlay nodes do not need any attributes anymore except the ones you want to overwrite**
> That is also valid for overwriting instances.
> Example: 
> ```yaml
> Vehicle:
>   type: branch
>   description: Vehicle
> 
> Vehicle.Door:
>   type: branch
>   description: Door
>   instances:
>   - Row[1,2]
> 
> Vehicle.Door.Row1:
>   description: Very special Door
> ```

> [!WARNING]
> This breaks the backwards compatibility to older vss versions and therefore the test has been marked to be skipped.
> Reason is that we handle unit and quantity files more strict and because of that, vss had to be patched (https://github.com/COVESA/vehicle_signal_specification/pull/758). Old release are decided not to be patched

# Points of interest / Questions / Code TODOs
- We are never exporting the `arraysize` attribute. On purpose? 🆗 Not on purpose but we keep not exporting it for now until needed
- vss2binary  -> export_node -> `validate` what is it? 🆗 Clarified. In future to be discussed whether it should be part of the model when  exporters explicitly mentioning it.
- Should type tree properties be compliant to name-style and should follow the same `--strict/--aborts name-style` behavior? 🆗 Type tree does not need to follow name style conventions. It would lead to weird property names

# Added
- Exporter to render a tree (`vspec export tree`)
image

# Todo

- [x] Tests for newly introduced validations
- [x] Inline docs
- [x] Add `has` as valid bool prefix

# ⚠️ Behavior changes to be discussed -> Change is ok 🆗 

## Extending instances implicitly

Take this config
```yaml
Vehicle.Door:
  type: branch
  description: Door
  instances:
  - Row[1,2]

Vehicle.Door.Row3:
  type: branch
  description: Got another door

Vehicle.Door.Id:
  type: branch
  description: Door ID
```
Although there are only `Row1` and `Row2` created the user wants to add another one implicitly.
However that is some weird magic since it is unlogical to distinguish the behavior from what is displayed in the config: Adding something to all instances. Just because the name of `Row3` matches the instance expansions we should not rely on that imo.

So, the outcome of the config as is would be in this implementation:
```
Vehicle
└── Door
    ├── Row1
    │   ├── Row3
    │   └── Id
    └── Row2
        ├── Row3
        └── Id
```
That is obviously not what the config intended but in my opinion the config is incorrect. When changing `instances` to `Row[1,3]` we get the correct result:
```
Vehicle
└── Door
    ├── Row1
    │   └── Id
    ├── Row2
    │   └── Id
    └── Row3
        └── Id
```

# Validation improvements

### Cross checking of `allowed-datatypes`

Taking this example config:
```yaml
kWh:
  definition: Energy consumption measured in kilowatt hours
  unit: kilowatt hours
  quantity: work
  allowed-datatypes: ['numerics']
```
Is detected as a misconfiguration since `numerics` does not exist but passes the current vss-tools model:
image

### Cross checking of `quantity` being defined

Pretty similar config but using a quantity that is not defined:
```yaml
kWh:
  definition: Energy consumption measured in kilowatt hours
  unit: kilowatt hours
  quantity: workie
  allowed-datatypes: ['numeric']
```
Is detected by tool but not by vss-tools:
image

### Strict units yaml structure

Protecting against wrong structure in units.yaml and quantities.yaml such as using `description` instead of `definition`, `units` instead of `unit`
We actually have already violations in the current vss:
```diff
diff --git a/spec/units.yaml b/spec/units.yaml
index 5f32a026..ee0c9ab3 100644
--- a/spec/units.yaml
+++ b/spec/units.yaml
@@ -58,7 +58,7 @@ cm/s^2:
 # Volume

 ml:
-  description: Volume measured in milliliters
+  definition: Volume measured in milliliters
   unit: milliliter
   quantity: volume
   allowed-datatypes: ['numeric']
@@ -138,7 +138,7 @@ g:
   allowed-datatypes: ['numeric']
 kg:
   definition: Mass measured in kilograms
-  label: kilogram
+  unit: kilogram
   quantity: mass
   allowed-datatypes: ['numeric']
 lbs:
@@ -205,7 +205,7 @@ weeks:
   allowed-datatypes: ['numeric']
 months:
   definition: Duration measured in months
-  units: months
+  unit: months
   quantity: duration
   allowed-datatypes: ['numeric']
 years:
```

### Cross checking `arraysize` to `datatype` being an array

Example config
```yaml
Vehicle:
    type: branch
    description: vehicle

Vehicle.Test:
    type: sensor
    datatype: uint8
    arraysize: 3
    description: foo
```

Is detected by the tool (note that datatype is not an array) but vss-tools has no problem with it:
image

Similar: Also checks that `arraysize` matches the entries of `default`:
image

### Checking `default` against `datatype`

Example:
```yaml
Vehicle:
    type: branch
    description: vehicle

Vehicle.Test:
    type: sensor
    datatype: uint8[]
    description: foo
    default: 3
```

Is detected but is not currently:
image

### Checking `allowed` entries matching `datatype`

Similar to the use case before, Config:
```yaml
Vehicle.Test:
    type: sensor
    datatype: uint8[]
    arraysize: 3
    description: foo
    allowed: [1,2,'foo']
```

Note that `foo` is not an uint8 and should not be accepted.
Is detected:
image

It also checks ranges e.g. -3 will violate uint8.

### Checking `min/max` set together with `allowed`

Config:
```yaml
Vehicle.Test:
    type: sensor
    datatype: uint8
    description: foo
    min: 3
    max: 5
    allowed: [2,3]
```
Result:
image

### Checking of `default` values against `allowed`

Config:
```yaml
Vehicle.Test:
    type: sensor
    datatype: uint8
    description: foo
    allowed: [1,2,3]
    default: 4
```
Result:
image

### Checking of `datatype` vs `allowed_values` in specified `unit`

Config:
```yaml
Vehicle.Speed:
    type: sensor
    datatype: string
    unit: kWh
```
Result:
image


erikbosch:
MoM:

- Feature complete, please review, but minor changes may still happen
- Remaining works is to fix unit test and address bugs found in testing

src/vss_tools/vspec/utils/misc.py
```diff
@@ -1,14 +1,17 @@
-# Copyright (c) 2023 Contributors to COVESA
```
erikbosch, line 1: Copyright removed?

src/vss_tools/vspec/vssexporters/vss2csv.py
```diff
@@ -36,53 +36,43 @@ def print_csv_header(file, uuid, entry_type: AnyStr, include_instance_column: bo
...
-# Format a data or header line according to the CSV standard (IETF RFC 4180)
-
-
-def format_csv_line(csv_fields):
-    formatted_csv_line = ""
-    for csv_field in csv_fields:
-        if csv_field is None:
-            csv_field = ""
-        formatted_csv_line = (
-            formatted_csv_line + '"' + str(csv_field).replace('"', '""') + '",'
```
erikbosch, line 55: I notice some changes in output of CSV exporters. This line was added once upon the time to handle at least some common scenarios, like if a description/comment contain a comma or a double quote, and then the string has to be quoted according to https://datatracker.ietf.org/doc/html/rfc4180

Now it seems we skip any form of escaping, or?

src/vss_tools/vspec/utils/misc.py
sschleemilch: Not on purpose. Will readd it

src/vss_tools/vspec/vssexporters/vss2csv.py
sschleemilch: I think the csv writer library handles escaping for us. And yes, the output of all csv's changed and removed unnecessary quoting

sschleemilch: But I will check some corner cases I guess

erikbosch: Would be quite easy to add some test cases unless we already have some.

sschleemilch: It works fine. I honestly see no real reason to test the official csv lib

docs/vspec.md
```diff
@@ -22,6 +22,7 @@ vspec export --help
 │ jsonschema                        Export as a jsonschema.                                              │
 │ protobuf                          Export as protobuf.                                                  │
 │ yaml                              Export as YAML.                                                      │
+│ tree                              Export as Tree.                                                      │
```
erikbosch, line 25: Do we possibly need to say somewhere something about the "Tree export format", like saying that it is a text file and giving  a simple example. Like creating a `tree.md` file and reference it from somewhere in this file.

src/vss_tools/vspec/units_quantities.py
```diff
@@ -0,0 +1,54 @@
...
+    files: list[Path], class_type: type[VSSUnit | VSSQuantity]
+) -> dict[str, VSSUnit | VSSQuantity]:
+    data = {}
+    for file in files:
+        content = yaml.safe_load(file.read_text())
+        if not content:
+            log.warning(f"{file}, empty")
+            continue
+        log.info(
+            f"Loaded '{class_type.__name__}', file={file.absolute()}, elements={len(content)}"
```
erikbosch, line 33: Should this possibly be on debug level instead, gives quite a lot of output for standard vss catalog

(based on our approach so far that you on Info-level should not get too many lines of output)

src/vss_tools/vspec/tree.py
```diff
@@ -0,0 +1,541 @@
...
+            connected = root.connect(fqn, orphan)
+            if connected:
+                connected_fqns.append(fqn)
+        for fqn in connected_fqns:
+            del orphans[fqn]
+
+    if orphans:
+        log.warning(f"Orphans: {len(orphans)}")
+
+    log.info(f"Tree, root='{root.name}', size={root.size}, height={root.height}")
```
erikbosch, line 515: debug level instead?

src/vss_tools/vspec/tree.py
```diff
@@ -0,0 +1,541 @@
...
+                # Appending all original children at the right place
+                # Possibly overwriting content if wished
+                add_expanded_instance_children(
+                    generated_instance_leaf_nodes, instance_node, instance_node_copy
+                )
+
+                instance_node.data.instances = []  # type: ignore
+            instance_nodes = self.get_instance_nodes()
+        if iterations:
+            log.info(f"Instances, iterations={iterations}, nodes={n_instance_nodes}")
```
erikbosch, line 255: debug level?

src/vss_tools/vspec/tree.py
```diff
@@ -0,0 +1,541 @@
...
+        self.data = get_vss_raw(data, fqn)
+        self.uuid: str | None = None
+
+    def _post_attach(self, parent: VSSNode):
+        """
+        Updating the data fqn when getting reattached.
+        We need the fqn in the data for validation purposes.
+        """
+        log.debug(
+            f"Got attached to parent='{parent.get_fqn()}', new fqn='{self.get_fqn()}'"
```
erikbosch, line 75: I do not know how consistent we have been in the past, but at least sometimes we have tried to avoid using f-strings in at least debug/info logging, with the rationale that it may result in extra processing to generate the f-string even if not printed in the end. I.e. better to use `%s`. See link below.

https://pylint.readthedocs.io/en/stable/user_guide/messages/warning/logging-fstring-interpolation.html

When testing with whole VSS catalog I get a gut feeling that generating for example JSON takes quite much longer time compared to before. I have not done any thorough tests. If you have time feel free to compare the result with this PR and previous to see if there are any significant changes.

erikbosch: On the other hand, just using `%s` will not help here as maybe the `get_fqn`call is the costly part. If we verify that it is costly then maybe we need a work-around like. (But I do not really know if it actually is a problem from performance perspective)

```
if logger.isEnabledFor(logging.DEBUG):
    logger.debug(f"var1 {var1}, var2 {var2}") 
```

docs/vspec.md
sschleemilch: Done

tests/vspec/test_overlay_on_instance/test_overlay_on_instance.py
```diff
@@ -20,25 +21,8 @@ def test_expanded_overlay(tmp_path):
...
-
-def test_expanded_overlay_no_type_datatype(tmp_path):
-    """
-    This test shows current limitation on type/datatype lookup.
-    We cannot do lookup on expanded names containing instances.
-    That would require quite some more advanced lookup considering instance declarations.
-    Not impossible, but more complex.
-    """
-    spec = HERE / "test.vspec"
-    overlay = HERE / "overlay_no_type_datatype.vspec"
```
erikbosch, line 39: Is this file needed now when you have removed the test case?

src/vss_tools/vspec/units_quantities.py
sschleemilch: Hm, for units and quantities I would like to show it in info, for vspec e.g. in the VSS, it is a lot of output, that's true. Will change that to maybe only show the toplevel vspec in the info and all subsequent loads in the debug

src/vss_tools/vspec/tree.py
sschleemilch: ok

src/vss_tools/vspec/tree.py
sschleemilch: ok

src/vss_tools/vspec/tree.py
sschleemilch: I am aware of the f string trade off for logging but usually still use them in favor for readability.
I will make some performance measurements.
Regarding `get_fqn()` I am thinking right now to make it a member and update it when the node gets attached / re-attached

tests/vspec/test_overlay_on_instance/test_overlay_on_instance.py
sschleemilch: The file still contains another test case.

tests/vspec/test_overlay_on_instance/overlay_2.vspec
```diff
@@ -50,22 +50,6 @@ A.B.Row3.Right.C:
   datatype: int8
   my_id: Lund
 
-# Adding one outside instance range, shall not inherit original comment
```
erikbosch, line 53: Should be OK to change behavior when using `Row5` like below, I could anyway not find that we have actually specified in documentation what shall happen. But I am thinking if we should remove the Row5 example, or rather keep it with a note saying that we do not do any "smart" name analysis and instead update expected.json

But I do not really have a strong opinion!

.github/workflows/check-header.yml
```diff
@@ -21,7 +21,7 @@ jobs:
 
     - name: Get changed files
       run: |
-        echo "files=$(git diff --name-only ${{ github.event.pull_request.base.sha }}..${{ github.event.pull_request.head.sha }} | grep -v '^obsolete' | tr '\n' ',')" >> $GITHUB_ENV
+        echo "files=$(git diff --name-only ${{ github.event.pull_request.base.sha }}..${{ github.event.pull_request.head.sha }} | grep -v '^obsolete' | grep -v '^tests' | tr '\n' ',')" >> $GITHUB_ENV
```
erikbosch, line 24: Any particular reason to exclude files in the test directory?

( I have no strong opinion here)

sschleemilch: Hm, there was a fail for the copyright header I did not quite get at first sight why that happens. Will revert and see what it was

src/vss_tools/vspec/tree.py
sschleemilch: Indeed performance is not good. Will investigate. Probably unnecessary deepcopys somewhere

tests/vspec/test_overlay_on_instance/overlay_2.vspec
sschleemilch: I will revert it and comment it

src/vss_tools/vspec/tree.py
sschleemilch: Fixed the performance. One performance finding is still there that is not a quick fix though and might need some more work

.github/workflows/check-header.yml
sschleemilch: Yeah so if you check the most recent check run, it tries to validate that test vspecs do have copyright headers.
Do we want that? If yes, I need to add headers  to all  test files that I have touched

erikbosch:
MoM:

#Questions

## We are never exporting the arraysize attribute. On purpose?

S Schildt: Not on purpose

## vss2binary -> export_node -> validate what is it?

@UlfBj to be the one to answer.

* Ulf: Checked by VISS server. Says what access control you have.
* Niclas: Written/Read as a string
* S Schildt: Will need to provide it
* Erik: @UlfBj can you check if the field `validate` is described somewhere? Extend https://github.com/COVESA/vss-tools/blob/master/binary/README.md with an example or formal syntax stating that it is a string.

## Should type tree properties be compliant to name-style and should follow the same --strict/--aborts name-style behavior?

* E: Does anyone used struct concept
* Adnan will check with S Schleemilch what it should be


* Please review
* Possible merge decision next week

sschleemilch:
> MoM:
> 
> #Questions
> 
> ## We are never exporting the arraysize attribute. On purpose?
> S Schildt: Not on purpose
> 
> ## vss2binary -> export_node -> validate what is it?
> @UlfBj to be the one to answer.
> 
> * Ulf: Checked by VISS server. Says what access control you have.
> * Niclas: Written/Read as a string
> * S Schildt: Will need to provide it
> * Erik: @UlfBj can you check if the field `validate` is described somewhere? Extend https://github.com/COVESA/vss-tools/blob/master/binary/README.md with an example or formal syntax stating that it is a string.
> 
> ## Should type tree properties be compliant to name-style and should follow the same --strict/--aborts name-style behavior?
> * E: Does anyone used struct concept
> * Adnan will check with S Schleemilch what it should be
> * Please review
> * Possible merge decision next week

- vss2binary validate feature is fixed
- not changing export behavior for now regarding arraysize
- not validating names for the types tree

From my perspective, it is done

UlfBj:
>> Erik: @UlfBj can you check if the field validate is described somewhere? Extend https://github.com/COVESA/vss-tools/blob/master/binary/README.md with an example or formal syntax stating that it is a string.

This information is already linked to:
More information can be found in: [VISS Access Control](https://www.w3.org/TR/viss2-core/#access-control-selection).

sschleemilch:
> > > Erik: @UlfBj can you check if the field validate is described somewhere? Extend https://github.com/COVESA/vss-tools/blob/master/binary/README.md with an example or formal syntax stating that it is a string.
> 
> This information is already linked to: More information can be found in: [VISS Access Control](https://www.w3.org/TR/viss2-core/#access-control-selection).

I am wondering why it is not a specified internal attribute if it is present in the code and explicitly mentioned?

SebastianSchildt:
> > > > Erik: @UlfBj can you check if the field validate is described somewhere? Extend https://github.com/COVESA/vss-tools/blob/master/binary/README.md with an example or formal syntax stating that it is a string.
> > 
> > 
> > This information is already linked to: More information can be found in: [VISS Access Control](https://www.w3.org/TR/viss2-core/#access-control-selection).
> 
> I am wondering why it is not a specified internal attribute if it is present in the code and explicitly mentioned?

Because it is "only" used in VISS (and thus the binary exporter, which currently is also only for the VISSR reference implementation), and not considered part of VSS (i.e. expected to be supported by all VSS tooling). That does not mean it is "not good", it is just similar to e.g. DBC mapping keys (currently afaik only used in KUKSA) or some specific deployment keys BMW might use in internal tooling.

I think, what would be a nice addition for the future, in case a specific tooling "requires" or understands special keys (like the binary exporter), if it could programmatically via an API whitelist those, instead of forcing the user to whitelist them on the command line.

SebastianSchildt:
Meeting 20th: Looking good,

What about the breaking changes? 
 - backport unit changes to old VSS releases?
     - Seb: maybe not
 - old tooling is still working on master
 - just warn?
  - will just postpone the problem

MoM: Merge

PR #384, opened 2024-07-22
Allow longer line in unit file tests

erikbosch:
Needed due to regression when running tests from vss repo


sschleemilch:
In general I think we should switch those tests and make use of `--log-file` and reading that file instead. Then we do not need to rely on terminal width...

PR #389, opened 2024-07-25
Remove cyclonedds as vss-tools dependency

erikbosch:
I recently added a pytest testcase running `ddsidl` to check that output from ddsidl generator is syntactically correct. That worked well in my amd64 environment and in CI which also uses amd64, but others have reported problems installing `cyclonedds` for example on `aarch64` architecture (Mac).

I still think it is good to have a sanity check that generated ddsidl is correct. This PR moves the check to CI only and removes it from Python dependencies. It could be discussed if this is the best approach, but at least it removes the need for cyclonedds for just building vss-tools package.


SebastianSchildt (approved):
I can confirm poetry works on Mac again without cyclone exploding. Also, changes look reasonable

sschleemilch (approved):

PR #392, opened 2024-08-07
Updating to 5.0a0

erikbosch:
Intended for creating a pypi alpha release for testing purposes, if someone wants to use the new `vspec` CLI without building from source.

I have locally created a PyPI release 5.0a1 based on this one and uploaded to pypi. Install by:

`pip install --pre vss-tools`

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -3,7 +3,7 @@ name = "vss-tools"
 homepage="https://github.com/COVESA/vss-tools"
 description='COVESA Vehicle Signal Specification tooling.'
 authors = ["COVESA VSS <covesa-dev@covesa.global>"]
-version = "4.2.0"
+version = "5.0a0"
 license = "Mozilla Public License 2.0"
 readme = ["README.md", "README-PYPI.md"]
 


erikbosch:
@sschleemilch - as Stefan in the meeting yesterday asked for some guidance on how to create a VSS PR I thought it could be good to have a pre-release of vss-tools on pypi for easy installation and testing. We do not have a 100% well defined process of how to name pre-released/dev-releases but we have some guidance at https://github.com/COVESA/vehicle_signal_specification/wiki/Release-Instructions-and-Checklist#decide-on-version-number-to-use

Anyhow, as you are currently very active on vss-tools I would like to ask you if you have any options using `5.0a0` on master branch (for now), or if you rather prefer `5.0.0.dev1` or something similar. I see a value anyway in getting rid of the `4.2` as it might be confusing when you do a local install.

sschleemilch:
Recently I have been using `rc` so I would vote for something like `5.0.0rc1`. But I don't care so much how we name it

erikbosch:
Previously we have used `rc` only when it actually is a release candidate, like when we think we are complete for the next release. using that approach this shall not be a release candidate as we are awaiting more functionality (like your PR). But as you say it does not matter that much - in pypi it will anyway be treated as pre-release

sschleemilch:
you are right, I also see the rc only viable when the target version is feature complete...

PR #393, opened 2024-08-12
😎 🚀 Modernized pre-commit config, 👋 ruff, ➖ flake

sschleemilch:
# About

Using `ruff` instead of `flake8` since ruff is faster and can do everything flake can do.

Default ruff rules are: 
```toml
select = ["E4", "E7", "E9", "F"]
```

Decided to use all "E" and add "I" for sorting imports.
`F` are pyflake rules: https://docs.astral.sh/ruff/rules/#pyflakes-f
`E` are pycodestyle rules: https://docs.astral.sh/ruff/rules/#error-e
`I` is sorting imports like `isort`: https://docs.astral.sh/ruff/rules/#isort-i

Also changed line length to 120 which seems reasonable for me in current screen size times 🖥️ 

## Added pre-commit checks and their meaning

- `check-ast`: Checks whether a python file is valid python. Is a low hanging fruit in my opinion. Will it trigger much? Probably not since every file should be in some unit test.
- `check-builtin-literals`: Checks whether initializing empty types: e.g. use `x = ()` instead of `x = tuple()`
- `check-case-conflicts`: Check for files with names that would conflict on a case-insensitive filesystem. I think you always want that one... You will debug a lot when you have such a case
- `check-docstring-first`: Checks that a file docstring comes at the beginning of the file only
- `check-merge-conflict`: Just checks files whether they contain git merge conflict markers
- `check-toml`: Checks whether tomls are valid tomls (we have pyproject.toml and poetry.toml)
- `mixed-line-ending`: Should be caught by git config but it depends on user config. This one checks that crlf and newline are not mixed
- `check-vcs-permalinks`: Checks that links to github content are permalinks
- `debug-statements`: Check for debugger imports and py37+ breakpoint() calls in python source.


SebastianSchildt:
Meeting 08/27: We need "opinions"

SebastianSchildt (approved):
Let's try this

PR #398, opened 2024-08-26
🐞 Exporting null values of given extended attributes

sschleemilch:
# About

Extended user attributes given with `-e` should always be exported, even if they contain `null` values. This was the case for the last release.
The recent core refactoring changed the behavior and filters all null values. Therefore this change here that handles extended attributes in a special way and not includes them.
This only concerns the `as_dict` method and methods using that (such as `as_flat_dict`). Exporters not using those methods have not been changed and might need to be updated aswell if users report problems.

Example vspec:
```yaml
Vehicle:
  type: branch
  description: Vehicle
  foo: ~
  bar: ~
```
Should result in the export (only when passing `-e foo -e bar`):
```yaml
Vehicle:
  bar: null
  description: Vehicle
  foo: null
  type: branch
```

### DIFF
--- a/src/vss_tools/vspec/model.py
+++ b/src/vss_tools/vspec/model.py
@@ -85,13 +85,21 @@ class VSSRaw(BaseModel):
         self,
         with_extra_attributes: bool = True,
         exclude_fields: list[str] = EXPORT_EXCLUDE_ATTRIBUTES,
+        extended_attributes: tuple[str, ...] = (),
     ) -> dict[str, Any]:
         excludes = exclude_fields.copy()
         if not with_extra_attributes:
-            excludes.extend(self.get_extra_attributes())
-        data = {
-            k: v for k, v in self.model_dump(mode="json", exclude_none=True, exclude=set(excludes)).items() if v != []
-        }
+            for extra_attribute in self.get_extra_attributes():
+                if extra_attribute not in extended_attributes:
+                    excludes.append(extra_attribute)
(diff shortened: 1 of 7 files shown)


sschleemilch:
> Bonusgoal: Add a test @sschleemilch

Done

src/vss_tools/vspec/model.py
```diff
@@ -85,13 +85,21 @@ def as_dict(
...
-        data = {
-            k: v for k, v in self.model_dump(mode="json", exclude_none=True, exclude=set(excludes)).items() if v != []
-        }
+            for extra_attribute in self.get_extra_attributes():
+                if extra_attribute not in extended_attributes:
+                    excludes.append(extra_attribute)
+        data = {}
+        for k, v in self.model_dump(mode="json", exclude_none=False, exclude=set(excludes)).items():
+            if k not in extended_attributes:
+                if v == [] and k:
```
erikbosch, line 98: Out of curiosity (or because I am too stupid after vacation) - why do we need the `and k` clause here but not on line 100? I.e. in which cases does `k` evaluate to false?

sschleemilch: I also don't see why we need `and k`. Good catch 👍

erikbosch:
MoM:

- OK to be merged when Erik has reviewed/approved

erikbosch (approved):
LGTM

PR #405, opened 2024-09-16
✨ New exporter: go

sschleemilch:
Added new generator: go

It does produce structs to work with the vss tree.

It only produces simple structs for now. In future, we can also generate get/set methods and use the `allowed` as well as `min/max` fields.

### DIFF
--- a/docs/go.md
+++ b/docs/go.md
@@ -1,0 +1,66 @@
+# Go lang struct exporter
+
+This exporter produces type struct definitions for the [go](https://go.dev/) programming language.
+
+## Exporter specific arguments
+
+### `--package`
+
+The name of the package the generated sources (output and types output) will have.
+
+# Example
+
+Input model:
+```yaml
+# model.vspec
+Vehicle:
+  type: branch
+  description: Vehicle
+Vehicle.Speed:
+  type: sensor
+  description: Speed
+  datatype: uint16
+Vehicle.Location:
+  type: sensor
+  description: Location
+  datatype: Types.GPSLocation
+```
+
+Input type definitions:
+```yaml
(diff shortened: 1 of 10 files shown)


erikbosch:
Could you possibly add a README in https://github.com/COVESA/vss-tools/tree/master/docs that gives a bit on background and how the tool can be used or is intended to be used. Like linking to https://go.dev/ as possible not every one is familiar with Go.

We use Go in https://github.com/COVESA/vissr and I am curious if what we have here could be a replacement to the "binary" format currently used by that tool. ideas @UlfBj ?

sschleemilch:
> Could you possibly add a README in https://github.com/COVESA/vss-tools/tree/master/docs that gives a bit on background and how the tool can be used or is intended to be used. Like linking to https://go.dev/ as possible not every one is familiar with Go.
> 
> We use Go in https://github.com/COVESA/vissr and I am curious if what we have here could be a replacement to the "binary" format currently used by that tool. ideas @UlfBj ?

Done. Added README: https://github.com/sschleemilch/vss-tools/blob/feature/go-exporter/docs/go.md

sschleemilch:
Lets not merge it yet. I would like to make some improvements.

erikbosch:
MoM:

- Erik: Not planned for 5.0
- Ulf: Did a quick review, this one generates structs, slightly different scope compared to binary format, which is language agnostic, so not a replacement for binary 
- Erik: Good if @UlfBj checks that code follows "good Go practice"
- Ulf: I can do it
- Please review

UlfBj:
It looks good to me.

erikbosch:
MoM: Ok to merge

PR #406, opened 2024-09-16
Updating CHANGELOG and minor chnage in UUID handling

erikbosch:
I took a look at the changelog, and this PR contains some changes related to that

- Rearranged order to start with newest release. 
- Added some lines on changes that will be part of 5.0, best reviewed in https://github.com/boschglobal/vss-tools/blob/erik_uuid/CHANGELOG.md
- Reviewing the uuid support. As part of the CLI refactoring `-no--uuid` was re introduced again (I think the new CLI framework expects boolean parameters to be a "pair"). But I assume still we want to remove it completely for VSS 6.0 so I added a deprecation note if `--uuid` is used

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,17 +7,110 @@ It includes changes that are included in released version, but also changes plan
 *This document only contain changes introduced in VSS-Tools 3.0 or later!*
 
 
+## Planned changes for VSS-Tools 6.0
 
-## VSS-Tools 3.0
+### Change in UUID handling.
 
-[Complete release notes including VSS-Tools changes](https://github.com/COVESA/vehicle_signal_specification/releases/tag/v3.0)
+As the tool [vspec2id](docs/id.md) has been added the VSS-project has agreed that there no longer is a need to support
+the legacy uuid functionality.
 
-### Overlay Support
+* The parameters `--uuid`/`--no-uuid` are now removed.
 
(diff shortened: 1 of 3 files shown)


sschleemilch:
> I think the new CLI framework expects boolean parameters to be a "pair"

That is not required but optional to do it like that: https://click.palletsprojects.com/en/8.1.x/options/#boolean-flags

erikbosch:
> > I think the new CLI framework expects boolean parameters to be a "pair"
> 
> That is not required but optional to do it like that: https://click.palletsprojects.com/en/8.1.x/options/#boolean-flags

As they recommend having both I have no problems with it.

CHANGELOG.md
```diff
@@ -7,52 +7,74 @@ It includes changes that are included in released version, but also changes plan
...
-The tools vspec2c and vspec2ocf in the contrib folder has been moved to the obsolete folder.
-The background is that they have been broken for a long period and no one has volunteered to fix them.
+The project has been switched to poetry and all tools are available in your PATH once vss-tools is installed via pip.
 
-### Support for specifying unit files
+### Logging arguments
+
+General args have been extended with logging arguments `--log-level` and `--log-file`.
+
+## SAMM Exporter added
```
erikbosch, line 44: Use `###`

erikbosch:
MoM: Ok to merge as part of release prep as trivial

PR #408, opened 2024-10-03
Fixing anytree dependency

erikbosch:
We have started to use 2.9 functionality
Latest is 2.12 but keeping 2.9 for maximum compatibility

Also updating poetry.lock by `poetry.lock`

### DIFF
--- a/poetry.lock
+++ b/poetry.lock
@@ -63,63 +63,83 @@ files = [
 
 [[package]]
 name = "coverage"
-version = "7.6.0"
+version = "7.6.1"
 description = "Code coverage measurement for Python"
 optional = false
 python-versions = ">=3.8"
 files = [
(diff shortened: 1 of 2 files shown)


erikbosch:
FYI @sschleemilch @SebastianSchildt 

If test works I will merge this for 5.0 and create a new rc. But we can discuss wanted approach for main, I see two main alternatives

- Specifying minimum version that works, in this case 2.9
- Specifying latest and greatest, i.e. what we committers actually use and what is used in CI, in this case 2.12

The latter approach gives less risk for problems as detected by @UlfBj (as we do not test older versions like 2.8/2.9 in CI), but might give more problems if you want to run other tools which does not support 2.12 in the same python env/venv.

sschleemilch:
Good finding. We should think about pinning dependencies if we encounter something like that again

PR #409, opened 2024-10-03
Removing uuid feature

erikbosch:
This would be the last step in removing the old UUID functionality. I have generally tried to remove UUID support in all tools, i.e. fields/columns with UUID are removed. One exception though - I am keeping the UUID byte for binary output as I believes tools using that format (VISSR) could have difficult handling a changd format.

What I would like to bring up in a VSS meeting

- Are we (finally) ready to do this?
- Would the changes in this PR be problematic for anyone (we can like currently for binary keep a column/field even if value always remain zero/null/empty)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,7 @@ It includes changes that are included in released version, but also changes plan
 *This document only contain changes introduced in VSS-Tools 3.0 or later!*
 
 
-## Planned changes for VSS-Tools 6.0
+## Implemented changes for VSS-Tools 6.0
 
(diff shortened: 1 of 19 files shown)


SebastianSchildt:
So I do thing wie should do this, now that we have a 5.0 released that still supports it (and we can still backport fixes for some time to come). Moving forward, loosing some code

sschleemilch (approved):

erikbosch:
MoM:

- Please review
- Check if it has any impacts on you

erikbosch:
MoM: Merge

PR #416, opened 2024-10-15
Removed PYPI Readme

sschleemilch:
Concatenating of those READMEs is not what we want for Pypi. Therefore removing the PYPI specific README.


erikbosch (approved):
LGTM

erikbosch:
MoM: Merge

PR #418, opened 2024-10-23
:bug: Fixing model datatype checks order

### LINKED ISSUE #417: Improve assert/datatype check
As of today `is_datatype` may return an exception, which  is not caught

 ```
 File "/home/ubjorken/Proj/covesa/vss-tools/src/vss_tools/model.py", line 214, in check_type_default_consistency
    assert Datatypes.is_datatype(
           ^^^^^^^^^^^^^^^^^^^^^^
  File "/home/ubjorken/Proj/covesa/vss-tools/src/vss_tools/datatypes.py", line 215, in is_datatype
    raise DatatypesException(f"Unsupported datatype: {datatype}")
vss_tools.datatypes.DatatypesException: Unsupported datatype: Types.Resources.FileDescriptor
```

How to reproduce:

Add a node like this with a `default` and a type `DoesNotExist` that (well) does not exist.

```
A.MySignal:
  datatype: DoesNotExist
  default: 23
  type: sensor
  description: Something.
```

I assume we should preferably get an error similar to if we exclude default:

```
A.MySignal:
  datatype: DoesNotExist
  type: sensor
  description: Something.
```
Result (without default):
```

erik@debian6:~/vss-tools/tests/vspec/test_structs$ vspec export yaml --types VehicleDataTypes.vspec -u ../test_units.yaml -q ../test_quantities.yaml --vspec test.vspec --output hej.yaml
[13:11:08] INFO     Loaded 'VSSQuantity', file=/home/erik/vss-tools/tests/vspec/test_structs/../test_quantities.yaml, elements=6           units_quantities.py:34
           INFO     Loaded 'VSSUnit', file=/home/erik/vss-tools/tests/vspec/test_structs/../test_units.yaml, elements=7                    units_quantities.py:34
           INFO     VSpecs (Types) loaded, amount=2                                                                                                  vspec.py:122
           INFO     Dynamic datatypes added=2                                                                                                         main.py:130
           INFO     VSpecs loaded, amount=1                                                                                                          vspec.py:122
           CRITICAL 'A.MySignal' has 1 model error(s):                                                                                                main.py:238
                    [                                                                                                                                            
                        {                                                                                                                                        
                            'type': 'assertion_error',                                                                                                           
                            'loc': (),                                                                                                                           
                            'msg': "Assertion failed, 'DoesNotExist' is not a valid datatype",                                                                   
                            'input': {                                                                                                                           
                                'fqn': 'A.MySignal',                                                                                                             
                                'type': <NodeType.SENSOR: 'sensor'>,                                                                                             
                                'description': 'Something.',                                                                                                     
                                'comment': None,                                                                                                                 
                                'delete': False,                                                                                                                 
                                'deprecation': None,                                                                                                             
                                'constUID': None,                                                                                                                
                                'fka': [],                                                                                                                       
                                'instantiate': True,                                                                                                             
                                'datatype': 'DoesNotExist'                                                                                                       
                            },                                                                                                                                   
                            'ctx': {                                                                                                                             
                                'error': AssertionError("'DoesNotExist' is not a valid datatype")                                                                
                            }                                                                                                                                    
                        }                                                                                                                                        
                    ]                                                                                                                                            
erik@debian6:~/vss-tools/tests/vspec/test_structs$ 
```

There are other scenarios were an exception also are given which maybe should be reworked. And if feasible we should better check datatype before default (then the exception is maybe not a problem, as we should not reach that far)

sschleemilch:
# Assuring a certain order of the data model validation functions.

Some data model checks are calling `Datatypes.is_datatype()` which can throw an exception when we have an unsupported datatype. Since `@model_validator("after")` do not have an order we need to assure that checks that are relying on datatypes run after the base check.

It is not really a critical bug but more like a visual thing to not confront users with a stacktrace.

Fixes #417

### DIFF
--- a/src/vss_tools/model.py
+++ b/src/vss_tools/model.py
@@ -192,7 +192,6 @@ class VSSDataDatatype(VSSData):
             assert is_array(self.datatype), f"'arraysize' set on a non array datatype: '{self.datatype}'"
         return self
 
-    @model_validator(mode="after")
     def check_type_default_consistency(self) -> Self:
         """
         Checks that the default value
@@ -230,7 +229,6 @@ class VSSDataDatatype(VSSData):
                 assert v in self.allowed, f"default value '{v}' is not in 'allowed' list"
         return self
 
-    @model_validator(mode="after")
     def check_allowed_datatype_consistency(self) -> Self:
         """
         Checks that allowed values are valid
@@ -254,6 +252,8 @@ class VSSDataDatatype(VSSData):
     def check_datatype(self) -> Self:
         assert self.datatype in get_all_datatypes(self.fqn), f"'{self.datatype}' is not a valid datatype"
         self.datatype = resolve_datatype(self.datatype, self.fqn)
+        self.check_type_default_consistency()
+        self.check_allowed_datatype_consistency()
         return self
 
     @field_validator("unit")


erikbosch (approved):
LGTM

@UlfBj - this solves one of the issues you experienced

PR #419, opened 2024-10-23
Give better error if unexpected Yaml format

erikbosch:
This intends to give a somewhat better error handling for the scenario when your Yaml/vspec file does not have intended indentation.

Example use case:
```
A:
type: branch
description: Branch A.
```

Which better should be:
```
A:
  type: branch
  description: Branch A.
```

As of today you get a cryptic exception. With this you still get exception, but a somewhat easier to understand

`vss_tools.tree.InvalidInputFormatException: Node 'A' expected to have a list as child`

@UlfBj - this addresses one of the other problems you experienced
@sschleemilch - I believe showing a stack trace to the user here does not help much, but we do not catch other exceptions like `MultipleRootsException` either, but maybe we should. Catch all user-defined exceptions higher up?


sschleemilch:
How about something like this? It is on a lower level when loading the vspec yaml into a dict and checks whether we have toplevel keys that are `None`

image

Patch:
```patch
diff --git a/src/vss_tools/main.py b/src/vss_tools/main.py
index 5a3b8fa..7a39a1b 100644
--- a/src/vss_tools/main.py
+++ b/src/vss_tools/main.py
@@ -24,7 +24,7 @@ from vss_tools.model import (
 )
 from vss_tools.tree import ModelValidationException, VSSNode, build_tree
 from vss_tools.units_quantities import load_quantities, load_units
-from vss_tools.vspec import InvalidSpecDuplicatedEntryException, load_vspec
+from vss_tools.vspec import InvalidSpecDuplicatedEntryException, InvalidSpecException, load_vspec
 
 
 class NameViolationException(Exception):
@@ -215,7 +215,7 @@ def get_trees(
     try:
         types_root = get_types_root(types, unique_include_dirs)
         vspec_data = load_vspec(unique_include_dirs, [vspec] + list(overlays))
-    except InvalidSpecDuplicatedEntryException as e:
+    except (InvalidSpecDuplicatedEntryException, InvalidSpecException) as e:
         log.critical(e)
         exit(1)
 
diff --git a/src/vss_tools/vspec.py b/src/vss_tools/vspec.py
index a36ef8c..f6bb886 100644
--- a/src/vss_tools/vspec.py
+++ b/src/vss_tools/vspec.py
@@ -22,6 +22,9 @@ class IncludeStatementException(Exception):
 class IncludeNotFoundException(Exception):
     pass
 
+class InvalidSpecException(Exception):
+    pass
+
 
 class InvalidSpecDuplicatedEntryException(Exception):
     pass
@@ -77,9 +80,14 @@ class VSpec:
         content = source.read_text()
 
         self.data = yaml.safe_load(content)
+
         if self.data is None:
             self.data = {}
 
+        for key, value in self.data.items():
+            if value is None:
+                raise InvalidSpecException(f"{self.source.absolute()}, Invalid key value: {key}=None")
+
         if prefix:
             tmp_data = {}
             for k, v in self.data.items():
```

sschleemilch (approved):

PR #420, opened 2024-10-23
Binary exporter support for structs. 

UlfBj:


src/vss_tools/exporters/binary.py
```diff
@@ -126,12 +129,15 @@ def cli(
     overlays: tuple[Path],
     quantities: tuple[Path],
     units: tuple[Path],
+    types: tuple[Path],
+    types_output: Path,
+    extend_all_attributes: bool,
```
sschleemilch, line 134: Not using that, right?

erikbosch:
Rebase needed to solve conflicts. Let me know if you experience any problems

src/vss_tools/exporters/binary.py
UlfBj: extend_all_attributes is probably not needed. The other two are referenced at other places in the code so I assume they are needed.

UlfBj:
@erikbosch , It seems uuid should no longer be used? Can it just be removed as a parameter in the export_node() call?

sschleemilch:
> @erikbosch , It seems uuid should no longer be used? Can it just be removed as a parameter in the export_node() call?

It is already removed from that on master

src/vss_tools/exporters/binary.py
sschleemilch: Yeah so please get rid of the unused ones

src/vss_tools/exporters/binary.py
```diff
@@ -140,9 +146,17 @@ def cli(
         uuid=uuid,
         quantities=quantities,
         units=units,
+        types=types,
         overlays=overlays,
     )
+
     log.info("Generating binary output...")
+    if datatype_tree:
+        if types_output:
```
sschleemilch, line 155: What happens if types_output is not defined?
Usually other exporters export the types then bundled into the main file.
I think either it needs to throw an error in that case for the user or defining a default types output file or ...

UlfBj: Test shows that if --types-output is not used in the make command the second if statement becomes false.

sschleemilch: does not really answer my question because `if datatype_tree` is still true then but we are not doing anything with that info in that case

src/vss_tools/exporters/binary.py
```diff
@@ -121,22 +123,32 @@ def cli(
...
+            with open(str(types_output), "wb") as f:
+                export_node(datatype_tree, f)
+            log.info("Binary datatype tree output generated in %s", types_output)
+
     with open(str(output), "wb") as f:
         export_node(tree, f)
-    log.info("Binary output generated in %s", output)
+    log.info("Binary main tree output generated in %s", output)
+
```
erikbosch, line 154: pre-commit complains on this empty line (i.e. two end of line at end of file, should only be one.)

src/vss_tools/exporters/binary.py
erikbosch: Ulf - I suggest you add an else statement, something like

```
else:
  log.warning("To generate datatype tree output you must specify wanted output file")
```
or 

```
else:
  log.warning("Ignoring type tree generation as no file name specified")
```


... that gives a hint that they maybe have forgotten to specify output file.

erikbosch:
MoM:

- Please review until end of week

src/vss_tools/exporters/binary.py
erikbosch: @UlfBj - if you just remove a line break here then two of the CI-failures should dissappear

erikbosch:
If you add the else statement suggested and remove the extra line break at end of file I think this PR should be good to go.

One idea - should you possibly say something about how to use the new feature in
https://github.com/COVESA/vss-tools/blob/master/binary/README.md

UlfBj:
Updted according to comments but did not fix it.

erikbosch:
Merged through #428

Issue #423, opened 2024-10-24
Allowing default in structs


UlfBj:
VSS does not support default being defined in structs, as stated in
[https://covesa.github.io/vehicle_signal_specification/rule_set/data_entry/data_types_struct/index.html#default-values ](https://covesa.github.io/vehicle_signal_specification/rule_set/data_entry/data_types_struct/index.html#default-values )
There is an exception for empty arrays of structs, and there is also the possibility to define the default in the property nodes defining the struct.

Even if the latter makes it possible to declare default values for all members of a struct it comes with the limitation that all usages of this struct definition will get the same default values as a consequence of the structs being defined in a separate Types tree. So to enable a generic default support it should be allowed to define the default in the struct node itself.

The question then is what syntax should be used to define the default value?

Using JSON could provide compatibility with VISS where the value of a struct has the format {“elementname1”:“value1”, …, “elementnameN”: “valueN”}.

The exporter tool may validate this by parsing of the Types tree but it should be possible to disable this validation to support also other syntax models.

erikbosch:
Some information/examples on how we already today handle json/yaml struct info for extended attributes. As you can we support both Yaml and JSON constructs, but I do not know if there are corner cases that gives problem 
```

VSPEC:

VersionVSS.Major:
  datatype: uint32
  type: attribute
  default: 3
  erik: {"a":2,"b":3, "C": "8"}
  description: Supported Version of VSS - Major version.

VersionVSS.Minor:
  datatype: uint32
  type: attribute
  default: 0  
  erik:
    C: '8'
    a: 2
    b: 3
  description: Supported Version of VSS - Minor version.

VersionVSS.Patch:
  datatype: uint32
  type: attribute
  default: 0
  erik: {"a":2,"b":3, "C": [1, 6, 5]}
  description: Supported Version of VSS - Patch version.
```

Generated YAML:

```
Vehicle.VersionVSS.Major:
  datatype: uint32
  default: 3
  description: Supported Version of VSS - Major version.
  erik:
    C: '8'
    a: 2
    b: 3
  type: attribute

Vehicle.VersionVSS.Minor:
  datatype: uint32
  default: 0
  description: Supported Version of VSS - Minor version.
  erik:
    C: '8'
    a: 2
    b: 3
  type: attribute

Vehicle.VersionVSS.Patch:
  datatype: uint32
  default: 0
  description: Supported Version of VSS - Patch version.
  erik:
    C:
    - 1
    - 6
    - 5
[...]

PR #425, opened 2024-10-25
:sparkles: Added struct default validation

sschleemilch:
# Adding `default` validation support for structs

- Generating json schemas dynamically for structs
- We can use those schemas then to validate the `default` input
- Arrays supported

## Also added additional validations
- `min/max` only on numeric datatypes
- `default` being in `min/max` boundary
- forbidding `allowed` to be used with structs

Example:

types.vspec
```yaml
Types:
  type: branch
  description: Types

Types.B:
  type: struct
  description: B

Types.B.x:
  type: property
  description: bx
  datatype: boolean[]

Types.B.y:
  type: property
  description: by
  datatype: numeric

Types.A:
  type: struct
  description: A

Types.A.x:
  type: property
  datatype: uint8
  description: ax

Types.A.y:
  type: property
  datatype: string[]
  description: ay

Types.A.z:
  type: property
  datatype: B[]
  description: az

```

model.vspec with violation (`x` is declared as `uint8`):
```yaml
A:
  type: branch
  description: A

A.X:
  type: attribute
  datatype: Types.A[]
  description: X
  default:
    - x: not a uint8
      y:
      - a
      - b
      z:
        x:
        - true
        y: 300.3
```

```bash
vspec export tree -s model.vspec -t types.vspec
```
Result:
image

Fixes https://github.com/COVESA/vss-tools/issues/423


erikbosch:
Great! Lets discuss on the meeting on Tuesday if we want to allow default for structs. The original limitation was added just as a limitation as we did not at that time prioritized to focus on how to define defaults, so there were no "political" reason to not allow default values.

@UlfBj - as you initiated this (in some way) - could you provide some examples on how you want to use default values for structs, so that we can check if it fits proposed syntax and validation

sschleemilch:
> Great! Lets discuss on the meeting on Tuesday if we want to allow default for structs. The original limitation was added just as a limitation as we did not at that time prioritized to focus on how to define defaults, so there were no "political" reason to not allow default values.
> 
> @UlfBj - as you initiated this (in some way) - could you provide some examples on how you want to use default values for structs, so that we can check if it fits proposed syntax and validation

I think the complete **default** functionality is discussable whether we want to have it at all. In my opinion it is out of scope for modelling the structure of data. It mixes defining a data structure and actually using it

erikbosch:
> 
> I think the complete **default** functionality is discussable whether we want to have it at all. In my opinion it is out of scope for modelling the structure of data. It mixes defining a data structure and actually using it

Historically we have promoted `default` primarily for attributes that change rarely. The idea has been that implementations (like VISSR, Kuksa) shall use default value unless any other value is given. In some cases like `Vehicle.VersionVSS.Major` the implementation will likely never give the value by some other mechanism. In many other cases it can be argued if the standard catalog should contain default values or not, some of the existing ones seems quite useless.

But I think there is a value in defining a syntax, so that the same spec with default info can be used both for VISSR and Kuksa. But I am open to if it we should consider it to be part of "core CSS syntax", or just a "VSS profile for default", similar to how we sometimes discuss if there should be "standardized" "VSS DBC profile" or "VSS Implementation Type profile" that can tools can decide to support or not.

sschleemilch:
> https://raw.githack.com/COVESA/vehicle-information-service-specification/main/spec/VISSv3.0_TransportExamples.html#wss-search-read

Yeah, maybe we can distinguish that a bit.
It feels a bit like an init file then that is being passed to services that work with vss.
But It would not need the full blown vss to do that. On the other hand it probably does not hurt to have them supported.
Users could then always define overlays to set default values.

UlfBj:
Example:
```
DownloadFile:
  type: actuator
  datatype: Types.Resources.FileDescriptor
#  default: {"name":"downloadfile.txt", "hash":"20e87e71b6948d6e6dd11d776e9be79c374751bb", "uid":"1d878212")
  description: File to be used by the vehicle. Default contains internal filesystem path.

UploadFile:
  type: sensor
  datatype: Types.Resources.FileDescriptor
#  default: {"name":"uploadfile.txt", "hash":"20e87e71b6948d6e6dd11d776e9be79c374751bb", "uid":"1d878212")
  description: File created by the vehicle. Default contains internal filesystem path incl filename.
```

erikbosch:
MoM:

- Please review
- Erik: Please comment if you have objections supporting it for struct

sschleemilch:
> Example:
> 
> ```
> DownloadFile:
>   type: actuator
>   datatype: Types.Resources.FileDescriptor
>   default: {"name":"downloadfile.txt", "hash":"20e87e71b6948d6e6dd11d776e9be79c374751bb", "uid":"1d878212")
>   description: File to be used by the vehicle. Default contains internal filesystem path.
> 
> UploadFile:
>   type: sensor
>   datatype: Types.Resources.FileDescriptor
>   default: {"name":"uploadfile.txt", "hash":"20e87e71b6948d6e6dd11d776e9be79c374751bb", "uid":"1d878212")
>   description: File created by the vehicle. Default contains internal filesystem path incl filename.
> ```

It works fine except I do not understand why JSON is being used 😕 :

MODEL:
```yaml
Vehicle:
  type: branch
  description: Vehicle

Vehicle.DownloadFile:
  type: actuator
  datatype: Types.Resources.FileDescriptor
  default: 
    name: downloadfile.txt
    hash: 20e87e71b6948d6e6dd11d776e9be79c374751bb
    uid: 1d878212
  description: File to be used by the vehicle. Default contains internal filesystem path.

Vehicle.UploadFile:
  type: sensor
  datatype: Types.Resources.FileDescriptor
  default: 
    name: uploadfile.txt
    hash: 20e87e71b6948d6e6dd11d776e9be79c374751bb
    uid: 1d878212
  description: File created by the vehicle. Default contains internal filesystem path incl filename.
```

TYPES:
```yaml
Types:
  type: branch
  description: Types

Types.Resources:
  type: branch
  description: Resources

Types.Resources.FileDescriptor:
  type: struct
  description: File Descriptor

Types.Resources.FileDescriptor.name:
  type: property
  datatype: string
  description: Name

Types.Resources.FileDescriptor.hash:
  type: property
  datatype: string
  description: Hash

Types.Resources.FileDescriptor.uid:
  type: property
  datatype: string
  description: UID
```

Works fine and when e.g. removing `name`:
image

PR #426, opened 2024-10-28
Add support for handling of RegularExpression definitions for string based properties

Kostadin-Ivanov:
Add support for handling of RegularExpression definitions for string based properties.

The idea is to be able to specify a Regular Expression field, where we can define a RegExp string, which can be used for string based property specifications.

An example would be the **Vehicle.VehicleIdentification.VIN** which currently has definition of:
```
VehicleIdentification.VIN:
  datatype: string
  type: attribute
  description: 17-character Vehicle Identification Number (VIN) as defined by ISO 3779.
```

As per above description, there is not available RegExp or other way to validate such strings and users are "left" to deal with the validation of this field on their own.

Currently there is only **min/max** constraints, which are for number based properties.

With this update, we would like to allow option to add additional sort of validation (constraints) for string based values like VehicleIdentification.VIN, which can easy be used by VSS users.

As result, we are aiming to be able to have a RegExp definition for the VIN and any other VSS properties which might need such validation as follows:
```
VehicleIdentification.VIN:
  datatype: string
  type: attribute
  description: 17-character Vehicle Identification Number (VIN) as defined by ISO 3779.
  pattern: ^([0-9A-HJ-NPR-Z]{3})([0-9A-HJ-NPR-Z]{6})([0-9A-HJ-NPR-Z]{4}[0-9]{4})$
```

**NOTE:** the specified **pattern** is tested on: https://regex101.com/, with a test string of: **1CPH423GAP71E2745** and validates VIN strings as per quoted [ISO 3779](https://cdn.standards.iteh.ai/samples/52200/7d8a69aee84c4ad28231053f49f4966e/ISO-3779-2009.pdf) on below rules:

1. VIN length of 17 chars where:
1.1 First 3 characters are for the **VMI**: 1CP
1.2 Next 6 characters are for the **VDS**: H423GA
1.3 Last 8 characters are for the **VIS**: P71E2745
2. The last 4 characters of the **VIS** part should always be numbers
3. Letters: I, O and Q shall not be used

If required, a better RegExp might be provided / updated using overlays by users, but the above one can be used as a general example, which covers 6 rules from the above mentioned ISO Standard.

With regards to update to the **Vehicle.VehicleIdentification.VIN**, we provided another Pull Request: [Add RegExp pattern for VehicleIdentification.VIN property.](https://github.com/COVESA/vehicle_signal_specification/pull/786) in the [COVESA VehicleSignalSpecification](https://github.com/COVESA/vehicle_signal_specification) which currently fails, because of the **extra property**: **pattern** until this PR gets approved and released to the used by the VehiclSignalSpecification VSS-Tools.

In addition, this update also reflects and the **VSS - samm** exporter so the newly introduced node attribute: **pattern** to be properly handled in generated ESMF - SAMM aspect models with corresponding SAMM - RegularExpressionConstraint.


src/vss_tools/model.py
```diff
@@ -178,6 +178,10 @@ class VSSDataDatatype(VSSData):
     arraysize: int | None = None
     min: int | float | None = None
     max: int | float | None = None
+    # Field, used to allow definition of Regular Expression constraints
+    # for string based property nodes.
+    # Example: VSS - VehicleIdentification.VIN property
+    pattern: str | None = None
```
sschleemilch, line 184: We would also need a model validator function for `pattern` to evaluate it. It should fail when being defined and:
- `datatype` is  not a string
- `default`  being set and is not matching

src/vss_tools/exporters/samm/helpers/vss_helper.py
```diff
@@ -260,14 +260,18 @@ def get_node_description(vss_node: VSSNode) -> str:
 
 
 def has_constraints(vss_node: VSSNode) -> bool:
+    # NOTE: there is a special case for string attributes, which can have a RegExp: pattern field.
+    #       Example: VSS - VehicleIdentification.VIN property
     return (
         hasattr(vss_node.data, "type")
-        and vss_node.data.type in [NodeType.ACTUATOR, NodeType.SENSOR]
```
sschleemilch, line 265: Since the goal seems to be to check against attributes of the `VSSDataDatatype` node type it can probably be simplified to:

```python
data = node.data
if isinstance(data, VSSDataDatatype):
    return data.max or data.min or data.pattern
return False
```
Is `VSSDataProperty` not  being covered on purpose?

sschleemilch (changes requested):

src/vss_tools/model.py
Kostadin-Ivanov: Thank you Sebastian, I will update the model.

src/vss_tools/exporters/samm/helpers/vss_helper.py
Kostadin-Ivanov: Thank you.
I will simplify this as advised.

sschleemilch: Probably need to check them against `None`. So my implementation is not completely correct I suppose. At least the `max` and `min` ;-)

Kostadin-Ivanov: Sure, I will try to do my best here :)

Kostadin-Ivanov: Hello @sschleemilch , on the question about VSSDataProperty. The code above was a bit obsolete, since I had below check:

```python
vss_node.data.type in [NodeType.ACTUATOR, NodeType.SENSOR, NodeType.ATTRIBUTE]
```
which I guess adds for the VSSDataProperty check. However, since I think that all: sensor, actuator and attribute are children of the VSSDataDatatype, I updated the check as you advised and now it looks like:

```python
def has_constraints(vss_node: VSSNode) -> bool:
    return isinstance(vss_node.data, VSSDataDatatype) and (
        vss_node.data.max or vss_node.data.min or vss_node.data.pattern
    )
```

I tested this on the whole VSS spect and it seems to be fine for every node which has min/max and pattern constraints.

Kostadin-Ivanov: Done

src/vss_tools/model.py
Kostadin-Ivanov: Done

src/vss_tools/model.py
```diff
@@ -277,6 +281,31 @@ def check_datatype_matching_allowed_unit_datatypes(self) -> Self:
...
+        3. if default value is provided it is matching the specified pattern
+        """
+        if self.pattern:
+            assert bool(type(self.pattern) is str), f"Specified pattern: '{self.pattern}' must be a string"
+
+            # Datatypes.TUPLE[0] is the string name of the type.
+            allowed_for = f"It is allowed only for '{Datatypes.STRING[0]}' and '{Datatypes.STRING_ARRAY[0]}' types."
+            assert bool(
+                Datatypes.get_type(self.datatype) in [Datatypes.STRING, Datatypes.STRING_ARRAY]
+            ), f"Field 'pattern' is not allowed for '{self.fqn}' of type: '{self.datatype}'. {allowed_for}"
```
sschleemilch, line 300: The `fqn` info is already in the rendered exception and does not need to be given.

src/vss_tools/model.py
```diff
@@ -277,6 +281,31 @@ def check_datatype_matching_allowed_unit_datatypes(self) -> Self:
...
+
+        1. string only
+        2. defined only for string typed nodes i.e., STRING and STRING_ARRAY
+        3. if default value is provided it is matching the specified pattern
+        """
+        if self.pattern:
+            assert bool(type(self.pattern) is str), f"Specified pattern: '{self.pattern}' must be a string"
+
+            # Datatypes.TUPLE[0] is the string name of the type.
+            allowed_for = f"It is allowed only for '{Datatypes.STRING[0]}' and '{Datatypes.STRING_ARRAY[0]}' types."
```
sschleemilch, line 297: I would make this a bit shorter. Like 
```python
f"allowed: {[Datatypes.STRING[0], Datatypes.STRING_ARRAY[0]]}"
```

src/vss_tools/model.py
```diff
@@ -277,6 +281,31 @@ def check_datatype_matching_allowed_unit_datatypes(self) -> Self:
...
+            assert bool(
+                Datatypes.get_type(self.datatype) in [Datatypes.STRING, Datatypes.STRING_ARRAY]
+            ), f"Field 'pattern' is not allowed for '{self.fqn}' of type: '{self.datatype}'. {allowed_for}"
+
+            if self.default:
```
sschleemilch, line 302: if default is a list we need to iterate and do `re.match` on each element.

It can be something like:

```python
            check_values = [self.default]
            if is_array(self.default):
                check_values = self.default
            for value in check_values:
                assert re.match(self.pattern, value), f"...."
```

sschleemilch (changes requested):
Please add some tests for the pattern functionality.
Can be completely separate tests similar to how it has been done here: https://github.com/COVESA/vss-tools/pull/425/files

No need to care about exporters at that level

src/vss_tools/exporters/samm/helpers/vss_helper.py
```diff
@@ -260,15 +260,8 @@ def get_node_description(vss_node: VSSNode) -> str:
...
-        hasattr(vss_node.data, "type")
-        and vss_node.data.type in [NodeType.ACTUATOR, NodeType.SENSOR]
-        and (
-            hasattr(vss_node.data, "max")
-            and vss_node.data.max is not None
-            or hasattr(vss_node.data, "min")
-            and vss_node.data.min is not None
-        )
+    return bool(
+        isinstance(vss_node.data, VSSDataDatatype) and (vss_node.data.max or vss_node.data.min or vss_node.data.pattern)
```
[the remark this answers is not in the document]
Kostadin-Ivanov: I will update it for **not None**, totally forgot about the 0 :)

src/vss_tools/model.py
```diff
@@ -277,6 +281,31 @@ def check_datatype_matching_allowed_unit_datatypes(self) -> Self:
...
+    def check_datatype_pattern(self) -> Self:
+        """
+        Checks that regular expression datatype 'pattern' field is:
+
+        1. string only
+        2. defined only for string typed nodes i.e., STRING and STRING_ARRAY
+        3. if default value is provided it is matching the specified pattern
+        """
+        if self.pattern:
+            assert bool(type(self.pattern) is str), f"Specified pattern: '{self.pattern}' must be a string"
```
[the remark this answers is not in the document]
Kostadin-Ivanov: Sure, I noticed it when I was testing it manually, but forgot to remove this check.

src/vss_tools/model.py
Kostadin-Ivanov: I will update the error response.

src/vss_tools/model.py
Kostadin-Ivanov: Thank you for this catch.

src/vss_tools/model.py
```diff
@@ -277,6 +281,31 @@ def check_datatype_matching_allowed_unit_datatypes(self) -> Self:
...
+
+            # Datatypes.TUPLE[0] is the string name of the type.
+            allowed_for = f"It is allowed only for '{Datatypes.STRING[0]}' and '{Datatypes.STRING_ARRAY[0]}' types."
+            assert bool(
```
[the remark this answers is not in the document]
Kostadin-Ivanov: Ok.
I will check it and probably remove the casting.

Kostadin-Ivanov:
> Please add some tests for the pattern functionality. Can be completely separate tests similar to how it has been done here: https://github.com/COVESA/vss-tools/pull/425/files
> 
> No need to care about exporters at that level

I was thinking about this and will add some unit tests.
Thank you for the hint :)

src/vss_tools/exporters/samm/helpers/vss_helper.py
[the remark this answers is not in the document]
Kostadin-Ivanov: Done

src/vss_tools/model.py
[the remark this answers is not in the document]
Kostadin-Ivanov: Done

src/vss_tools/model.py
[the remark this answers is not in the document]
Kostadin-Ivanov: Done

src/vss_tools/model.py
Kostadin-Ivanov: Done

src/vss_tools/model.py
Kostadin-Ivanov: Done

src/vss_tools/model.py
Kostadin-Ivanov: Done.

Since we are going to use patterns for string arrays, probably would never happen, but just in case I also added validation of **allowed** data.

Kostadin-Ivanov:
Hello @sschleemilch , 

I added some, I think :), integration tests to cover the validations of the usage of **pattern** datatype field. 

Ah, I forgot to add check for wrong **pattern** type, i.e. a **number** instead of a **string**. Please let me know if it is OK like that, or I should add the **number** check, and I will do it straight away :)

sschleemilch (approved):
looks good now

Kostadin-Ivanov:
Thank you very much :)

erikbosch:
> Thank you very much :) Now it is time to get it merged and see when it will get released.
> 
> Just one question, maybe also for @erikbosch .
> 
> Since this will add sort of new functionality related to properties definitions. Mainly for string based, but still, shall I (we) update some place in the documentation to mention the usage of **pattern** to define RegExp constraints on string based properties?

I added a PR at https://github.com/COVESA/vehicle_signal_specification/pull/788 mentioning something about both `pattern` and proposed changes to `default`, as well as some other minor changes. Please check if the description for pattern fits your intentions and propose improvements, if needed. I suggest that we check on the VSS meeting if there are any general objections to adding pattern, if not lets check if there are remaining issues on individual PRs.

Kostadin-Ivanov:
Thank you @erikbosch .
I checked your PR and the **pattern** part looks OK to me. I just left one comment there, with regards to the datatype `string` part.
The model, allows to specify **pattern** and for string arrays i.e., `string[]` datatypes.

erikbosch:
MoM: Ok to merge

PR #428, opened 2024-11-11
Binary exporter support for structs.

erikbosch:
Based on https://github.com/COVESA/vss-tools/pull/420


erikbosch:
@UlfBj - can you check this one. It should be identical to your PR except that I replaced some tabs with blanks. if ok I will merge this one and close yours.

PR #429, opened 2024-11-11
💄 Replaced poetry with uv

sschleemilch:
- Way faster
- Can also install/handle Python versions
- No change for consumers
- Should simplify contributor's "does not work on my machine" effects due to the feature of using a defined `.python-version` and also downloads it


README.md
```diff
@@ -62,90 +63,82 @@ It is not guaranteed that newer or older versions of vss-tools can successfully
...
+Alternatively you can use `uv` to run the `vspec` tool:
+
 ```bash
-poetry run vspec --help
+uv run vspec --help
 ```
 
-direnv:
+If you have a working setup of [direnv](https://direnv.net/), just navigating into the directory
+if enough (after an initial `uv sync`)
```
sschleemilch, line 126: typo

erikbosch:
MoM:

- Please review
- Can be merged next week if no comments or outstanding discussions/remarks.

erikbosch (commented):
LGTM. Did some tests locally.

When merged I can add some words in the wiki on how to update dependencies with `uv lock -U`

Will wait with merging until Monday in case someone else wants to review

Issue #436, opened 2025-01-23
Concept for specifying VSS-tools profiles?


erikbosch:
A topic that is mentioned now and then is the possibility to add and possibly standardize additional metadata to VSS signals to describe "feature of interest", "safety/security classification", "CAN dbc mapping" and similar. In VSS that is already possible by using extended attributes.

```
Vehicle.Speed
  type: sensor
  datatype: float
  unit: km/h
  my_own_attribute_1: 123
  my_own_attribute_2: "free text"
```
The tooling supports extended attributes, if you specify `-e my_own_attribute_1 -e my_own_attribute_2` no warnings will be given and the extra attribute will be present in generated files like yaml, csv, json.

I am thinking of two potential "improvements"

**Concept for defining profiles**

To provide better re-usability and control of extended attributes I am thinking if it would make sense to define a profile format, possibly like below. That profile could be loaded by vss-tools and be used to validate signals in the catalog. That could be useful if you for a particular deployment need to provide some extended attributes for every signal.

```
myprofile:
  mandatory: true # If true all signals must use this profile, so they must have "my_own_attribute_1"
  my_own_attribute_1:
    mandatory: true
    unit: uint16
    description: "Describes xxxxx"  
  my_own_attribute_2:
    mandatory: false
    unit: string
    description: "Describes xxxxx"  
```

**Managing a catalog of profiles**

It would be possible for VSS (or VSS-tools) repo to host a folder with contributed and/or recommended profiles. I.e. if a COVESA member or a downstream project has defined a profile they think is generally useful, then it could be contributed to COVESA so that also other downstream solutions could use it.

Note that the profile catalog would only define the extended attributes that are possible to use, it would (most likely) not provide actual values to use for the signals in VSS Standard catalog as that would differ from case to case.

erikbosch:
MoM:

- Presented at meeting
- Ulf: Correct to add in vss-tools rather vss. For VISSr we use "validate"
- Ted: Used for data campaigns in commercial vehicles
- Please discuss in issue

jdacoello:
Unfortunately, I am not fully understanding your goal with this feature. I think the following work flow would be ideal:
- The specification has all the possible properties (i.e., sensor, actuator, attribute).
- Before exporting it, one can arbitrarily select a sub set of what has been specified.

Is this what you are trying to achieve? If so, it seems that the proposed approach is too complicated.

SebastianSchildt:
[...]

PR #440, opened 2025-03-03
VSS profile

erikbosch:
Related to #436 

This is an example on how VSS-tools possible could define a profile syntax and collect contributed profiles. Published as an example to be discussed at VSS meeting March 4th

### DIFF
--- a/docs/extensions/can.md
+++ b/docs/extensions/can.md
@@ -1,0 +1,143 @@
+# CAN DBC Profile
+
+This profile can be used to specify how VSS signals can be mapped to CAN signals defined in DBC.
+For more information on DBC see this [tutorial](https://www.csselectronics.com/pages/can-dbc-file-database-intro).
+This profiles is based on the mapping definined in the [Eclipse KUKSA project](https://github.com/eclipse-kuksa/kuksa-can-provider/blob/main/mapping/README.md).
+
+The profile define two extended attributes, `vss2dbc` and `dbc2vss`.
+The reason for this is that handling is different.
+Exact semantics for the information is unless specified in this document up to the implementation.
+
+
+## Mapping syntax
+
+The syntax for a DBC definition of a signal (in an overlay) is specified below.
+
+Syntax
+
+```yaml
+<VSS Signal name>:
+  type: <VSS type>
+  datatype: <VSS datatype>
+  vss2dbc:
+    signal: <DBC signal name>
+    [transform: ...]
+  dbc2vss:
+    signal: <DBC signal name>
+    [interval_ms: <interval in milliseconds>]
+    [on_change: {true|false}]
+    [transform: ...]
+```
+
+* `dbc2vss` shall be specified if there is a need to transform from DBC/CAN to VSS.
+  It is assumed that tra
+* `vss2dbc` shall be specified if there is a need to transform from VSS to DBC/CAN.
+
+Specifying DBC signal name is mandatory. It must correspond to a DBC signal name defined in a DBC file.
+
+The `transform` entry can be used to specify conversion between DBC and VSS representation.
+If transform is not specified values will be transformed as is.
+
+### dbc2vss
+
+It is possible that a VSS-based system does not need to get updated signal values as frequent as they are sent on CAN.
+
+`interval_ms` and `on_change` are optional and control under which conditions a value shall be forwarded.
+The `interval_ms` value indicates the (minimum) interval between updates in milliseconds. A value of `1000` indicates that the VSS system wants at least 1000 milliseconds between each update.
+The `on_change: true` argument specifies that the VSS signal only shall be sent if the DBC value has changed.
+
+If both are given it shall be considered as an AND-criteria. An update will be sent if DBC value has changed and the given interval has passed.
+If none of them are specified it corresponds to that the signal shall be sent as soon as value has changed, i.e. `on_change: true`.
+
+### vss2dbc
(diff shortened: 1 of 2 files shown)


SebastianSchildt:
I want to suggest to take the chance to define profiles in a "existing" language  instead of rolling our own.



GraphQL interfaces come to mind, they even seem to be composable

```graphql
interface QualityProfile {
   quality: Int! //maybe uint16!....
   qualityDescription: String
}
```

```graphql
interface DBCProfile {
   DBCSignal: String
   DBCtransform: DBCTransform
   DBCdescription: String
}
```

```graphql
type DBCTransform {
  transform: String
  mapping: ...
}
```

The advantage being: It seems a simple way to specify what a metadata you want, it IS defined, so in case of corner cases (what are the rules when field names collide?), we can _look them up_ instead of making new ones. 

While this would of course fit with somebody using GraphQL, I think it is not contingent on that: This works fine with VSS as is, and I think it is not harder to parse than the custom format suggested in the PR.

Taking the example from the PR


```yaml
  source:
    mandatory: true
    description: "Gives some info"
    xpos:
      list: True # Indicates that children may be in a list
      a:
        datatype: uint8
        description "A value"
      b:
        datatype: uint8
        description "B value"
    ypos:
      datatype: uint8
      description "Y value"
  quality:
    datatype: uint8
    description "Q value"
```

Might translate to sth like


```yaml
interface QualityProfile {
   Quality: uint16!
   source: QualitySource!
}

type QualitySource {
  xpos:  [Int2Tuple!]!
  ypos:  uint8!
}

type Int2Tuple {
  a: Int!
  b: Int!
}
```





As I am not an expert in this technology I'd like an opinion @jdacoello  , whether this is feasible or should be considered "misuse".

sschleemilch:
> I want to suggest to take the chance to define profiles in a "existing" language instead of rolling our own.
> 
> GraphQL interfaces come to mind, they even seem to be composable
> 
> ```graphql
> interface QualityProfile {
>    quality: Int! //maybe uint16!....
>    qualityDescription: String
> }
> ```
> 
> ```graphql
> interface DBCProfile {
>    DBCSignal: String
>    DBCtransform: DBCTransform
>    DBCdescription: String
> }
> ```
> 
> ```graphql
> type DBCTransform {
>   transform: String
>   mapping: ...
> }
> ```
> 
> The advantage being: It seems a simple way to specify what a metadata you want, it IS defined, so in case of corner cases (what are the rules when field names collide?), we can _look them up_ instead of making new ones.
> 
> While this would of course fit with somebody using GraphQL, I think it is not contingent on that: This works fine with VSS as is, and I think it is not harder to parse than the custom format suggested in the PR.
> 
> Taking the example from the PR
> 
> ```yaml
>   source:
>     mandatory: true
>     description: "Gives some info"
>     xpos:
>       list: True # Indicates that children may be in a list
>       a:
>         datatype: uint8
>         description "A value"
>       b:
>         datatype: uint8
>         description "B value"
>     ypos:
>       datatype: uint8
>       description "Y value"
>   quality:
>     datatype: uint8
>     description "Q value"
> ```
> 
> Might translate to sth like
> 
> ```yaml
> interface QualityProfile {
>    Quality: uint16!
>    source: QualitySource!
> }
> 
> type QualitySource {
>   xpos:  [Int2Tuple!]!
>   ypos:  uint8!
> }
> 
> type Int2Tuple {
>   a: Int!
>   b: Int!
> }
> ```
> 
> As I am not an expert in this technology I'd like an opinion @jdacoello , whether this is feasible or should be considered "misuse".

JSON Schema would be an "existing language".
How would you connect the GraphQL Types to the VSS Elements?

SebastianSchildt:
".
> 
> JSON Schema would be an "existing language". How would you connect the GraphQL Types to the VSS Elements?

Agreed on JsonSchema.

For bindng/limiting GraphQL Types: Basically I would only allow „primitive VSS“ Types as member datatypes. So the „Int“ in the tuple toy example was probably not good, should have been one of the VSS types (similar lije in JSONSchema you would likely not limit yourself to native JSON datatypes only.

Nesting of Types (if we allow it) woul easily transfer to the vspec yaml view, similar to how in current VSS vanilla overlay for DBC everything is grouped under dbc2vss or vss2dbc

sschleemilch:
In the end I am also fine with GraphQL also regarding the possible future of VSPEC being modelled completely using GraphQL...

erikbosch:
I agree that if we want to define a syntax for defining profiles. it should better be something that would work in GraphQL, as that seems to be the language that most are interested in for "VSS next gen". For now there anyway does not seem to be anyone that has interest in developing a profile validator for vss-tools, it rather seems that the validation (if any) will happen in the tool using the info (Kuksa, VISSR, ....)

The new PRs that recently has been submitted focus on vspec/json examples and informal definition of the profiles, so that is maybe a sufficient scope for now.

slawr:
Re the CAN mapping I can think of a few projects defining something in that regard:
1. Kuksa CAN provider - presumably the mapping used here
2. VISSR feeder template has a CAN capability
3. AoS Edge has a boardnet to VSS feature in its VISS implementation - can't recall the details
4. Possibly something from the work Tietoevry did for the early Autosar gateway PoC?
5. Do Autosar WG Cloud have something from their work subsequent to 4?

erikbosch:
Replaced by https://github.com/COVESA/vehicle_signal_specification/pull/817

PR #443, opened 2025-03-18
 add cybersecurity profile

cpodalak:
add cybersecurity profile as extension to VSS
Related to https://github.com/COVESA/vss-tools/issues/436.


cpodalak:
Refer  #445

PR #448, opened 2025-03-31
Stat scripts

Giaccomole:
@sschleemilch 

Statistics exporters, to be used both on their own and to be imported by VSS docs generator for graphics. 

(There are version radio buttons on Pie Chart to see how the model grew across versions, and the reflexive .csv generator populates those by relying on another .csv which keeps track of the statistics before.)


src/vss_tools/stats/piechart.py
```diff
@@ -0,0 +1,179 @@
...
+    # The actual data of first 5 version for fallbacks
+    # template_data = {
+    #     'Type': ['Attribute', 'Branches', 'Sensors', 'Actuators'],
+    #     'V2': [78, 117, 203, 101],
+    #     'V3': [86, 117, 263, 128],
+    #     'V4': [97, 147, 286, 179],
+    #     'V5': [110, 131, 313, 195]
+    # }
+
+    latest = pd.read_csv("docs-gen/static/data/piechart.csv")
```
sschleemilch, line 146: should not be hard coded but a cli argument

src/vss_tools/stats/piechart.py
```diff
@@ -0,0 +1,179 @@
...
+
+    metadata = data_metadata
+
+    metadata["Default"] = pd.to_numeric(metadata["Default"], errors="coerce")
+
+    major_version = None
```
sschleemilch, line 152: There is a function already in the samm exporter that extracts the Version from the vss_tree: https://github.com/COVESA/vss-tools/blob/master/src/vss_tools/exporters/samm/__init__.py#L56

Since it is the second time where the VSS Version seems to be of interest: Try to reuse that function and place it in a more central place

src/vss_tools/stats/piechart.py
```diff
@@ -0,0 +1,179 @@
...
+        if column_name not in latest.columns:
+            latest[column_name] = pd.Series(
+                [counts["Attributes"], counts["Branches"], counts["Sensors"], counts["Actuators"]]
+            )
+        version_cols = [col for col in latest.columns if col.startswith("V")]
+        for i in range(len(version_cols) - 1):
+            v1 = int(version_cols[i][1])
+            v2 = int(version_cols[i + 1][1])
+            if v2 != v1 + 1:
+                print(f"MISSING VERSION: V{v1+1}")
```
sschleemilch, line 177: use logging instead of prints

src/vss_tools/stats/radial.py
```diff
@@ -0,0 +1,116 @@
...
+    types: tuple[Path],
+    types_output: Path | None,
+    pretty: bool,
+    extend_all_attributes: bool,
+):
+    """
+    Export JSON Data for Radial Tree.
+    """
+
+    tree, datatype_tree = get_trees(
```
sschleemilch, line 69: Same here, copy pasting the JSON exporter is not the way to go here. Just make the JSON the exporter produced an Input to the stats cli

src/vss_tools/stats/radial.py
```diff
@@ -0,0 +1,116 @@
...
+    if datatype_tree:
+        types_data: dict[str, Any] = {datatype_tree.name: get_data(datatype_tree, extend_all_attributes)}
+        if not types_output:
+            log.info("Adding custom data types to signal dictionary")
+            signals_data["ComplexDataTypes"] = types_data
+        else:
+            with open(types_output, "w") as f:
+                json.dump(types_data, f, indent=2, sort_keys=True)
+
+    def build_json_structure(obj, parent_name=None):
```
sschleemilch, line 94: I see no real reason to make this an inline function.
Also I have trouble understanding what this function tries to do

src/vss_tools/stats/radial.py
```diff
@@ -0,0 +1,116 @@
...
+            else:
+                for prop, prop_value in value.items():
+                    if prop != "children":
+                        item[prop] = prop_value
+            result.append(item)
+
+        result.sort(key=lambda x: ("children" not in x, x.get("type", "")))
+        return result
+
+    new_data = {
```
sschleemilch, line 109: what is new about it?
Essentially, those are the stats, right? Maybe just name it like that: `stats = {}`

src/vss_tools/stats/sankey.py
```diff
@@ -0,0 +1,152 @@
...
+    rows = [get_header(entry_type, with_instance_column)]
+    add_rows(rows, tree, with_instance_column)
+    if generic_entry and datatype_tree:
+        add_rows(rows, datatype_tree, with_instance_column)
+
+    if not generic_entry and datatype_tree:
+        rows = [get_header("Node", with_instance_column)]
+        add_rows(rows, datatype_tree, with_instance_column)
+
+    data_metadata = pd.DataFrame(rows[1:], columns=rows[0])
```
sschleemilch, line 131: Same here, copy pasting csv generation not necessary

src/vss_tools/stats/piechart.py
```diff
@@ -0,0 +1,179 @@
...
+from anytree import PreOrderIter  # type: ignore[import]
+
+import vss_tools.cli_options as clo
+from vss_tools import log
+from vss_tools.main import get_trees
+from vss_tools.tree import VSSNode
+from vss_tools.utils.misc import getattr_nn
+
+
+def get_header(entry_type: str, with_instance_column: bool) -> list[str]:
```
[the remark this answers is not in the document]
Giaccomole: I had put the exporters back to back so I would not rely on interim static files, if this is not a problem I can change this.

erikbosch:
Closing it as replaced by #453

PR #451, opened 2025-05-16
add typealias for min/max and allowed/default values

wuan:
The introduced type aliases make it easier to integrate `vss-tools` into a typed environment as it simplifies the type expressions:

```
class VSSDataDatatype(VSSData):
    ...
    min: Number | None = None
    max: Number | None = None
    ...
    allowed: ValueList | None = None
    ...
```

It would even allow to go back to the current (version 5.0) definition of `default` as
```
    default: list[str | int | float | bool] | str | int | float | bool | None = None
```
would become
```
    default: ValueListOrScalar | None = None
```

### DIFF
--- a/src/vss_tools/model.py
+++ b/src/vss_tools/model.py
@@ -7,7 +7,7 @@
 # SPDX-License-Identifier: MPL-2.0
 import re
 from enum import Enum
-from typing import Any
+from typing import Any, TypeAlias
 
 import jsonschema
 from pydantic import (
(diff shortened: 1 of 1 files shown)


sschleemilch (approved):
Thanks for the contribution. Makes sense and is a good addition

wuan:
Yes, our internal implementation uses
```
type Name = <definition>
```
but this is only available starting with 3.12.

PR #452, opened 2025-06-06
feat: allowing redefinition of units and quantities

sschleemilch:
# About

Right now we are not allowing users to overwrite units and quantity definitions.
However, it probably is more flexible to allow users to do that if they want.

It is already allowed to pass more than one unit and quantity file.
This change enables users to overwrite existing unit and quantity definitions with their own (maybe extended) one.

# Example

## Files

model.yaml
```yaml
Vehicle:
  type: branch
  description: Vehicle

Vehicle.ProductionDate:
  type: attribute
  description: ProductionDate
  datatype: string
  unit: unix-time
```

quantities.yaml
```yaml
datetime:
  definition: Specific point in time
  remark: This quantity is not described in ISO 80000.
  comment: Example usage of this quantity could be ISO 8601 string representation of date and time, or UNIX timestamp
```

units.yaml
```yaml
unix-time:
  definition: Number of non-leap seconds which have passed since 00:00:00 UTC on Thursday, 1 January 1970
  unit: UNIX Timestamp
  quantity: datetime
  allowed-datatypes: ["uint32", "uint64", "int64"]
```

units_overlay.yaml
```yaml
unix-time:
  definition: Number of non-leap seconds which have passed since 00:00:00 UTC on Thursday, 1 January 1970
  unit: UNIX Timestamp
  quantity: datetime
  allowed-datatypes: ["uint32", "uint64", "int64", "string"]
```

## Invocation

Without overlay:
```bash
vspec export tree -s test.vspec -u units.yaml -q quantities.yaml
```
Result:
image

With overlay:
```bash
vspec export tree -s test.vspec -u units.yaml -q quantities.yaml -u units_overlay.yaml
```

image

### DIFF
--- a/src/vss_tools/units_quantities.py
+++ b/src/vss_tools/units_quantities.py
@@ -14,10 +14,6 @@ from vss_tools import log
 from vss_tools.model import ModelValidationException, VSSQuantity, VSSUnit
 
 
-class UnitQuantityRedefinitionException(Exception):
-    pass
-
-
 class MalformedDictException(Exception):
     pass
 
@@ -25,7 +21,7 @@ class MalformedDictException(Exception):
 def load_units_or_quantities(
     files: list[Path], class_type: type[VSSUnit | VSSQuantity]
 ) -> dict[str, VSSUnit | VSSQuantity]:
-    data = {}
+    data: dict[str, VSSUnit | VSSQuantity] = {}
     for file in files:
         content = yaml.safe_load(file.read_text())
         if not content:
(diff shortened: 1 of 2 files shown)


SebastianSchildt:
Seems like a nice quality of life improvement 

In this specific example the quesiton (as it should always be for overlay stuff) whether this should be a general change in the standard catalogue, but that is a discussion is for another place(issue)

sschleemilch:
> Seems like a nice quality of life improvement
> 
> In this specific example the quesiton (as it should always be for overlay stuff) whether this should be a general change in the standard catalogue, but that is a discussion is for another place(issue)

Agreed. Personally I would like to express unix-time as a number but yeah, I am facing that issue actually right now

erikbosch (approved):
LGTM

erikbosch:
MoM:

- Merge

PR #453, opened 2025-06-12
Newest stats scripts

Giaccomole:
Replaces #448


sschleemilch:
I was wondering whether we can just add `--stats-*` arguments to normal generators instead of introducing new commands and then still call the required other commands. Probably easier to do both in one go if requested. Then we would not even need to load the required file but have it already in RAM.

PR #457, opened 2025-07-08
Even newer stat scripts

Giaccomole:
> I was wondering whether we can just add `--stats-*` arguments to normal generators instead of introducing new commands and then still call the required other commands. Probably easier to do both in one go if requested. Then we would not even need to load the required file but have it already in RAM.

Now i carried the stat generation to stats_utils.py and created --stats-sankey --stats-radial and --stats-piechart to the CLI options.

### DIFF
--- a/src/vss_tools/exporters/csv.py
+++ b/src/vss_tools/exporters/csv.py
@@ -12,11 +12,13 @@ import csv
 from pathlib import Path
 from typing import Any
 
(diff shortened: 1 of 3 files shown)


erikbosch:
MoM:

- Please review

erikbosch:
MoM:

- Erik to check if any remaining work
- If not ok to merge

PR #459, opened 2025-07-29
fix: _post_attach needs to update fqns recursively

sschleemilch:
The `_post_attach()` method changed the fqn for the node being attached to a different parent.
However, it also means that all fqns of all children need to be updated as well


erikbosch (approved):
LGTM

I will ask at the VSS meeting tonight  if anyone else  wants to review it

erikbosch:
MoM: Merge

Issue #465, opened 2025-09-18
BUG: Found problem with Version handling in VSS to samm exporter


Kostadin-Ivanov:
After upgrading my local VSS latest 5.1 and tried the **vss-tools: samm exporter** to generate new set of VSS 5.1 Aspect Models I found that **instead of** getting a NEW version: **5.1** under my dedicated aspect models namespace, **the existing 5.0** aspect models were updated and with wrong - 5.0 version namespace.

**For Example:**
**Before conversion** of VSS 5.1 to SAMM models, my current folder structure with aspect models looked like:

`
com.covesa.vss.spec/
   5.0/
`
**After call to VSS SAMM exporter**  the folder structure was as above.

**Expected**
After conversion of VSS 5.1 to SAMM, above folder structure should look like:
`
com.covesa.vss.spec/
   5.0/
   5.1/
`
where _5.1/_ holds the NEW VSS 5..1 aspect models.

Kostadin-Ivanov:
Hello @erikbosch , could you please assign this ticket to me?

I have a forked version of VSS-Tools and already found the problem, and I can provide a fix for it.

Thank you in advance :)

PR #466, opened 2025-09-18
FIX: #465: problem with Version handling in VSS to samm exporter

Kostadin-Ivanov:
After done tests, I found that the problem was in **__get_version_vss** function.
Reworked function so it will make sure that it starts with **default version: 1.0.0**, then if VSS :: VersionVSS is provided and differs from the 1.0.0 it will get updated as per the VersionVSS **default** Major.Minor.Patch values, if these are higher than 1.0.0 respectively.


erikbosch (approved):
LGTM

PR #468, opened 2025-10-03
Initial commit for adding ros2 exporter in vss-tools

ali-momin12:
- ros2interface.py (ros2 exporter core functionality)
- test-ros2-interface.py (custom test cases for ros2 exporter)
- test-units-ros2.yaml (custom units file for ros2 custom test case file)
- test-quantities-ros2.yaml (custom quantities file for ros2 custom test case file)
- ros2interface.md (ros2 exporter explanation guide)

Signed-off-by: Momin Ali <ali_momin@jp.honda>

### DIFF
--- a/docs/ros2interface.md
+++ b/docs/ros2interface.md
@@ -6,13 +6,25 @@ This exporter plugs into the `vspec export` CLI like other vss-tools exporters. 
 
 ## Generated Output Structure
 ```
-\<output><br>
-└── \<package-name><br>
-    ├── msg  # generated .msg definitions<br>
-            │ &emsp; └── \<MSG>.msg<br>
-            └── srv  # generated .srv (if setting is enabled)<br>
-                        ├── Get\<MSG>.srv<br>
-                        └── Set\<MSG>.srv<br>
+\<output>
+└── <package-name>
+    ├── msg  # generated .msg definitions
+    |   └── \<MSG>.msg
+    └── srv  # generated .srv (if setting is enabled)
+        ├── Get\<MSG>.srv
+        └── Set\<MSG>.srv
+```
+
+**Example Output**
+
+```
+OutputFolder
+└── Vss-Interface
+    ├── msg
+    |   └── VehicleSpeed.msg
+    └── srv
+        ├── GetVehicleSpeed.srv
+        └── SetVehicleSpeed.srv
 ```
 
 - .msg files include VSS metadata as comments (description, unit, min/max, allowed values).


docs/ros2interface.md
```diff
@@ -0,0 +1,114 @@
+
+# ROS 2 Interface Exporter (`ros2interface`)
+
+Export a VSS model to a ROS 2 interface package: generates `.msg` files (per leaf or aggregated by parent branch) and optional `.srv` files for Get/Set operations. This exporter plugs into the `vspec export` CLI like other vss-tools exporters. For generic exporter usage and common arguments, see the `vspec` documentation.
+
+
+## Generated Output Structure
+
+\<output><br>
```
sschleemilch, line 9: Probably should be wrapped into a code block

docs/ros2interface.md
```diff
@@ -0,0 +1,114 @@
+
+# ROS 2 Interface Exporter (`ros2interface`)
```
sschleemilch, line 2: Should we simplify the name to just `ros` and put what it does into the cli description of the sub command?

src/vss_tools/exporters/ros2interface.py
```diff
@@ -0,0 +1,634 @@
...
+    "uint32": "uint32",
+    "int64": "int64",
+    "uint64": "uint64",
+    "float": "float32",
+    "double": "float64",
+    "string": "string",
+}
+
+
+def to_snake(name: str) -> str:
```
sschleemilch, line 118: https://github.com/COVESA/vss-tools/blob/master/src/vss_tools/utils/misc.py contains already quite few converter functions. Maybe add it there. We can also think about just using https://pypi.org/project/case-converter/ instead

src/vss_tools/exporters/ros2interface.py
```diff
@@ -0,0 +1,634 @@
...
+    s = re.sub(r"_+", "_", s).strip("_")
+    return s.lower()
+
+
+def to_pascal(name: str) -> str:
+    parts = re.split(r"[^0-9a-zA-Z]+", name)
+    return "".join(p[:1].upper() + p[1:] for p in parts if p)
+
+
+def fqn_to_msg_basename(fqn: str) -> str:
```
sschleemilch, line 131: the tree `get_fqn()` has a Separator argument that could be used instead of converting it here: https://github.com/COVESA/vss-tools/blob/master/src/vss_tools/tree.py#L108

src/vss_tools/exporters/ros2interface.py
```diff
@@ -0,0 +1,634 @@
...
+    return to_pascal(fqn.replace(".", "_"))
+
+
+def field_name_from_tail(tail: str) -> str:
+    return to_snake(tail)
+
+
+def map_vss_to_ros_field(datatype: str, arraysize: Optional[int]) -> str:
+    base = datatype.strip()
+    is_var_array = base.endswith("[]")
```
sschleemilch, line 141: there is already a `is_array()` which pretty much does the same but probably should be used

src/vss_tools/exporters/ros2interface.py
```diff
@@ -0,0 +1,634 @@
...
+def fqn_to_msg_basename(fqn: str) -> str:
+    return to_pascal(fqn.replace(".", "_"))
+
+
+def field_name_from_tail(tail: str) -> str:
+    return to_snake(tail)
+
+
+def map_vss_to_ros_field(datatype: str, arraysize: Optional[int]) -> str:
+    base = datatype.strip()
```
sschleemilch, line 140: should not be necessary. Datatypes are checked against valid ones

src/vss_tools/exporters/ros2interface.py
```diff
@@ -0,0 +1,634 @@
...
+    is_var_array = base.endswith("[]")
+    base = base[:-2] if is_var_array else base
```
sschleemilch, line 142: can just do:

```python
base = datatype.rstrip("[]")
```

src/vss_tools/exporters/ros2interface.py
```diff
@@ -0,0 +1,634 @@
...
+        units=units_files,
+        types=types_files,
+        overlays=overlays_files,
+        expand=expand,
+    )
+    return root
+
+
+# ------------------------------ Naming & type mapping --------------------------------------
+ROS_PRIMITIVE_MAP: Dict[str, str] = {
```
sschleemilch, line 102: `Dict` syntax is deprecated, should use `dict`

src/vss_tools/exporters/ros2interface.py
```diff
@@ -0,0 +1,634 @@
...
+    if is_var_array:
+        return f"{ros_base}[]"
+    return ros_base
+
+
+# ------------------------------------- Tree traversal --------------------------------------
+LEAF_TYPES = {"sensor", "actuator", "attribute", "property"}
+
+
+def iter_leaves(root: VSSNode) -> Iterable[Tuple[VSSNode, object]]:
```
sschleemilch, line 157: I think you should use `PreOrderIter` https://anytree.readthedocs.io/en/latest/api/anytree.iterators.html or `findall` instead of implementing it on your own

src/vss_tools/exporters/ros2interface.py
```diff
@@ -0,0 +1,634 @@
...
+        children = getattr(node, "children", None) or []
+        if children:
+            stack.extend(children)
+        data = node.get_vss_data()
+        ntype = getattr(data, "type", None)
+        if ntype in LEAF_TYPES:
+            yield node, data
+
+
+def direct_parent_fqn(node: VSSNode) -> str:
```
sschleemilch, line 170:
```python
if node.parent:
  return node.get_fqn()
```

src/vss_tools/exporters/ros2interface.py
```diff
@@ -0,0 +1,634 @@
...
+#   - 'name:<value>'     -> tail name equality (or glob if contains * or ?)
+#   - implicit:
+#       contains * or ?  -> treat as glob
+#       contains '.'     -> treat as fqn
+#       otherwise        -> treat as name
+Rule = Callable[[str, str], bool]  # (tail, fqn) -> bool
+
+
+def _compile_rule(pattern: str, case_insensitive: bool) -> Rule:
+    kind: Optional[str] = None
```
sschleemilch, line 192:
```suggestion
    kind: str | None = None
```

src/vss_tools/exporters/ros2interface.py
```diff
@@ -0,0 +1,634 @@
...
+        p = pat_ci if case_insensitive else pat
+        return tt == p
+
+    return _n
+
+
+class TopicMatcher:
+    def __init__(
+        self,
+        include_patterns: Sequence[str],
```
sschleemilch, line 250: probably use `set` or `list`

src/vss_tools/exporters/ros2interface.py
```diff
@@ -0,0 +1,634 @@
...
+
+
+def render_msg_file(
+    fqn: str,
+    fields: List[Dict[str, str]],
+    header_comment: List[str],
+    enum_comment: Optional[List[str]] = None,
+) -> str:
+    buf: List[str] = []
+    buf.append("# Auto-generated from VSS by vss-tools (ros2msg exporter)")
```
sschleemilch, line 309: Would make that a global var since used more than once:

```python
GEN_HEADER = "# AUTO-GENERATED by VSS-TOOLS"
```

src/vss_tools/exporters/ros2interface.py
```diff
@@ -0,0 +1,634 @@
...
+    buf.append("---")
+    buf.extend(response_lines if response_lines else [])
+    buf.append("")
+    return "\n".join(buf)
+
+
+# ----------------------------- Build fields & msgs ------------------------------
+
+
+def build_field_from_leaf(leaf_node: VSSNode, data) -> Dict[str, str]:
```
sschleemilch, line 345: not sure why data is separated from the node and passed separately. Data is always part of the node itself

src/vss_tools/exporters/ros2interface.py
```diff
@@ -0,0 +1,634 @@
...
+    fqn = leaf_node.get_fqn()
+    tail = fqn.split(".")[-1]
```
sschleemilch, line 347: The tail is just `leaf_node.name`

src/vss_tools/exporters/ros2interface.py
```diff
@@ -0,0 +1,634 @@
...
+    outputs: List[Tuple[str, str, List[Dict[str, str]]]] = []
+    items = list(preselected) if preselected is not None else list(iter_leaves(root))
+
+    for node, data in items:
+        fqn = node.get_fqn()
+        field = build_field_from_leaf(node, data)
+        header_comment = [f"Signal: {fqn}"]
+        allowed_values = getattr(data, "allowed", None)
+        enum_comment = None
+        if isinstance(allowed_values, (list, tuple)) and allowed_values:
```
sschleemilch, line 426: is always a list of None

src/vss_tools/exporters/ros2interface.py
```diff
@@ -0,0 +1,634 @@
...
+@click.option(
+    "--output",
+    type=click.Path(dir_okay=True, path_type=Path, file_okay=False),
+    help="Output Directory.",
+    required=True,
+)
+@click.option(
+    "--package-name",
+    required=True,
+    default="vss_interfaces",
```
sschleemilch, line 486: maybe add `show_default=True`

src/vss_tools/exporters/ros2interface.py
```diff
@@ -0,0 +1,634 @@
...
+@clo.expand_opt
+@click.option(
+    "--output",
+    type=click.Path(dir_okay=True, path_type=Path, file_okay=False),
+    help="Output Directory.",
+    required=True,
+)
+@click.option(
+    "--package-name",
+    required=True,
```
sschleemilch, line 485: does not make sense in combination with `default` I think

src/vss_tools/exporters/ros2interface.py
```diff
@@ -0,0 +1,634 @@
...
+@click.option(
+    "--mode",
+    type=click.Choice(["aggregate", "leaf"], case_sensitive=False),
+    default="aggregate",
+    show_default=True,
+    help="Aggregate leaves by direct parent branch, or generate one .msg per leaf.",
+)
+@click.option(
+    "--srv",
+    type=click.Choice(["none", "get", "set", "both"], case_sensitive=False),
```
sschleemilch, line 498: maybe none should just be the absence of that option instead of the string "none"?

src/vss_tools/exporters/ros2interface.py
```diff
@@ -0,0 +1,634 @@
...
+    ),
+)
+@click.option(
+    "--exclude-topics",
+    multiple=True,
+    help=("Exclude patterns (same syntax as --topics). " "Applied after includes; any match here is removed."),
+)
+@click.option(
+    "--topics-file",
+    type=click.Path(path_type=Path),
```
sschleemilch, line 525: dir_okay=False, exists=True

src/vss_tools/exporters/ros2interface.py
```diff
@@ -0,0 +1,634 @@
...
+    package_name: str,
+    mode: str,
+    srv: str,
+    srv_use_msg: bool,
+    topics: Tuple[str, ...],
+    exclude_topics: Tuple[str, ...],
+    topics_file: Optional[Path],
+    topics_case_insensitive: bool,
+):
+    # 1) Load VSS
```
sschleemilch, line 552: No additional value

src/vss_tools/exporters/ros2interface.py
ali-momin12: build the whole flow against the separate node and data objects. would require changes at many places, resulting in lots of re-work. is this critical enough to address?

src/vss_tools/exporters/ros2interface.py
ali-momin12: user may want to create custom interfaces name as per the needs of the project. (As i am facing in my current project, that i need custom interface names for each topic)

src/vss_tools/exporters/ros2interface.py
ali-momin12: what is "is always a list of None"?

src/vss_tools/exporters/ros2interface.py
sschleemilch: Well, that comment does not provide any additional value. The same thing is stated right below of the comment in the info log statement as well as in the function name that loads the tree

src/vss_tools/exporters/ros2interface.py
```diff
@@ -0,0 +1,634 @@
...
+    else:
+        msgs = generate_msgs_leaf(root, preselected=preselected)
+        msgs = generate_msgs_aggregate(root, preselected=preselected)  # list[(fname, content, fields)]
+
+    # 4) Prepare output dirs
+    pkg_dir = output / package_name
+    msg_dir = pkg_dir / "msg"
+    srv_dir = pkg_dir / "srv"
+    msg_dir.mkdir(parents=True, exist_ok=True)
+    if srv.lower() != "none":
```
sschleemilch, line 605: here we can see how removing `none` from the choices would make it  a bit cleaner:

```python
if srv:
    srv_dir.mkdir(parents=True, exist_ok=True)
```

PR #470, opened 2025-10-15
:sparkles: added strict exceptions config file

sschleemilch:
# About

Idea is to be able to enable `--strict` and/or `--aborts` e.g. in CI but allowing current existing violations.

Added `--strict-exceptions` that should point to a YAML file defining exceptions in the following way:

```yaml
Vehicle.Foo: # Allow all violations

Vehicle.Bar: # Allow name violoations
  - name-style

Vehicle.Zed: # Allow unknown attribute violations
  - unknown-attribute

Vehicle.Yada: # Allow specific violations
  - name-style
  - unknown-attribute
```

Refactored available strict options into an enum and using it.

### DIFF
--- a/tests/vspec/test_strict/test_strict_e2e.py
+++ b/tests/vspec/test_strict/test_strict_e2e.py
@@ -49,3 +49,13 @@ def test_strict_error(vspec_file: str, tmp_path):
     cmd = f"vspec export json --pretty --strict -u {TEST_UNITS} -q {TEST_QUANT} --vspec {spec} --output {output}"
     process = subprocess.run(cmd.split())
     assert process.returncode != 0
+
+
+@pytest.mark.parametrize("vspec_file", ["faulty_case.vspec", "faulty_name_boolean.vspec"])
+def test_strict_exceptions(vspec_file: str, tmp_path):
+    spec = HERE / vspec_file
+    output = tmp_path / "out.json"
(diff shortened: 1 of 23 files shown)


sschleemilch:
I was wondering: should we have that really in every exporter? How about a `check` sub command  that does only validate whether the given specs are valid and nothing else?

tests/vspec/test_strict/test_strict_e2e.py
erikbosch: Do you intend to add any test with the new attribute here?

sschleemilch:
> > I was wondering: should we have that really in every exporter? How about a `check` sub command that does only validate whether the given specs are valid and nothing else?
> 
> You mean that you need to do things in two steps, like `vspec check` followed by `vspec export`? In general we need to do some form of checks to be able to export, but the "strict" part could as I see it be a separate command, but then maybe not call it `check` as we need to do basic checks for all commands.

Yeah, when implementing I was wondering whether it is worth doing that for each exporter when the use case of CI could be just to run `vspec validate` and later `vspec export`. I can find arguments for both. Alright, let's stick with this or a separate command? Your opinion?

tests/vspec/test_strict/test_strict_e2e.py
sschleemilch: Added

PR #475, opened 2025-12-18
Refactor/s2dm exporter

jdacoello:
This PR does some rework that enhances the way vspec is exported to fit the s2dm modeling approach based on GraphQL SDL. It includes aspects like:

- Handling corner cases
- Using extended attributes
- Annotation of mappings
- Improvements to the way how vspec is processed with pandas DataFrame
- ...


erikbosch:
MoM:

- D: Likely ready for review next week

PR #478, opened 2026-01-09
Fixes #477 : Binary exporter not able to parse allowed values as Int/Float

### LINKED ISSUE #477: Binary exporter not able to parse allowed values as Int
**Issue description:** 
```
x.y.z:
  allowed: [0,1,2,3]
  datatype: uint8
  description: signal returning states defined as integers: 0 = state zero, 1 = state one , 2 = state two, 3 = state three
  type: sensor
```

When running the vss-tools to export the tree in binary format, the exporter fails to get the length of the allowed list elements, as python len() func expect a string data type and not int 

```
  File "vss_tools/exporters/binary.py", line 93, in export_node
    f.write(create_l16v_string(allowedString(data.get("allowed", ""))))
  File "vss_tools/exporters/binary.py", line 43, in allowedString
    allowedStr += hexAllowedLen(elem) + elem
  File "vss_tools/exporters/binary.py", line 48, in hexAllowedLen
    hexDigit1 = len(allowed) // 16
TypeError: object of type 'int' has no len()
```

**Proposed solution:** 
Stringify the element in the allowed list before calculating the length of the elem: 

```
def allowedString(allowedList):
    allowedStr = ""
    for elem in allowedList:
        allowedStr += hexAllowedLen(str(elem)) + str(elem)
    return allowedStr
```

abdelkadersellami:


erikbosch (approved):
LGTM

PR #502, opened 2026-04-10
feat(protobuf): generate nested enum for string fields with allowed values

aki1770-del:
Closes #493

## Problem

When a VSS signal has `datatype: string` and an `allowed` attribute, the protobuf exporter emits a plain `string` field. This silently accepts any value at the protobuf layer — the constraint declared in the VSS spec is invisible to protobuf consumers.

```python
# Before — allowed values ['DRY', 'WET', 'SNOW', 'ICE'] are lost
string Surface = 2;
```

## Fix

`print_messages()` now uses a two-pass approach in `src/vss_tools/exporters/protobuf.py`:

- **Pass 1**: for each `string` + `allowed` field, emit a nested `enum <FieldName>Enum` with allowed values assigned indices starting at 0 (satisfying the proto3 zero-value requirement).
- **Pass 2**: emit the field using the enum type instead of `string`.

`repeated string[]` fields with `allowed` become `repeated <EnumType>`. String fields without `allowed` and all non-string fields are unchanged.

```proto
# After
message A {
  enum SurfaceEnum {
    DRY = 0;
    WET = 1;
    SNOW = 2;
    ICE = 3;
  }
  repeated SurfaceEnum Surface = 2;
}
```

**Note on proto3 zero-value semantics**: proto3 uses the first enum entry as the default for unset fields. This implementation assigns index 0 to the first entry in the VSS `allowed` list. For signals that follow the convention of listing `UNKNOWN` first (e.g., `Vehicle.Exterior.RoadSurfaceCondition`), this aligns naturally. For signals without a natural "unset" first value, the first allowed entry becomes the proto default. Happy to discuss if a different index assignment strategy is preferred.

## Changes

| File | Change |
|------|--------|
| `src/vss_tools/exporters/protobuf.py` | +33 lines — `_enum_type_name()`, `_write_nested_enum()`, two-pass logic in `print_messages()` |
| `tests/vspec/test_protobuf_enum_allowed/` | New test: scalar, repeated, and unchanged-float cases |
| `tests/vspec/test_protobuf_comments/expected_*.proto` | Updated expected output to reflect enum generation for `StringWithAllowed` fixture |

## Tests

```
$ python3 -m pytest tests/ -k "proto" -v
13 passed in 5.36s
```

All 13 protobuf-related tests pass, including the two existing `test_protobuf_comments` parametrized cases (no-comments / with-comments) which already contained a `string` + `allowed` field in their fixture.

*AI-assisted — authored with Claude, reviewed by Komada.*

### DIFF
--- a/src/vss_tools/exporters/protobuf.py
+++ b/src/vss_tools/exporters/protobuf.py
@@ -25,6 +25,7 @@ from vss_tools.model import (
     VSSDataDatatype,
     VSSDataStruct,
 )
+from vss_tools.datatypes import Datatypes
 from vss_tools.tree import VSSNode
 
 PATH_DELIMITER = "."
@@ -183,21 +184,13 @@ def _write_nested_enum(fd: TextIOWrapper, field_name: str, allowed: list, indent
 def print_messages(
     nodes: tuple[VSSNode], fd: TextIOWrapper, static_uid: bool, add_optional: bool, include_comments: bool
 ):
-    # Pass 1: write nested enum definitions for every string field with allowed values.
-    for node in nodes:
-        if isinstance(node.data, VSSDataDatatype):
-            base = node.data.datatype.strip("[]")
-            if base == "string" and node.data.allowed:
-                _write_nested_enum(fd, node.name, node.data.allowed)
-
-    # Pass 2: write field declarations.
     usedKeys: dict[int, str] = {}
     for i, node in enumerate(nodes, 1):
         if isinstance(node.data, VSSDataDatatype):
             dt_val = node.data.datatype
             base = dt_val.strip("[]")
-            if base == "string" and node.data.allowed:
-                # Replace plain string with the nested enum type.
+            if base == Datatypes.STRING[0] and node.data.allowed:
+                _write_nested_enum(fd, node.name, node.data.allowed)
                 data_type = _enum_type_name(node.name)
             else:
                 data_type = mapped.get(base, base)
--- a/src/vss_tools/exporters/protobuf.py
+++ b/src/vss_tools/exporters/protobuf.py
@@ -162,14 +162,45 @@ def write_comment(fd: TextIOWrapper, node: VSSNode, indent: str = "  "):
             fd.write(f"{indent}// {line}\n")
 
 
+def _enum_type_name(field_name: str) -> str:
+    """Return the nested enum type name for a string field with allowed values."""
+    return f"{field_name}Enum"
+
+
+def _write_nested_enum(fd: TextIOWrapper, field_name: str, allowed: list, indent: str = "  ") -> None:
+    """Write a nested proto3 enum for a string field constrained to allowed values.
+
+    Proto3 requires the first enum value to equal 0.  We assign values in the
+    order they appear in the VSS `allowed` list, starting from 0.
+    """
+    enum_name = _enum_type_name(field_name)
+    fd.write(f"{indent}enum {enum_name} {{\n")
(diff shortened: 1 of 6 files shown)


src/vss_tools/exporters/protobuf.py
```diff
@@ -162,14 +162,45 @@ def write_comment(fd: TextIOWrapper, node: VSSNode, indent: str = "  "):
...
+    for idx, val in enumerate(allowed):
+        fd.write(f"{indent}  {val} = {idx};\n")
+    fd.write(f"{indent}}}\n")
+
+
 def print_messages(
     nodes: tuple[VSSNode], fd: TextIOWrapper, static_uid: bool, add_optional: bool, include_comments: bool
 ):
+    # Pass 1: write nested enum definitions for every string field with allowed values.
+    for node in nodes:
```
sschleemilch: Why do we need to iterate twice?

src/vss_tools/exporters/protobuf.py
```diff
@@ -162,14 +162,45 @@ def write_comment(fd: TextIOWrapper, node: VSSNode, indent: str = "  "):
...
+        if isinstance(node.data, VSSDataDatatype):
+            base = node.data.datatype.strip("[]")
+            if base == "string" and node.data.allowed:
```
sschleemilch: "string" literal will probably never change. Still use the the one from `datatypes.Datatypes`

sschleemilch (changes requested):

src/vss_tools/exporters/protobuf.py
```diff
@@ -162,14 +163,37 @@ def write_comment(fd: TextIOWrapper, node: VSSNode, indent: str = "  "):
...
 
+def _enum_type_name(field_name: str) -> str:
+    """Return the nested enum type name for a string field with allowed values."""
+    return f"{field_name}Enum"
+
+
+def _write_nested_enum(fd: TextIOWrapper, field_name: str, allowed: list, indent: str = "  ") -> None:
+    """Write a nested proto3 enum for a string field constrained to allowed values.
+
+    Proto3 requires the first enum value to equal 0.  We assign values in the
```
erikbosch, line 174: VSS supports the use of `default` for allowed values, see e.g. https://github.com/COVESA/vehicle_signal_specification/blob/b6e55988161ca67b20a3cdc3bdc1953c2faae4df/spec/Powertrain/CombustionEngine.vspec#L50

VSS in itself does not give the order for `allowed` any meaning. I took a quick look at the VSS catalog and in the few cases where we use both `allowed` and `default` the default value is the first item. But theoretically it could be something else and the question is then if the string specified as `default` better should be placed first, or if that just would make things more complex and error-prone.

Not saying that anythings needs to be changed in this PR, it was just a reflection I made.

sschleemilch (approved):

aki1770-del:
Thanks @erikbosch — sharp catch on the proto3 zero-default ↔ VSS-default alignment. You're right that ordering `allowed` with `default` first would make the proto-implicit-default carry semantic meaning instead of being "whatever happened to be first in the list."

To respect the scope of this PR (which you and @sschleemilch have already reviewed), I've filed it as a follow-up: #505 — happy to take it as a separate PR if the direction is confirmed there.

*AI-assisted — authored with Claude, reviewed by Komada.*

erikbosch:
MoM:

- Please review

erikbosch:
MoM:

- ok to merge after approval from @erikbosch

erikbosch (approved):
LGTM

Issue #508, opened 2026-04-17
Stable identifier governance for VSS-based modeling projects


jdacoello:
# Problem
VSS modeling projects face two compounding identification problems:

## 1. FQN-as-identifier is fragile and implicit

In VSPEC projects, the Fully Qualified Name (`FQN`) (e.g., `Vehicle.Cabin.Seat.Row1.DriverSide.Position`) is almost always used as the unique identifier for VSS nodes — in databases, APIs, tooling integrations, and data pipelines. This creates a structural fragility: the identifier is coupled to the tree structure of the model.

When an organization takes the public COVESA VSS as a basis for an internal project and begins overlaying, the ownership question becomes immediately ambiguous. 
* Is `Vehicle.Cabin.MyFeature` a COVESA concept or an internal one? 
* If the overlay modified the relevant metadata of a node, then its semantics have changed in an incompatible way. 
* Who governs its lifecycle? 
* If COVESA restructures their tree in a future release, does the internal `FQN` break? 
* Renaming or moving a node becomes a painful process with breaking consequences for every downstream consumer that stored the string.

## 2. The existing `export id` approach partially solves it but leaves governance implicit

vss-tools introduced an id exporter that hashes each node together with its associated metadata, producing a deterministic fingerprint per node. If relevant metadata changes, the hash changes and a new ID is produced. The concept of tracking changes was handled via the "formerly known as" (fka) mechanism.

This is a meaningful step forward, but it has limits. The IDs are entirely derived from content — there is no explicit declaration of who minted the ID, under what authority, or for which namespace. When a private project reuses a public specification, the ownership and governance of the ID space remains implicit and ambiguous. A "stable" identifier that changes whenever metadata changes is, in practice, not stable enough for long-term external references. 

Suggestion
Borrowing from basic Semantic Web principles: assign identifiers sequentially and bind them to an explicitly declared URI namespace with a preferred prefix for compactness (CURIE form).

A project declares ownership once:
```yaml
namespace:
  prefix: vss
  uri: "https://www.covesa.global/model/vss#"
```

Nodes are assigned IDs in the order they first appear, starting from 0, per namespace:
```
vss:0  →  Vehicle
vss:1  →  Vehicle.ADAS
vss:2  →  Vehicle.Cabin
```
A private project that overlays COVESA's model declares COVESA as an import and mints only its own new concepts:
```yaml
namespace:
  prefix: myns
  uri: "https://www.example.org/myModel#"
imports:
  vss:
[...]

PR #509, opened 2026-04-17
feat(registry): implement stable identifier registry

jdacoello:
# Context
This PR addresses the issue #508 . 

It introduces a new `vspec registry` command group that assigns stable, persistent, namespace-scoped identifiers to VSS nodes.

## What this adds
New top-level command group: `vspec registry`. It is not meant to be an exporter, but a governance tool. 

Two subcommands:

- `vspec registry sync` — reads the current vspec, mints IDs for any FQNs not yet in the registry, writes the updated CSV. Safe to run repeatedly; existing rows are never modified or deleted.
- `vspec registry validate` — checks schema and row-level hash integrity without writing anything. Designed for pre-commit hooks and CI gates.

## Working file format (two files):
* `namespaces.yaml` — declares the one namespace this project owns (prefix + URI) and any external namespaces it imports. Human-maintained, rarely changes.
* `registry.csv` — append-only ledger of `composed_id → fqn` mappings. Tool-managed, committed to version control. Each row carries a SHA-256 hash of its immutable fields to detect manual edits.

## Optional JSON-LD sidecar with flag `--export-jsonld`:

A derived artifact regenerated on every sync. Provides a `@context` + `@graph` representation consumable by any linked data toolchain without changing the working format.

## Key design decisions
- One owned namespace per project — enforced at load time. A project cannot accidentally mint into an imported namespace.
- IDs are sequential per prefix, starting from 0 — simple, human-readable, independent per namespace.
- IDs are never deleted — when a node disappears from the vspec, its row remains and its `status` transitions to deprecated. Downstream consumers are never left with a broken reference.
- Hash-based tamper detection — verifies every row's hash on read. Any manual edit is caught before the tool proceeds.
- Prefix in `namespaces.yaml`, not a CLI flag — cannot be overridden by accident in a script.

### DIFF
--- a/docs/registry.md
+++ b/docs/registry.md
@@ -1,0 +1,249 @@
+# Spec Registry
+
+## Why this exists
+
+When a VSS-based model grows and gets consumed by multiple tools, teams, and systems, a common problem emerges: how do you refer to a concept — a signal, a branch, a struct — in a way that is stable across time and across different modeling projects?
+
+The fully qualified name (FQN) like `Vehicle.Cabin.Seat.Row1.DriverSide.Position` is human-readable and useful during development, but it is fragile as a long-term identifier. If a node gets renamed, restructured, or merged into another branch, any system that stored the FQN as a reference now has a broken link. There is no way to know that `Vehicle.Cabin.Seat.Row1.DriverSide.Position` and its successor `Vehicle.Cabin.Seat.Row1.Left.Position` refer to the same real-world concept.
+
+The spec registry solves this by assigning a **stable, short numeric ID** to every node the first time it appears in the model. That ID never changes, even if the node is renamed or restructured. Systems downstream can store `covss:142` instead of the full FQN and always resolve it back to the current name via the registry.
+
+This pattern is widely used in ontology engineering (RDF/OWL), standards bodies (ISO, W3C), and linked data systems. The registry tool brings the same discipline to VSS-based modeling projects.
+
+---
+
+## Important note: IDs are not deterministic
+
+The IDs assigned by this tool depend on the **order in which nodes first appear** in the registry and the **current state of the modeling project at the time of the first sync**. Two projects that independently sync the same vspec will produce different numeric IDs for the same FQNs if they use different namespace prefixes or run the sync at different points in the model's history.
+
+This is by design. Each project owns its own namespace and is responsible for when it first mints IDs. The IDs are only meaningful within a given namespace.
+
+As a direct consequence, **the registry files must be treated as governed artifacts**:
+
+- They should be committed to version control and never edited manually.
+- Write access in CI should be restricted to tagged release pipelines only.
+- The `vspec registry validate` command should run as a pre-commit hook and in every CI pipeline to catch any corruption early.
+
+If the registry is corrupted or lost, previously published IDs cannot be reconstructed automatically. Treat the `registry.csv` file with the same care as a database migration history or a package lock file.
+
+---
+
+## Concepts
+
+### Namespace
+
+A namespace is a URI that scopes a set of IDs. Each project declares exactly one namespace it **owns** — the namespace under which it mints new IDs. It may also declare additional namespaces it **imports** — external namespaces it references but does not mint into.
+
+### Prefix
+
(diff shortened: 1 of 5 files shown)


docs/registry.md
```diff
@@ -0,0 +1,249 @@
...
+covss:1,Vehicle.Cabin,1,active,a87ff679a2f3e71d9181...
+covss:2,Vehicle.Cabin.Seat,2,active,1679091c5a880faf6fb...
+```
+
+Column meanings:
+- `composed_id` — the stable identifier, e.g. `covss:0`
+- `fqn` — the fully qualified node name at the time the ID was minted
+- `int_id` — the numeric part of `composed_id`, per-prefix counter
+- `status` — `active` or `deprecated`. If a concept is not present in a future release of the model, then it is because it was deleted from the model. However, the id is persisted forever in the registry.
+- `row_hash` — SHA-256 of `"{composed_id}|{fqn}"`, used to detect tampering
```
sschleemilch, line 91: Not sure about the benefit of having a row sha. What does it protect against? Someone modifying the row by accident? But since it's commited you would see that anyway?

If we go for sha, I would probably vote for a hash chain to also fix the ordering and not allow inserting something. Meaning the sha is the sha of the row before + hashing the current row

docs/registry.md
```diff
@@ -0,0 +1,249 @@
...
+
+**Step 2**: The project creates its own `namespaces.yaml`:
+
+```yaml
+namespace:
+  prefix: myns
+  uri: "https://www.example.org/myModel#"
+  description: "Private VSS extension"
+
+imports:
```
sschleemilch, line 200: Could we maybe also just add more then one `namespaces.yaml` and therefore we don't need `imports`?

jdacoello: The idea of that separation is that the tool creates ids under the provided `namespace` only. Otherwise, the function that creates the identity would always need to receive the namesake as parameter. The declaration of `imports` serve as a validation of the registry. Such that only declared namespaces are in use.

docs/registry.md
jdacoello: Thanks for the observation. I later realized that using `sha` here is of little value because I actually want to remove the dependency of `fqn`. So, hashing the `fqn` contradicts that. The initial idea of hashing was to be able to detect changes easily. But, now I am re working this idea and new commits will refactor it.

erikbosch:
Is the intention to store the registry (for VSS) in the vss repo? If so we need to come up with a working model. I assume, as long as we just work on master, that generating a new csv file as part of release preparations would be easy. One could theoretically do it also for every merge to master, but that is possibly not needed and it is also so that we sometimes on master may roll back or fix unreleased changes.

jdacoello:
> Is the intention to store the registry (for VSS) in the vss repo? If so we need to come up with a working model. I assume, as long as we just work on master, that generating a new csv file as part of release preparations would be easy. One could theoretically do it also for every merge to master, but that is possibly not needed and it is also so that we sometimes on master may roll back or fix unreleased changes.

The intention is to release the registry as one more artifact in every release. So, it becomes an immutable reference of identity for the concepts. We can create the registry on demand and also uploaded it to previous releases (at least from V5 on). For sub sequent use, the previous released registry can be used as an input for the update.

jdacoello:
Update:

I am elaborating this idea of the registry further. This is the plan:

- [x] Implement a `snapshot` exporter that composes all the modular `vspec` files and creates a unique stand alone sorted `vspec`. --> Useful to identify differences between releases
- [x] Implement a `diff` function/command that compares two stand-alone `vspec` files and reports the differences with a pre defined structure. --> Useful to assign the proper semantic version tag
- [ ] Split the registry idea into two: 
  - [ ] `concepts_registry` --> To persist the future-proof identity of the concepts and their meaning
  - [ ] `variants_registry` --> To persist the realization of the concepts with specific metadata

erikbosch:
MoM:

- Daniel working on it

erikbosch:
MoM:

- D: Will need to manage registry as stand-alone packages
- D: Working on it
- D: Will take 10 minutes of the call when PR is ready

jdacoello:
The registry has been decomposed into smaller parts:
* `compose` and `diff` commands --> vss-tools #521
* Spec registry tables manipulation will be implemented outside vss-tools and then imported as dependency (work in progress)

Thus, this PR is closed since the initial solution was too simple for the things we want to achieve with the identity management.

PR #512, opened 2026-04-23
feat(strict): add DEFAULT_MATCHES_FIRST_ALLOWED check for default==allowed[0]

### LINKED ISSUE #507: Validate default == allowed[0] when field generates a nested enum (proto3)
Follow-up from #505 (closed in favor of this narrower scope) and #502.

> **Edit 2026-04-16:** Original proposal included a check (a) "default ∈ allowed." @sschleemilch pointed out this is already enforced in pydantic `VSSDataDatatype` (see [tests/test_model.py L39](https://github.com/COVESA/vss-tools/blob/master/tests/test_model.py#L39)). Removing (a) from scope; the actual new work is (b) below.

## Context

PR #502 generates proto3 nested enums for string fields with `allowed` values, using `allowed` list order (first entry → enum value 0). Proto3 has no mechanism to set a non-zero default for strings or ints, so the first `allowed` entry is the implicit wire-unset value. VSS `default`, by contrast, is an explicit initial-value assertion — a different concept.

Per @erikbosch in #505, the cleanest reconciliation is a **validation** rule, not a reordering transformation.

## Proposal (narrowed)

- **(b)** If a string field has both `allowed` and `default`, require `default == allowed[0]`. Per @erikbosch, this belongs in `src/vss_tools/strict.py` as a new `StrictOption` (tentatively `DEFAULT_MATCHES_FIRST_ALLOWED`), enforced on the VSS standard catalog and relaxable via the existing exceptions file for any spec that intentionally diverges.
- **(c)** Clear error messages pointing to the canonical VSS pattern: `allowed: ['UNKNOWN', ...]` + `default: 'UNKNOWN'` (as in [CombustionEngine.Configuration](https://github.com/COVESA/vehicle_signal_specification/blob/b6e55988161ca67b20a3cdc3bdc1953c2faae4df/spec/Powertrain/CombustionEngine.vspec#L50)). Lands alongside (b).

## Expected impact on existing VSS specs

Minimal. Specs following the idiomatic `UNKNOWN`-first pattern already satisfy (b). Only specs that set `default` to a non-first `allowed` value would fail under strict mode — and such divergence was likely unintentional.

## References

- #502 (PR implementing nested enum generation)
- #505 (closed; reshuffling proposal withdrawn)
- Proto3 default semantics: https://protobuf.dev/programming-guides/proto3/#default
- Strict-check machinery: [`src/vss_tools/strict.py`](https://github.com/COVESA/vss-tools/blob/master/src/vss_tools/strict.py)

aki1770-del:
Closes #507.

Per @erikbosch (#507 comment 2026-04-15): "I think check b (i.e. that default value matches first value) at least should be part of the 'strict' check we have". Per @sschleemilch: original proposal (a) "default ∈ allowed" is already enforced in pydantic `VSSDataDatatype` (see `tests/test_model.py#L39`), so this PR lands only the narrowed (b) check.

## Why this matters

Proto3 nested-enum generation from `allowed` (landed in #502) uses `allowed[0]` as the implicit wire-unset default value — proto3 has no mechanism to set a non-zero default for strings or ints. If VSS `default` does not match `allowed[0]`, the serialized-and-deserialized value silently becomes `allowed[0]`, diverging from the VSS author's declared intent.

The canonical VSS pattern is:

```yaml
Vehicle.Cabin.Mode:
  datatype: string
  allowed: ['UNKNOWN', 'ECO', 'SPORT']
  default: 'UNKNOWN'
```

See `Powertrain/CombustionEngine.vspec` `Configuration` for an in-tree example (referenced in #507 body).

## What this PR does

Adds a new strict-check option, parallel to the existing `NAME_STYLE` and `UNKNOWN_ATTRIBUTE` machinery:

- `StrictOption.DEFAULT_MATCHES_FIRST_ALLOWED` enum member
- `StrictExceptions.defaults` per-FQN exemption set
- `load_strict_exceptions` dispatch for the new option (including the null-options fallback that exempts an FQN from all strict checks)
- `VSSNode.get_default_first_allowed_violations()` tree walker
- `check_default_first_allowed_violations()` enforcement in `get_root()`
- `DefaultFirstAllowedException` signal
- `--aborts` CLI choice extended with `default-matches-first-allowed`

Gated behind `--strict` or `--aborts default-matches-first-allowed`. Existing runs without either flag are unchanged — no behavior regression.

## Error message

When a violation is detected (under strict/abort mode), the warning is:

```
default != allowed[0]: 'Vehicle.Cabin.Mode' (default='ECO' != allowed[0]='UNKNOWN').
Canonical pattern: allowed: ['UNKNOWN', ...] + default: 'UNKNOWN'.
```

## Testing

- 5 new tests (1 in `test_strict.py` for the new enum + dispatch; 4 in `test_vss_node.py` for the tree-walk violation collector covering clean / mismatch / no-allowed / no-default branches)
- Existing `test_strict.py` tests extended with `.defaults` set assertions for consistency with `.names` / `.attributes`
- Full suite: 132/132 pass on `tests/ --ignore=tests/vspec --ignore=tests/binary` (pre-existing `test_binary` failure verified unrelated — reproduces on clean `master`)
- `ruff check src/ tests/test_strict.py tests/test_vss_node.py` clean

## Chain context

Chain A: #502 → #507. PR #502 adds the proto3 nested-enum exporter that creates the default-vs-allowed divergence risk; this PR adds the strict validation that prevents authors from committing VSS specs with that silent-divergence.

### DIFF
--- a/src/vss_tools/cli_options.py
+++ b/src/vss_tools/cli_options.py
@@ -76,11 +76,13 @@ aborts_opt = option(
     "--aborts",
     "-a",
     multiple=True,
-    type=click.Choice([
-        StrictOption.NAME_STYLE.value,
-        StrictOption.UNKNOWN_ATTRIBUTE.value,
-        StrictOption.DEFAULT_MATCHES_FIRST_ALLOWED.value,
-    ]),
+    type=click.Choice(
+        [
+            StrictOption.NAME_STYLE.value,
+            StrictOption.UNKNOWN_ATTRIBUTE.value,
+            StrictOption.DEFAULT_MATCHES_FIRST_ALLOWED.value,
+        ]
+    ),
     help="Abort on selected option. The '--strict' option enables all of them.",
     show_choices=True,
 )
--- a/src/vss_tools/cli_options.py
+++ b/src/vss_tools/cli_options.py
@@ -76,7 +76,11 @@ aborts_opt = option(
     "--aborts",
     "-a",
     multiple=True,
-    type=click.Choice([StrictOption.NAME_STYLE.value, StrictOption.UNKNOWN_ATTRIBUTE.value]),
+    type=click.Choice([
+        StrictOption.NAME_STYLE.value,
+        StrictOption.UNKNOWN_ATTRIBUTE.value,
+        StrictOption.DEFAULT_MATCHES_FIRST_ALLOWED.value,
+    ]),
     help="Abort on selected option. The '--strict' option enables all of them.",
     show_choices=True,
 )
--- a/src/vss_tools/main.py
+++ b/src/vss_tools/main.py
@@ -116,9 +116,7 @@ def check_default_first_allowed_violations(
                     f"default != allowed[0]: '{violation[0]}' ({violation[1]}). "
                     "Canonical pattern: allowed: ['UNKNOWN', ...] + default: 'UNKNOWN'."
                 )
-            raise DefaultFirstAllowedException(
-                f"default != allowed[0] violations detected: {violations}"
-            )
+            raise DefaultFirstAllowedException(f"default != allowed[0] violations detected: {violations}")
 
 
 def get_types_root(types: tuple[Path, ...], include_dirs: list[Path]) -> VSSNode | None:
--- a/src/vss_tools/main.py
+++ b/src/vss_tools/main.py
@@ -36,6 +36,10 @@ class ExtraAttributesException(Exception):
     pass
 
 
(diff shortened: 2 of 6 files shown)


sschleemilch (approved):

aki1770-del:
Thanks, Sebastian. Should I land this myself, or are you handling the merge?

erikbosch:
MoM:

- Ok to merge after approval from @erikbosch

erikbosch (approved):
LGTM

PR #514, opened 2026-05-05
Remove Franca Exporter

### LINKED ISSUE #511: Deprecation of Franca exporter?
We have some old issues on the Franca exporter, and I have not really seen any signs of usage of the Franca exporter. I would like to know if there is someone using it and if so possibly willing to invest some time in it. If not I would suggest that we retire/remove it

https://github.com/COVESA/vss-tools/blob/master/src/vss_tools/exporters/franca.py

https://github.com/COVESA/vehicle_signal_specification/issues/735
#239 
#336

erikbosch:
Fixes #511

We have quite a few limitations in the Franca exporter, we have not seen any users of it for a long time.

**If you use the Franca exporter or want it to be kept and you can volunteer as maintainer for it, please let us know!**

_I suggest this PR to be open for at least a month to make sure that no-one are about to miss that we intend to remove the exporter._

Related: #239


erikbosch:
MoM:

- PR presented

erikbosch:
MoM:

- Wait with merge until 6.1 released

PR #515, opened 2026-05-05
ddsidl: fix dead Python-keyword check in getAllowedName

SoundMatt:
## Problem

`getAllowedName` in `src/vss_tools/exporters/ddsidl.py` is meant to
prefix reserved-word identifiers (C, IDL, or Python) with a leading
underscore so the generated IDL is valid. The Python branch of the
check was dead:

```python
if name.lower() in c_keywords or name.lower() in idl_keywords or keyword.iskeyword(name.lower):
                                                                                   ^^^^^^^^^^
                                                                            missing call parens
```

`name.lower` (without `()`) is the bound-method object, not the
lowered string. `keyword.iskeyword(<method object>)` always returns
False, so VSS signal names that are Python keywords but NOT also
C/IDL keywords slipped through silently. Examples that bypass the
check: `class`, `def`, `import`, `lambda`, `try`, `with`, `yield`,
`async`, `await`.

The downstream effect is that the IDL exporter emits these names
unprefixed, producing syntactically invalid IDL.

## Fix

Extract `name.lower()` to a local once and pass it explicitly to
`keyword.iskeyword`:

```python
def getAllowedName(name):
    lower = name.lower()
    if lower in c_keywords or lower in idl_keywords or keyword.iskeyword(lower):
        return "_" + name
    else:
        return name
```

Behaviour-preserving for all inputs the current code handles
correctly; corrects the dead Python-keyword arm.

## Test

Adds `tests/test_ddsidl_naming.py` with a regression test that
asserts `getAllowedName("class") == "_class"` (and similar for
`def`, `import`, `lambda`). With the original bug, that test fails.
Also covers normal names, IDL keywords, and C keywords to lock in
current behaviour.

## Notes

- Only `getAllowedName` and the new test file are touched. No other
  call sites or behaviour changed.
- The redundant `name.lower()` call (three times in the original) is
  now called once, which is incidentally cleaner but the primary
  motivation is correctness.

### DIFF
--- a/src/vss_tools/exporters/ddsidl.py
+++ b/src/vss_tools/exporters/ddsidl.py
@@ -147,13 +147,6 @@ idl_keywords = [
 
 def getAllowedName(name):
     lower = name.lower()
-    # `keyword.iskeyword(name.lower)` (without call parens) passed a
-    # bound-method object to iskeyword instead of the lowered string,
-    # so iskeyword always returned False. VSS signals named with Python
-    # keywords that are NOT also C/IDL keywords (e.g. 'class', 'def',
-    # 'import', 'lambda') were therefore emitted unprefixed and produced
-    # syntactically invalid IDL. Compute lower once and pass it
-    # explicitly to iskeyword.
     if lower in c_keywords or lower in idl_keywords or keyword.iskeyword(lower):
         return "_" + name
     else:
--- a/src/vss_tools/exporters/ddsidl.py
+++ b/src/vss_tools/exporters/ddsidl.py
@@ -146,7 +146,15 @@ idl_keywords = [
 
 
 def getAllowedName(name):
-    if name.lower() in c_keywords or name.lower() in idl_keywords or keyword.iskeyword(name.lower):
+    lower = name.lower()
+    # `keyword.iskeyword(name.lower)` (without call parens) passed a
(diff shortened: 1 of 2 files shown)


erikbosch (approved):
LGTM

erikbosch:
Thanks for the PR. I think it looks good but will leave it open for a while to give others a chance to review/comment if they would like to.

src/vss_tools/exporters/ddsidl.py
```diff
@@ -146,7 +146,15 @@
 
 
 def getAllowedName(name):
-    if name.lower() in c_keywords or name.lower() in idl_keywords or keyword.iskeyword(name.lower):
+    lower = name.lower()
+    # `keyword.iskeyword(name.lower)` (without call parens) passed a
```
sschleemilch: remove the verbose explanation of the past

erikbosch:
MoM:

- Presented at meeting
- ok to merge after maintainer approval

erikbosch:
@SoundMatt - will you take a look at the comments from Sebastian

SoundMatt:
Addressed — verbose comments removed.

erikbosch:
@sschleemilch - do you want to take  second look

PR #516, opened 2026-05-05
model: replace assert-based validation with raise ValueError (works under python -O)

SoundMatt:
## Problem

The validators in `src/vss_tools/model.py` use `assert COND, "msg"`
to enforce ~32 invariants on `VSSData`, `VSSDataDatatype`, and
`VSSUnit` fields. Pydantic's own documentation explicitly warns
against this pattern: Python's `-O` optimisation flag disables
asserts, allowing invalid data to pass through.

Anyone running vss-tools under `python -O` (some packaging
pipelines, performance-tuned environments) gets all VSS spec
validation silently bypassed. The parser accepts invalid specs as
valid, and downstream exporters then act on garbage data.

## Reproduction

```bash
python -O -c "
from vss_tools.model import VSSDataDatatype
# This *should* raise ValidationError ('foo' is not a datatype)
VSSDataDatatype(type='attribute', description='x', datatype='foo')
"
# Before this PR: silently succeeds (wrong).
# After this PR:  raises ValidationError (correct).
```

## Fix

Mechanically replace each:

```python
assert COND, "msg"
```

with:

```python
if not COND:
    raise ValueError("msg")
```

Pydantic wraps a `ValueError` raised inside a validator as a
`ValidationError` automatically, so externally-observable behaviour
is identical in normal mode. The new form is preserved under
`python -O`.

## Test

`tests/test_model.py` already parametrises 50+ invalid-input cases
that expect `pydantic.ValidationError`. All continue to pass after
this PR in normal mode. They would have *failed* under `python -O`
before this change; now they pass in both modes.

I considered adding an explicit `python -O` job to CI to catch
future regressions of this anti-pattern, but kept it out of this PR
to stay focused. Happy to follow up if maintainers want.

## Notes

- 32 `assert` statements replaced across `VSSData`, `VSSUnit`, and
  `VSSDataDatatype` validators.
- A few asserts had `not (X and Y)` form; those became `if X and Y`
  for natural inversion. Equivalent semantics, idiomatic Python.
- One pre-existing falsy-check bug (`if not self.default:` in
  `check_default_min_max`) is intentionally left untouched — it's
  a separate validation bug that deserves a focused PR (signal
  with `default=0, min=10` silently passes today).

### DIFF
--- a/src/vss_tools/model.py
+++ b/src/vss_tools/model.py
@@ -123,15 +123,18 @@ class VSSData(VSSRaw):
         if v is None:
             return v
         pattern = r"^0x[0-9A-Fa-f]{8}$"
-        assert bool(re.match(pattern, v)), f"'{v}' is not a valid 'constUID'"
+        if not re.match(pattern, v):
+            raise ValueError(f"'{v}' is not a valid 'constUID'")
         return v
 
     @model_validator(mode="after")
     def ensure_description(self) -> Self:
         """Give better explanation for empty description."""
-        assert (
-            self.description != ""
-        ), "All nodes in the final tree must have a description. Implicit branches are not allowed in final tree!"
+        if self.description == "":
+            raise ValueError(
+                "All nodes in the final tree must have a description. "
+                "Implicit branches are not allowed in final tree!"
+            )
         return self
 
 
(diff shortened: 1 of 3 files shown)


sschleemilch (approved):
Yeah, that's a valid one

SoundMatt:
CI caught a regression: six tests in `test_description_error/` and
`test_overlay/` asserted on the literal string `'type': 'assertion_error'`
in pydantic's error output, which becomes `'type': 'value_error'` after
the assert -> raise refactor. Pushed a follow-up commit updating those
test expectations. Same coverage, same intent — only the type-tag string
differs.

erikbosch (approved):

erikbosch:
MoM:

- Presented at meeting
- ok to merge

PR #518, opened 2026-05-05
exporters: encoding="utf-8" across text writes; samm/file_helper to context manager

SoundMatt:
## Problem

Two related file-handling hardening fixes across the exporters:

### 1. Missing `encoding="utf-8"` on text-write `open()` calls

Fourteen exporters (`json`, `tree`, `yaml`, `csv`, `protobuf`,
`plantuml`, `go`, `id`, `franca`, `ddsidl`, `stats_utils`, `apigear`,
`s2dm/__init__`, `s2dm/reference_generator`) opened output files
with `open(path, "w")` in text mode without specifying an encoding.

On platforms whose default text-file encoding is not UTF-8 (notably
Windows cp1252), writing a VSS spec containing non-ASCII characters —
e.g. `°` in unit names, `Ω` in resistance units, accented characters
in descriptions — crashes with `UnicodeEncodeError`. Same shape on
non-UTF-8 Linux locales.

`jsonschema.py` already does this correctly; everything else didn't.

### 2. `samm/helpers/file_helper.write_graph_to_file` not using a context manager

The function opened the output `.ttl` file with `output_file.open("w")`,
wrote to it, then manually called `.close()` at the end. If any of the
intermediate `.write()` calls raised, the file descriptor would leak.
It also lacked the encoding fix from #1.

## Fix

Two commits, one per problem:

1. **`exporters: add encoding="utf-8" to all text-write open() calls`**
   — adds `encoding="utf-8"` to every `open(path, "w")` (and
   `open(path, "w", newline="")` for CSV) across the 14 exporters.

2. **`exporters/samm: use context manager and utf-8 encoding in file_helper`**
   — wraps the two `.write()` calls in a `with` block and pins the
   encoding.

## Notes

- `yaml.py` had a confusing `encoding="utf-8"` kwarg passed to
  `yaml.dump()`, which is silently ignored when the dump target is a
  text-mode file (PyYAML only honours that kwarg on byte-mode
  streams). Left in place with a comment explaining why; the real
  fix is on `open()`.
- CSV's existing `newline=""` argument is preserved (correct for
  cross-platform CSV writing).
- No new tests added: the fix is mechanical and the failure mode it
  prevents is platform/locale-specific. Happy to add a test if
  maintainers can suggest a clean way to simulate non-UTF-8 default
  encoding in CI.

## Related

Companion to the validation hardening in #515, #516, and #517.

### DIFF
--- a/src/vss_tools/exporters/apigear.py
+++ b/src/vss_tools/exporters/apigear.py
@@ -400,7 +400,7 @@ def generate_solution(
 
 
 def export_yaml(file_name, content_dict):
-    with open(file_name, "w") as f:
+    with open(file_name, "w", encoding="utf-8") as f:
         yaml.dump(
             content_dict,
             f,
--- a/src/vss_tools/exporters/csv.py
+++ b/src/vss_tools/exporters/csv.py
@@ -74,7 +74,7 @@ def add_rows(
 
 
 def write_csv(rows: list[list[Any]], output: Path):
-    with open(output, "w", newline="") as f:
+    with open(output, "w", encoding="utf-8", newline="") as f:
         writer = csv.writer(f)
         writer.writerows(rows)
 
--- a/src/vss_tools/exporters/ddsidl.py
+++ b/src/vss_tools/exporters/ddsidl.py
@@ -428,10 +428,10 @@ def cli(
 
     if datatype_tree is not None:
         exporter = StructExporter()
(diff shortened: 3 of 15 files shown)


sschleemilch (approved):

erikbosch (approved):
LGTM

PR #523, opened 2026-05-22
fix(units): accept unit keys case-insensitively

### LINKED ISSUE #489: Making unit key case insensitive
The [latest release of the COVESA VSS domain model introduced a breaking change](https://github.com/COVESA/vehicle_signal_specification/blob/master/CHANGELOG.md#vss-60)

> Units Updates
>
>`celsius` changed to `Celsius` (backward incompatible change)
>Fahrenheit and some other units added
>mpg unit deprecated

The meaning of the unit did not change. However, as a result, models that were using an older COVESA VSS as a reference would need to replace `celsius` to `Celsius`. This is non-trivial when multiple users already depend on that. Especially if private branches were specified based on an old units file. So it won't be about only updating public references.

Hence, the proposal is that the tools reserve the characters of the unit key, being flexible about the naming case. For example:

- `celsius = CELSIUS = CELSius = CeLSiUs = ...` etc. 
- `celsius != celsiusDegrees != degreecelsius != gradosCentigrados != ...` etc. 

So, when loading and processing the units, the tools can simply normalize them to avoid errors. For example: converting to upper case.

SoundMatt:
Unit keys in YAML unit files and \`unit:\` fields in vspec signal
definitions are now normalised to lowercase on load.

This means \`Celsius\`, \`CELSIUS\`, and \`celsius\` all resolve to
the same unit without error, matching the expectation raised in #489.

## Changes

- \`units_quantities.py\`: lowercase each YAML key before storing it
  in the loaded-units dict
- \`model.py\` \`check_valid_unit\`: try the lowercase form first so
  that any capitalisation in a vspec file is accepted; the stored
  value is normalised to lowercase
- New test vspec \`signals_mixed_case_units.vspec\` and a corresponding
  \`test_unit_key_case_insensitive\` test case

Fixes #489

### DIFF
--- a/src/vss_tools/model.py
+++ b/src/vss_tools/model.py
@@ -420,6 +420,9 @@ class VSSDataDatatype(VSSData):
     def check_valid_unit(cls, v: str | None) -> str | None:
         if v is None:
             return v
+        v_lower = v.lower()
+        if v_lower in dynamic_units:
+            return v_lower
         if v not in dynamic_units:
             raise ValueError(f"'{v}' is not a valid unit")
         return v
--- a/src/vss_tools/model.py
+++ b/src/vss_tools/model.py
@@ -423,7 +423,7 @@ class VSSDataDatatype(VSSData):
         v_lower = v.lower()
         if v_lower not in dynamic_units:
             raise ValueError(f"'{v}' is not a valid unit")
-        return v_lower
+        return dynamic_units[v_lower].key or v_lower
 
     @model_validator(mode="after")
     def check_datatype_matching_allowed_unit_datatypes(self) -> Self:
@@ -434,7 +434,7 @@ class VSSDataDatatype(VSSData):
         if self.unit:
             if not Datatypes.get_type(self.datatype):
                 raise ValueError(f"Cannot use 'unit' with complex datatype: '{self.datatype}'")
-            allowed_datatypes = dynamic_units[self.unit].allowed_datatypes
+            allowed_datatypes = dynamic_units[self.unit.lower()].allowed_datatypes
             if allowed_datatypes is None:
                 allowed_datatypes = []
             if not any(Datatypes.is_subtype_of(self.datatype.rstrip("[]"), a) for a in allowed_datatypes):
--- a/src/vss_tools/model.py
+++ b/src/vss_tools/model.py
@@ -421,11 +421,9 @@ class VSSDataDatatype(VSSData):
         if v is None:
             return v
         v_lower = v.lower()
-        if v_lower in dynamic_units:
-            return v_lower
-        if v not in dynamic_units:
+        if v_lower not in dynamic_units:
             raise ValueError(f"'{v}' is not a valid unit")
-        return v
+        return v_lower
 
     @model_validator(mode="after")
     def check_datatype_matching_allowed_unit_datatypes(self) -> Self:
--- a/src/vss_tools/units_quantities.py
+++ b/src/vss_tools/units_quantities.py
@@ -33,6 +33,7 @@ def load_units_or_quantities(
             continue
         log.info(f"Loaded '{class_type.__name__}', file={file.absolute()}, elements={len(content)}")
         for k, v in content.items():
+            k_original = k
             k = k.lower()
             if v is None:
                 raise MalformedDictException(f"'{class_type.__name__}', '{k}' is 'None'")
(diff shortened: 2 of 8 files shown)


erikbosch:
MoM: Please review (valid for all now PRs from @SoundMatt)

src/vss_tools/model.py
```diff
@@ -420,6 +420,9 @@ def check_datatype(self) -> Self:
     def check_valid_unit(cls, v: str | None) -> str | None:
         if v is None:
             return v
+        v_lower = v.lower()
```
sschleemilch, line 423: not chaing the `v not in dynamic_units` on purpose?

SoundMatt:
Good catch @sschleemilch — the double-check was redundant and confusing.

The previous code tested `v_lower in dynamic_units` for the early-return path, then fell through to `v not in dynamic_units` (using the original mixed-case `v`). Since `dynamic_units` is keyed entirely in lowercase (after the `units_quantities.py` change), the second branch could never match anything the first hadn't already caught, making `return v` dead code.

Simplified to a single lookup:

```python
v_lower = v.lower()
if v_lower not in dynamic_units:
    raise ValueError(f"'{v}' is not a valid unit")
return v_lower
```

All 25 tests in `test_units` and `test_consistency` pass with this change.

erikbosch:
Did a check using std catalog (and csv exporter). To me it seems that this PR actually changes output as well, from `Nm` to `nm`. I would expect that capitalisation used in https://github.com/COVESA/vehicle_signal_specification/blob/master/spec/units.yaml#L508 to be used in output artefacts. 


```

1528,1531c1528,1531
< Vehicle.MotionManagement.Brake.Axle.Row2.Wheel.Left.Torque,sensor,int16,,Nm,,0,Estimated friction brake torque. Brake Torque < 0Nm.,,,
< Vehicle.MotionManagement.Brake.Axle.Row2.Wheel.Left.TorqueArbitrated,sensor,int16,,Nm,,0,Brake system internally calculated friction brake torque target. Brake Torque < 0Nm.,,,
< Vehicle.MotionManagement.Brake.Axle.Row2.Wheel.Left.TorqueFrictionMaximum,actuator,int16,,Nm,,0,Maximum wheel torque request for friction brake. Brake Torque < 0Nm.,,,
< Vehicle.MotionManagement.Brake.Axle.Row2.Wheel.Left.TorqueFrictionMinimum,actuator,int16,,Nm,,0,Minimum wheel torque request for friction brake. Brake Torque < 0Nm.,,,
---
> Vehicle.MotionManagement.Brake.Axle.Row2.Wheel.Left.Torque,sensor,int16,,nm,,0,Estimated friction brake torque. Brake Torque < 0Nm.,,,
> Vehicle.MotionManagement.Brake.Axle.Row2.Wheel.Left.TorqueArbitrated,sensor,int16,,nm,,0,Brake system internally calculated friction brake torque target. Brake Torque < 0Nm.,,,
> Vehicle.MotionManagement.Brake.Axle.Row2.Wheel.Left.TorqueFrictionMaximum,actuator,int16,,nm,,0,Maximum wheel torque request for friction brake. Brake Torque < 0Nm.,,,
> Vehicle.MotionManagement.Brake.Axle.Row2.Wheel.Left.TorqueFrictionMinimum,actuator,int16,,nm,,0,Minimum wheel torque request for friction brake. Brake Torque < 0Nm.,,,

```

SoundMatt:
Good catch @erikbosch — fixed in the latest commit.

The root cause: `load_units_or_quantities` was lowercasing the YAML key before injecting it into `VSSUnit.key`, and `check_valid_unit` was returning `v_lower` directly, so the canonical casing from `units.yaml` was lost in output.

The fix:
- `load_units_or_quantities` now stores the original YAML key casing in `VSSUnit.key` (while still using the lowercase form as the dict lookup key)
- `check_valid_unit` returns `dynamic_units[v_lower].key` — the canonical form — instead of `v_lower`
- `check_datatype_matching_allowed_unit_datatypes` updated to look up via `self.unit.lower()` since the dict is keyed lowercase

`Nm` → `Nm` again. Added a test (`test_unit_key_canonical_case_preserved`) that feeds a wrong-case unit key and asserts the output preserves the YAML-defined casing.

erikbosch (approved):
Previous observation fixed

src/vss_tools/model.py
erikbosch: This comment is resolved now, or?

SoundMatt:
@erikbosch — this has been approved since June 2, anything blocking merge on your end? Happy to rebase if upstream has drifted.

erikbosch:
> @erikbosch — this has been approved since June 2, anything blocking merge on your end? Happy to rebase if upstream has drifted.

I was away last week. Wanted to give Sebastian some time for additional comments but I think we are good to go. I will push the button :-)

PR #525, opened 2026-05-22
fix(franca): use double-quoted strings in allowed value output

### LINKED ISSUE #336: vss2franca: Output with latest VSS catalog not accepted by CommonAPI tooling
How to reproduce (Linux):

- Download zip from https://github.com/COVESA/capicxx-core-tools/releases/tag/3.2.14
- Unzip
- Run `/commonapi-core-generator-linux-x86_64 ~/vehicle_signal_specification/vss_rel_5.0-dev.fidl`

Issue copied from https://github.com/COVESA/vehicle_signal_specification/issues/735, see more information there.

**NOTE: Do we have anyone using `vspec2franca` interested in maintaining it? If not we may consider deprecating it or making it obsolete**

SoundMatt:
The `allowed` field was emitted using Python's `repr()` of the list,
producing single-quoted strings like `['January', 'February']`.  This
is not valid Franca IDL syntax, and for string values containing spaces
(e.g. `'Park Brake'`) the spaces would appear unescaped inside the
downstream parser's interpretation of the literal.

Emit string elements as `"value"` (double-quoted) and leave numeric
elements as bare literals, which is consistent with the rest of the
Franca output.

```
# Before
allowed: ['January', 'February']

# After
allowed: ["January", "February"]
```

Expected franca output updated for `test_allowed` accordingly.

Note: PR #514 proposes removing this exporter; this fix applies in the
meantime.

Fixes #336 (franca side)

### DIFF
--- a/src/vss_tools/exporters/franca.py
+++ b/src/vss_tools/exporters/franca.py
@@ -75,7 +75,8 @@ def print_franca_content(file: TextIOWrapper, root: VSSNode) -> None:
                 output += f",\n\tmax: {max}"
             allowed = getattr(data, "allowed", None)
             if allowed:
-                output += f",\n\tallowed: {allowed}"
+                formatted = ", ".join(f'"{v}"' if isinstance(v, str) else str(v) for v in allowed)
+                output += f",\n\tallowed: [{formatted}]"
             output += "\n}"
     file.write(output)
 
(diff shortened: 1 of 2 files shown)


erikbosch (approved):
LGTM.

Just a bit curious - do you have an interest in keeping Franca or was this just a regression you noticed when running "all" exporters.

SoundMatt:
Just a passing regression.

PR #526, opened 2026-05-22
feat: add C++ header-only exporter for embedded/MCU use

### LINKED ISSUE #513: C++ exporter (and why not Rust later)
There is already exporter for VSS that would allow someone to easily use the VSS catalog in a C++ application (csv, yaml, protobuf, etc ...). But sometimes, having to rely on a separate file is not always possible (microcontroller with no filesystem library, no ability to interact with OS because of permissions, etc ...).

That is why, I'd like to create an exporter for C++ (and later Rust) that would create a header only file with all the VSS Catalog in it. This way, the catalog would be compiled inside the binary.

I'll detail more in a few days, I just wanted to write this down to talk about it at today's meeting.

SoundMatt:
## Summary

- Adds `vspec export cpp-header` (new file `src/vss_tools/exporters/cpp_header.py`)
- Emits a single `.hpp` with a `VssSignal` struct and one `constexpr` aggregate per leaf signal
- Allowed-value lists become null-terminated `constexpr const char*[]` arrays declared immediately above their signal
- `VssSignal.cpp_type` maps VSS datatypes to C++ equivalents (`uint8_t`, `float`, `const char*`, `bool`, …) using the same type table as `datatypes.py`
- `--namespace` (default `vss`) and `--include-branches/--no-include-branches` CLI flags
- No new runtime dependencies; header requires only `<cstddef>` and `<cstdint>`

Resolves #513.

## Test plan

- [ ] `uv run pytest tests/vspec/test_cpp_header/` passes
- [ ] `vspec export cpp-header -s <spec> -o out.hpp` produces valid C++ (compile-checked manually with `g++ -std=c++17 -c out.hpp`)
- [ ] Signals with no allowed values emit `nullptr` for `allowed_values`
- [ ] Signals with allowed values emit a named null-terminated array above the aggregate


src/vss_tools/exporters/cpp_header.py
```diff
@@ -0,0 +1,215 @@
+# Copyright (c) 2024 Contributors to COVESA
```
erikbosch: New file, change to 2026 unless copied in full from somewhere else

tests/vspec/test_cpp_header/expected.hpp
```diff
@@ -0,0 +1,81 @@
+// SPDX-FileCopyrightText: Copyright (c) 2024 Contributors to COVESA
```
erikbosch: New file, use 2026

erikbosch: By the way - does our license checker complains if this file does not have a license text? If this tool is intended to be used by customers the output does not necessarily needs to be with a MPL license, and copyright is not necessarily only COVESA as it is based on signal definitions that might be proprietary.

I am thinking of removing the copyright/license part here, and if needed adjust the license reviewer so it does not care about this file.

erikbosch:
@jmorceaux - is this what you need?

erikbosch:
Awaiting feedback from Jeremy. before merging the file https://github.com/COVESA/vss-tools/blob/master/docs/vspec.md shall be expanded and reference a file `cpp-header.md`that gives a short overview of this tool. See how other tools are documented

JMorceaux:
Thanks @erikbosch for the ping ! Just came back from vacation, I'll definitely look at at this PR in details because it might indeed be exactly what I need :)

src/vss_tools/exporters/cpp_header.py
```diff
@@ -0,0 +1,215 @@
...
+@clo.units_opt
+@clo.types_opt
+@clo.strict_exceptions_opt
+@click.option("--namespace", default="vss", show_default=True, help="C++ namespace name.")
+@click.option(
+    "--include-branches/--no-include-branches",
+    default=False,
+    show_default=True,
+    help="Include branch nodes in addition to leaf signals.",
+)
```
JMorceaux, line 269: @SoundMatt Thank you for this PR. In the current state, your exporter would export the entire vss specs and put all signals in the .hpp. The VSS specs is starting to be a bit heavy and I fear the .text/.bss is gonna be huge if we export all signals. I think it would be nice if the user could give the tool a list of signals (as an option) to export and only these signals would be exported. This would reduce .text/.bss and better suited for MCU use.

tests/vspec/test_cpp_header/expected.hpp
```diff
@@ -0,0 +1,81 @@
...
+    const char* const* const allowed_values;  // null-terminated, or nullptr
+    const char* const default_value;
+};
+
+constexpr VssSignal A_Speed = {
+    "A.Speed",
+    "sensor",
+    "float",
+    "float",
+    "km",
```
JMorceaux: The advantage of using const char*/string is that it's pretty robust to change but if we want to make an exporter for embedded system, it is best to think about memory footprint. 

I think it would be best if an enum was created for unit/type/datatype/cpp_type

Same thing for the min/max value, I don't think a string is appropriate here

SoundMatt:
Updated in the latest commit:

- **Enums for type/datatype/cpp_type/unit** — `VssNodeType`, `VssDataType`, `VssCppType` are fixed enums covering all VSS types. `VssUnit` is a per-export enum containing only the units present in the output (so the emitted enum stays small on a filtered spec). All four replace the previous `const char*` fields.
- **`double` for min/max** — `min_value`/`max_value` are now `double` with a `kNoValue = std::numeric_limits<double>::quiet_NaN()` sentinel for unset values (requires `<limits>`).
- **Copyright year** — updated to 2026 in both the exporter source and the generated header.

---

@JMorceaux — on the `--signals` filter: before I build it, can you describe the expected UX? My current thinking is a repeatable `--signal` option that accepts dotted paths, e.g.:

```bash
vspec export cpp-header --signal Vehicle.Speed --signal Vehicle.Body.Lights.IsBackupOn -s spec/... -o out.hpp
```

Only the named paths (and their parent branches, if `--include-branches` is set) would be emitted. Glob patterns welcome too (`Vehicle.ADAS.*`).

Does that match what you had in mind, or did you want something different (e.g., a file listing paths, prefix-based filtering, etc.)? Happy to implement once we're aligned on scope.

docs/cpp-header.md
```diff
@@ -0,0 +1,144 @@
...
+// Do not edit manually.
+
+#pragma once
+
+#include <cstddef>
+#include <cstdint>
+
+namespace vss {
+
+struct VssSignal {
```
sschleemilch: Why not just `Signal`. Its already in the `vss` namespace

src/vss_tools/exporters/cpp_header.py
```diff
@@ -0,0 +1,393 @@
...
+from anytree import PreOrderIter
+
+import vss_tools.cli_options as clo
+from vss_tools import log
+from vss_tools.main import get_trees
+from vss_tools.model import VSSDataBranch
+from vss_tools.tree import VSSNode
+
+_CPP_TYPE_MAP: dict[str, str] = {
+    "uint8": "uint8_t",
```
sschleemilch, line 39: should use `datatypes.py` definitions

src/vss_tools/exporters/cpp_header.py
```diff
@@ -0,0 +1,393 @@
...
+    "double[]": "const double*",
+    "boolean[]": "const bool*",
+    "string[]": "const char* const*",
+    "numeric": "double",
+    "numeric[]": "const double*",
+    "int": "int64_t",
+}
+
+_NODE_TYPE_ENUM: dict[str, str] = {
+    "sensor": "VssNodeType::kSensor",
```
sschleemilch: Why not just `NodeType`?

src/vss_tools/exporters/cpp_header.py
```diff
@@ -0,0 +1,393 @@
...
+    "actuator": "VssNodeType::kActuator",
+    "attribute": "VssNodeType::kAttribute",
+    "branch": "VssNodeType::kBranch",
+}
+
+_DATATYPE_ENUM: dict[str, str] = {
+    "uint8": "VssDataType::kUint8",
```
sschleemilch: `DataType`

src/vss_tools/exporters/cpp_header.py
```diff
@@ -0,0 +1,393 @@
...
+    "double[]": "VssDataType::kDoubleArray",
+    "boolean[]": "VssDataType::kBooleanArray",
+    "string[]": "VssDataType::kStringArray",
+    "numeric": "VssDataType::kNumeric",
+    "numeric[]": "VssDataType::kNumericArray",
+    "int": "VssDataType::kInt64",
+}
+
+_CPP_TYPE_ENUM: dict[str, str] = {
+    "uint8_t": "VssCppType::kUint8T",
```
sschleemilch:
```suggestion
    "uint8_t": "kUint8T",
```

src/vss_tools/exporters/cpp_header.py
```diff
@@ -0,0 +1,393 @@
...
+    "const bool*": "VssCppType::kConstBoolPtr",
+    "const char* const*": "VssCppType::kConstCharConstPtr",
+}
+
+
+def _unit_enum_member(unit: str) -> str:
+    """Convert a VSS unit string to a C++ enum member name prefixed with k."""
+    s = unit
+    s = s.replace("/", "_Per_")
+    s = s.replace("^", "_Pow_")
```
sschleemilch, line 141: what vss unit contains those symbols?

src/vss_tools/exporters/cpp_header.py
```diff
@@ -0,0 +1,393 @@
...
+        lines.append(f"constexpr const char* {ident}_kAllowed[] = {{")
+        for v in allowed:
+            lines.append(f'    "{_escape(str(v))}",')
+        lines.append("    nullptr,")
+        lines.append("};")
+        allowed_ref = f"{ident}_kAllowed"
+    else:
+        allowed_ref = "nullptr"
+
+    lines.append(f"constexpr VssSignal {ident} = {{")
```
sschleemilch: I feel like this should be a real template engine like jinja2

src/vss_tools/exporters/cpp_header.py
```diff
@@ -0,0 +1,393 @@
...
+    seen_units: dict[str, str] = {}  # unit_str → enum member name
+    for node in leaf_nodes:
+        u = getattr(node.get_vss_data(), "unit", None)
+        if u and u not in seen_units:
+            seen_units[u] = _unit_enum_member(u)
+
+    unit_map = {u: f"VssUnit::{m}" for u, m in seen_units.items()}
+
+    out: list[str] = []
+    out.append("// SPDX-FileCopyrightText: Copyright (c) 2026 Contributors to COVESA")
```
sschleemilch: jinja2

src/vss_tools/cli.py
```diff
@@ -49,6 +49,7 @@ def cli(ctx: click.Context, log_level: str, log_file: Path):
         "yaml": "vss_tools.exporters.yaml:cli",
         "tree": "vss_tools.exporters.tree:cli",
         "samm": "vss_tools.exporters.samm:cli",
+        "cpp-header": "vss_tools.exporters.cpp_header:cli",
```
sschleemilch, line 52: how  about `hpp`?

SoundMatt:
@sschleemilch — pushed a round addressing your review:

- Renamed `VssSignal`→`Signal`, `VssNodeType`→`NodeType`, `VssDataType`→`DataType` (and `VssCppType`→`CppType`, `VssUnit`→`Unit` for consistency, since the same "already in `namespace vss`" reasoning applies to those too — let me know if you'd rather keep those two prefixed).
- `_CPP_TYPE_MAP` is now built from `vss_tools.datatypes.Datatypes` instead of a hand-written dict, so it can't drift from the datatypes vss-tools already treats as canonical.
- Dropped the `°`/`%`/`·`/`*` handling in `_unit_enum_member` — checked `spec/units.yaml` and none of those actually occur; only `/`, `^`, `-` do (`cm/s^2`, `mpg-uk`).
- Applied your suggested diff pattern across the board: the enum dicts now hold bare member names (`kUint8T`) rather than prefixed ones, with the prefix added in the template.
- Converted the `out.append(...)` boilerplate to a Jinja2 template (`templates/cpp_header.hpp.j2`). `jinja2` is now a dependency.
- Left `cpp-header` as the CLI export name for now rather than renaming to `hpp` — that changes the public CLI surface, so didn't want to guess. Happy to make the change if you'd still like it.

@JMorceaux — also implemented the `--signal` filter we discussed back in June: repeatable, dotted-path + glob (`Vehicle.ADAS.*`). With `--include-branches`, only the ancestor branches of matched signals come along, not the whole tree, so it still serves the original memory-footprint goal.

Also updated `docs/cpp-header.md` — the worked example wasn't actually buildable (it referenced an undeclared `Vehicle.Powertrain` branch), so I regenerated it from the real tool output and it's now covered by the test suite.

Full suite (431 tests, 4 new for `--signal`) passes, ruff/ruff-format/mypy clean, and I compiled the generated header with both g++ and clang++ to confirm it's valid C++17.

JMorceaux:
Hello @SoundMatt !

First of all, I wanted to apologize for the lack of response on my end this last few weeks ! I was away and coulnd't reach out.

Secondly, thank you again for all of your work. The `--signal` that you implemented feels like it's exactly what I needed. I'm gonna try the tool on my PC and will give you feedback ASAP !

Thank you again !

JMorceaux:
Hello again @SoundMatt,

So I tested the tool and here is my feedback:

- It compiles for C++14, which is a good thing for me.
- It would be nice to have an option `--no-description` which would remove the `const char* const description` from the `Signal` struct. I don't think the description is very useful in embedded environment so an option to remove it would be nice.
- It would also be nice if we could add a constexpr with the version of vss used during generation.

That would be the first usable version of this exporter :)

I've never developed anything in the vss-tools repository but I'll try to get my nose into it and implement the feature I just mentionned :)

src/vss_tools/exporters/templates/cpp_header.hpp.j2
```diff
@@ -0,0 +1,133 @@
...
+{% endfor %}
+};
+
+struct Signal {
+    const char* const path;
+    NodeType type;
+    DataType datatype;
+    CppType cpp_type;
+    Unit unit;
+    const char* const description;
```
JMorceaux, line 95:
```suggestion
{% if include_description %}
    const char* const description;
{% endif %}
```

src/vss_tools/exporters/templates/cpp_header.hpp.j2
```diff
@@ -0,0 +1,133 @@
...
+    nullptr,
+};
+{% endif %}
+constexpr Signal {{ signal.ident }} = {
+    {{ signal.fqn }},
+    NodeType::{{ signal.type_enum }},
+    DataType::{{ signal.datatype_enum }},
+    CppType::{{ signal.cpp_type_enum }},
+    Unit::{{ signal.unit_enum }},
+    {{ signal.description }},
```
JMorceaux, line 117:
```suggestion
{% if include_description %}
    {{ signal.description }},
{% endif %}
```

src/vss_tools/exporters/cpp_header.py
```diff
@@ -0,0 +1,315 @@
...
+            seen_units[u] = _unit_enum_member(u)
+
+    signals = [_signal_context(node, seen_units) for node in emitted_nodes]
+
+    template = _ENV.get_template("cpp_header.hpp.j2")
+    return template.render(
+        namespace=namespace,
+        units=list(seen_units.values()),
+        signals=signals,
+    )
```
JMorceaux, line 248:
```suggestion
        signals=signals,
        include_description=include_description
    )
```

src/vss_tools/exporters/cpp_header.py
```diff
@@ -0,0 +1,315 @@
...
+    "--signal",
+    "signals",
+    multiple=True,
+    default=(),
+    help=(
+        "Restrict output to specific signal paths (dotted, glob '*' supported, e.g. "
+        "'Vehicle.ADAS.*'). Repeatable. If --include-branches is set, only the "
+        "ancestor branches of matched signals are included. Omit to export every signal."
+    ),
+)
```
JMorceaux, line 280:
```suggestion
)
@click.option(
    "--include-description",
    default=True,
    show_default=True,
    help="Add the description field in the output Signal structure",
)
```

src/vss_tools/exporters/cpp_header.py
```diff
@@ -0,0 +1,315 @@
...
+    strict: bool,
+    aborts: tuple[str],
+    overlays: tuple[Path],
+    quantities: tuple[Path],
+    units: tuple[Path],
+    types: tuple[Path],
+    namespace: str,
+    include_branches: bool,
+    signals: tuple[str],
+    strict_exceptions: Path | None,
```
JMorceaux, line 295:
```suggestion
    signals: tuple[str],
    include_description: bool,
    strict_exceptions: Path | None,
```

src/vss_tools/exporters/cpp_header.py
```diff
@@ -0,0 +1,315 @@
...
+        aborts=aborts,
+        strict=strict,
+        extended_attributes=extended_attributes,
+        quantities=quantities,
+        units=units,
+        types=types,
+        overlays=overlays,
+        strict_exceptions_file=strict_exceptions,
+    )
+    content = generate(tree, namespace, include_branches, signals)
```
JMorceaux, line 313:
```suggestion
    content = generate(tree, namespace, include_branches, signals, include_description)
```

src/vss_tools/exporters/cpp_header.py
```diff
@@ -0,0 +1,315 @@
...
+        "default_value": _str_literal(default),
+    }
+
+
+def generate(
+    root: VSSNode,
+    namespace: str,
+    include_branches: bool,
+    signal_patterns: tuple[str, ...] = (),
+) -> str:
```
JMorceaux, line 206:
```suggestion
    signal_patterns: tuple[str, ...] = (),
    include_description: bool
) -> str:
```

erikbosch:
MoM:

- @JMorceaux wants to get in contact with @SoundMatt to discuss changes, potentially help with changes
- E: @SoundMatt do you have a COVESA slack channel account or some other means we can use to contact you
- D: This profile is from Matt Jones - contribution is possibly from Ford
- J: Will contact him

wiedmama:
>     * It would be nice to have an option `--no-description` which would remove the `const char* const description` from the `Signal` struct. I don't think the description is very useful in embedded environment so an option to remove it would be nice.
> 
Adding the description as comment might be also an option. This way it is available during development for a quick check and don't consumes resources in the final binary.

erikbosch:
@SoundMatt  - will you take a look at the comments and conflicts?

PR #527, opened 2026-05-22
feat: add overlay 'new:' directive for create vs modify intent

### LINKED ISSUE #227: Overlay annotations to indicate if a signal is supposed to already exist or not.
Today you can use overlays to both add new signals as well refactor existing signals. This gives a potential problem, if you do a spelling error or if a signal has been renamed or removed in standard VSS catalog your overlay may not do what you expect.

I am thinking if vss-tools possibly should support an attribute `overlay: ['add'|'refactor']` ( or similar like `new: True|False`) that could be used in overlays to describe if the overlay intend to add a new signal or change an existing signal. Then tooling could give a warning if the directive does not match reality.

```
Vehicle.Boddy.RefuelPosition:
  datatype: string
  type: attribute
  allowed: ['ALT1','ALT2']
  overlay: ['add'|'refactor']
  description: Location of the fuel cap or charge port.
```

SoundMatt:
## Summary

- Adds `new: bool | None = None` to `VSSData` so overlay authors can annotate whether a node is being created or modified
- Adds `check_overlay_directives()` in `main.py` that runs before the full overlay merge:
  - `new: true` → node must **not** exist in the base spec; errors if it does
  - `new: false` → node **must** exist in the base spec; errors if it doesn't
- `new` is added to `EXPORT_EXCLUDE_ATTRIBUTES` so it never appears in exported output
- Validation is opt-in: overlays without `new:` continue to work exactly as before

Resolves #227.

## Example

```yaml
# overlay.vspec

# This MUST already be in the base spec — enforced:
Vehicle.Speed:
  new: false
  description: Overriding speed description.

# This must NOT already be in the base spec — enforced:
Vehicle.MyCustomSignal:
  new: true
  datatype: uint8
  type: sensor
  description: Custom OEM signal.
```

## Test plan

- [ ] `uv run pytest tests/vspec/test_overlay/` — all 10 tests pass
- [ ] `new: true` on existing key → non-zero exit, "already exists in the base spec" in log
- [ ] `new: false` on absent key → non-zero exit, "does not exist in the base spec" in log
- [ ] Overlay without `new:` → unchanged behaviour
- [ ] `new:` field absent from all exported formats (JSON, YAML, CSV, etc.)


sschleemilch:
I don't really see the point of this one.

Using the example:

```yaml
Vehicle.Speed:
  new: false
  description: Overriding speed description.
```

This will not parse successfully anyway if its not in the base.

I don't understand the reasoning behind the `new: true` where you want to ensure that it does NOT exist in the base spec. Why does it matter whether it exists or not? It's overwritten anyway?

SoundMatt:
Thanks for the review @sschleemilch.

On `new: false`: VSS overlays can add new nodes, they're not modify-only. Without this guard, a typo like `Vehicle.Speeed` silently creates a new spurious signal rather than failing — that's exactly the scenario Erik raised in #227. `new: false` turns that silent mistake into an explicit error.

On `new: true`: the value is forward-compatibility. An OEM asserting `new: true` on a custom signal will get an immediate error if a future base spec update introduces that same name, rather than silently overwriting a definition they may be semantically incompatible with.

Both flags are opt-in, so existing overlays are unaffected.

erikbosch:
I have not yet tested this but I can see that this is useful, both to catch spelling errors and when updating to a newer VSS version to catch regression, like if the source signal has been renamed or removed.

This PR shall have a companion PR in VSS repo where the new syntax shall be documented, in https://github.com/COVESA/vehicle_signal_specification/blob/master/docs-gen/content/extensions/overlay.md Feel free to create one, if you have questions or do not succeed to build the documentation locally or experience problems in build let me know

SoundMatt:
Companion docs PR raised: COVESA/vehicle_signal_specification#914

Adds a new section to `docs-gen/content/extensions/overlay.md` covering both directives — `new: false` (modify intent, catches typos and upstream removals) and `new: true` (create intent, guards against future base-spec naming collisions) — with examples for each.

erikbosch (approved):
LGTM, but conflict in companion PR

sschleemilch:
I am not that happy about the attribute. Maybe it should be named `overlay` instead, which makes it clear that it controls the overlay behavior and then having an enum string controlling the behavior. Maybe:

```yaml
Vehicle.Speeed:
  overlay: add | edit
```

SoundMatt:
Rebased onto latest master — all 10 overlay tests pass.

One question before I rename: @sschleemilch, @erikbosch — what's your preference on the attribute name?

- **Current:** `new: true` / `new: false`
- **sschleemilch's suggestion:** `overlay: add` / `overlay: edit`

I'm happy to rename to either — just want consensus before making the change so we don't go back and forth. Once confirmed I'll update the implementation, tests, docs, and the companion VSS PR (#914) in one go.

wiedmama:
I was not asked, but I would prefer 
`overlay: add` / `overlay: edit`
as it shows more clearly the purpose (overlay handling) of the new attribute. Otherwise it might be misunderstood as the signal itself is new or not.

SoundMatt:
Anyone is welcome to join the conversation - hopefully no offense was caused me only tagging @sschleemilch, & @erikbosch... Very hard to guess how to directly ask ;)

I will implement if / when consensus is reached...

erikbosch:
Will bring it up in the VSS meeting today. I like the proposed "overlay" term. And off course -everyone is more than welcome to add comments to PRs/issues.

erikbosch:
MoM:

- @UlfBj : There should be a default
- E: Could give backward compatibility problems for older overlay
- U: think there should be a default, and then having it as optional
- U: No strong opinion on bool/enum
- M: Prefer overlay: add / overlay: edit
- U: If not used/found the tool should have a default use-case, and for instance use as edit.
- E: Would make this a backward incompatible change
- E: One could also have an "undefined" state as default, not giving error for any scenario. But possibly require it for overlays in vss repo, or give warning if not defined. Or highlight that it will be mandatory in future
- M: What would happen if default is "add"
- E: Old overlay not specifying add/edit would then give an error if it in practice is "edit"
- E: Do we long term want to make it mandatory to specify add or edit
- U: Keeping it optional makes it possible to keep it undefined
- The three alts we have now:
  -  `overlay: add` - give error if already exists
  - `overlay: edit` - give error if NOT already exists
  - _undefined_ - the signal may exist or not - no error - like today. keeps backward compatibility

**Decision: This is the preferred construct**

**overlay: add / overlay: edit**

SoundMatt:
Thanks for the meeting notes! Applied the naming decision — updated to `overlay: add` / `overlay: edit` in the latest push:

- `new: true` → `overlay: add` (key must not exist in base spec)
- `new: false` → `overlay: edit` (key must exist in base spec)
- Omitted = no check, backward-compatible (the default)

Test fixtures and function renamed accordingly. The companion docs PR (VSS#914) will be updated to match.

src/vss_tools/model.py
```diff
@@ -116,6 +116,7 @@ class VSSData(VSSRaw):
     constUID: str | None = None
     fka: list[str] = []
     instantiate: bool = True
+    overlay: str | None = None
```
sschleemilch: Shouldn't be a string but an enum type

src/vss_tools/main.py
```diff
@@ -198,6 +198,41 @@ def validate_tree(root: VSSNode) -> None:
...
+    overlays: tuple[Path, ...],
+) -> None:
+    """
+    Validates 'overlay:' directives on overlay nodes against the base spec.
+
+    overlay: add  — the key must NOT exist in the base spec (creating a new node)
+    overlay: edit — the key MUST exist in the base spec (modifying an existing node)
+    (omitted)     — no check; node may exist or not (backward-compatible default)
+    """
+    base_keys = set(load_vspec(include_dirs, [vspec]).data.keys())
```
sschleemilch: Not sure it is that great since it loads the vspec while `get_trees()` does that as well.
Can't we do that only once saying something like that in `get_trees()`?

```python
vspec_data = load_vspec(...)
overlay_data = load_vspec(...)

apply_overlay(vspec_data, overlay_data)
```

And let `apply_overlay()` fail in case of violations

src/vss_tools/main.py
```diff
@@ -198,6 +198,41 @@ def validate_tree(root: VSSNode) -> None:
...
+    for overlay_path in overlays:
+        overlay_data = load_vspec([overlay_path.parent] + include_dirs, [overlay_path]).data
+        for key, value in overlay_data.items():
+            if not isinstance(value, dict):
+                continue
+            directive = value.get("overlay")
```
sschleemilch: Should probably instantiate a pydantic instance of the modelled `Overlay` class

erikbosch:
@SoundMatt - could you take a  look at the latest comments from Sebastian

SoundMatt:
@erikbosch — done. Addressed all three of Sebastian's comments:

- `model.py`: `overlay` is now `OverlayDirective(str, Enum)` (`ADD`/`EDIT`) instead of a bare string, matching the existing `NodeType` enum pattern.
- `main.py`/`vspec.py`: replaced the standalone `check_overlay_directives()` pre-flight pass with `apply_overlay(base, overlay)` in `vspec.py`, which validates the directive and merges in one step. This also cuts the redundant `load_vspec()` calls — it was loading the vspec up to 3 times when overlays were present (once for base keys, once per overlay file, then again for the real tree); now it's 2 loads total.
- The directive value is now validated through the enum, so a typo like `overlay: ad` raises a clear error listing the valid values instead of silently no-op'ing.

One behavioral note worth flagging: with multiple overlay files, directives are still checked against the base spec only (unchanged from before), but the overlays are now merged with each other before the base-merge step. Only matters if two overlays touch the same key with conflicting directives, which isn't a scenario either version handled explicitly.

All 10 overlay tests + full suite (430 tests) pass, ruff/mypy clean.

sschleemilch (approved):
Looks good now. LGTM if merge conflicts are resolved

erikbosch:
Can you solve the merge conflict @SoundMatt ?

Does https://github.com/COVESA/vehicle_signal_specification/pull/914 require any updates as well. Otherwise I see no problem merging

PR #528, opened 2026-05-22
fix: id exporter now generates static UIDs for struct and property nodes

### LINKED ISSUE #439: id.py does not create identifiers for the specified structs.
It seems the tool is not considering the case of structs. Tools has to be extended to create also ids for them.

SoundMatt:
## Summary

`get_trees()` returns a `(tree, datatype_tree)` tuple. The `id` exporter was discarding the second value, so struct (`VSSDataStruct`) and property (`VSSDataProperty`) nodes defined via `--types` never received static UIDs.

One-line fix: capture `datatype_tree` and run `export_node` on it after processing the main tree.

Resolves #439.

## Test plan

- [ ] `uv run pytest tests/vspec/test_id_struct/` passes
- [ ] `vspec export id --types <types.vspec> -s <signals.vspec> -o out.yaml` — `out.yaml` contains `Types.*` entries with `staticUID` keys
- [ ] Existing `tests/vspec/test_static_uids/` tests unaffected (no `--types` used there)

### DIFF
--- a/tests/vspec/test_id_struct/test_id_struct.py
+++ b/tests/vspec/test_id_struct/test_id_struct.py
@@ -1,0 +1,44 @@
+# Copyright (c) 2024 Contributors to COVESA
+#
+# This program and the accompanying materials are made available under the
+# terms of the Mozilla Public License 2.0 which is available at
+# https://www.mozilla.org/en-US/MPL/2.0/
+#
+# SPDX-License-Identifier: MPL-2.0
+
+import subprocess
+from pathlib import Path
+
+import yaml
+
+HERE = Path(__file__).resolve().parent
+TEST_UNITS = HERE / ".." / "test_units.yaml"
+TEST_QUANT = HERE / ".." / "test_quantities.yaml"
+
+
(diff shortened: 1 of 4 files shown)


tests/vspec/test_id_struct/test_id_struct.py
```diff
@@ -0,0 +1,44 @@
+# Copyright (c) 2024 Contributors to COVESA
```
erikbosch: New file, use 2026

erikbosch (commented):
LGTM, but would like @jdacoello to take a look as well

SoundMatt:
Fixed the copyright year to 2026, rebased onto latest master.

erikbosch:
MoM: Ok to merge after maintainer approval

- D - static ids might be ambiguous if signals are renamed. BMW will not use static id any longer for ledger is introduced. Is anyone else using "vspec export id"? 
- Erik to create separate issue asking how to handle it going forward
- Ulf - ID can be useful to reduce bandwith in APIs like e.g. VISS

erikbosch (approved):
LGTM. This one can be merged independently to the discussion on whether we shall keep id or not

PR #529, opened 2026-05-22
feat: add optional 'url' field to VSSUnit for linking to standard specifications

### LINKED ISSUE #153: Description of units
As discussed in #151 the description of units is very short. Therefore we should think:
1. If a url pointing to an official unit spezification could be an alternative to the description
2. If a url pointing to an official unit spezification could be an addition to the description
3. If we want to have more detailed descriptions ourselves

SoundMatt:
## Summary

Adds `url: str | None = None` to the `VSSUnit` Pydantic model so unit definition files can reference an authoritative external specification (UCUM, BIPM SI Brochure, QUDT, etc.).

The field is entirely optional — all existing unit files load unchanged.

Example unit file entry:

```yaml
km:
  definition: Length measured in kilometers
  unit: kilometer
  quantity: length
  allowed-datatypes: ['numeric']
  url: https://ucum.org/ucum#km
```

Resolves #153.

## Test plan

- [ ] `uv run pytest tests/vspec/test_units/test_units.py::test_unit_url_field_accepted` passes
- [ ] All existing unit tests continue to pass (no regression)
- [ ] Unit files without `url:` still load correctly

### DIFF
--- a/src/vss_tools/model.py
+++ b/src/vss_tools/model.py
@@ -132,8 +132,7 @@ class VSSData(VSSRaw):
         """Give better explanation for empty description."""
         if self.description == "":
             raise ValueError(
-                "All nodes in the final tree must have a description. "
-                "Implicit branches are not allowed in final tree!"
+                "All nodes in the final tree must have a description. Implicit branches are not allowed in final tree!"
             )
         return self
 
@@ -165,6 +164,7 @@ class VSSUnit(BaseModel):
     unit: str | None
     quantity: str
     allowed_datatypes: list[str] | None = Field(alias="allowed-datatypes", default=None)
+    url: str | None = None
     key: str | None = None  # The actual unit key (e.g., 'km'), populated during loading
 
     @field_validator("quantity")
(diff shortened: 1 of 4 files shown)


sschleemilch (approved):

erikbosch:
This is related to #532 and the recently merged https://github.com/COVESA/vehicle_signal_specification/pull/900

Now with the qudt solution, do we see a need for a separate url entry. It could theoretically be useful for cases where we do not have a matching qudt, but I do not know if that is the case. Comments from @jdacoello  welcome

erikbosch:
MoM:

- Briefly discussed at meeting - possibly limited value now when we have the qudt solution

jdacoello:
Now, the public VSS catalog has a reference to the [QUDT](https://qudt.org) standard model. See: https://github.com/COVESA/vehicle_signal_specification/pull/900

QUDT is much bigger than the UCUM. And it actually contains all the annotated correspondence (i.e., mapping) to UCUM, too. Plus, it has conversion formulas, labels in multiple languages, etc.

See, for example: https://qudt.org/vocab/unit/KiloM 

If that is serving the initial purpose, I think then this PR could be closed without merging.

SoundMatt:
Closing this in light of COVESA/vehicle_signal_specification#900 — the QUDT integration provides unit URLs (and much more: multi-language labels, conversion formulas, UCUM mappings) as a first-class part of the standard catalog, which is exactly what this field was trying to enable. No need to add a separate `url` key to `VSSUnit` when the richer solution is already upstream.

Thanks @erikbosch and @jdacoello for the context, and @sschleemilch for the review.

PR #530, opened 2026-05-22
feat: support wildcard (glob) keys in overlay files

### LINKED ISSUE #193: AMM Idea: Overlays - introduce easy solution to add keys to multiple signals

SoundMatt:
## Summary

Overlay YAML keys that contain `*` are now expanded against the accumulated spec before merging, letting overlay authors apply attributes to multiple signals in one declaration:

```yaml
# Add a comment to every Row*.Signal leaf in one shot
"Vehicle.Cabin.Door.*.*.IsLocked":
  comment: Managed by fleet policy.

"A.*.Signal":
  comment: Applies to every matching row signal.
```

**Implementation** — `VSpec._expand_globs()` in `vspec.py`:

- Uses Python's `fnmatch`, where `*` matches any substring including dots (multi-segment wildcards work naturally)
- Glob keys are expanded to one concrete key per match before `deep_update` runs — ordering is deterministic (inherits from `self.data` iteration order)
- If a glob matches nothing a `WARNING` is logged and the load succeeds, allowing forward-compatible overlays

Resolves #193.

## Test plan

- [ ] `uv run pytest tests/vspec/test_overlay_glob/` — both tests pass
- [ ] `A.*.Signal` glob applies `comment:` to `A.Row1.Signal` and `A.Row2.Signal` but not `A.Other`
- [ ] A glob matching nothing logs `"matched no keys"` and exits 0
- [ ] Existing overlay tests (`tests/vspec/test_overlay/`) unchanged

### DIFF
--- a/src/vss_tools/vspec.py
+++ b/src/vss_tools/vspec.py
@@ -7,6 +7,8 @@
 # SPDX-License-Identifier: MPL-2.0
 from __future__ import annotations
 
+import copy
+import fnmatch
 from pathlib import Path
 from typing import Any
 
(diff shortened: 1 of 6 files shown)


erikbosch:
Have not had time to check, but we need to check if we need to change anything in https://github.com/COVESA/vehicle_signal_specification/blob/master/docs-gen/content/extensions/overlay.md based on this PR

SoundMatt:
Companion docs PR raised: COVESA/vehicle_signal_specification#915

Adds a "Wildcard keys" entry to the feature list and a new section to `overlay.md` covering the glob matching rules, multi-segment `*` behaviour, deterministic expansion order, and the non-fatal no-match warning — with examples for both typical use and forward-compatible overlays.

erikbosch:
MoM: Ok to merge after maintainer approval

PR #535, opened 2026-06-16
Fix s2dm regression

erikbosch:
This is fixing a regression introduced by #523, noticed when running s2dm from vss repo in CI.

- @SoundMatt : FYI
- @jdacoello : Do we run the tests in https://github.com/COVESA/vss-tools/tree/master/tests/vspec/test_s2dm in some way? a bit surprised this regression was not detected by vss-tools CI test.

### DIFF
--- a/src/vss_tools/exporters/s2dm/type_builders.py
+++ b/src/vss_tools/exporters/s2dm/type_builders.py
@@ -453,7 +453,7 @@ def _get_unit_args(leaf_row: pd.Series, unit_enums: dict[str, GraphQLEnumType]) 
     if not unit:
         return {}
 
-    unit_data = dynamic_units[unit]
+    unit_data = dynamic_units[unit.lower()]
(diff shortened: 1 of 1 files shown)


SoundMatt (approved):
Fix looks correct — `dynamic_units` keys are lowercase but VSS unit strings can be mixed-case (e.g. "RPM"), so the `.lower()` normalisation is the right fix.

On the test gap: the s2dm fixtures in `tests/vspec/test_s2dm/` only use lowercase units (`percent`, `mm`, `degrees`) so the lookup never hit a mixed-case key. Worth adding a signal with a unit like `RPM` to cover this path.

SoundMatt:
I'll work on a PR to improve the test coverage now. I'll ensure that you can edit as you see fit.

erikbosch:
MoM:

- Please review
- Will be merged tomorrow, as it brakes main

PR #536, opened 2026-06-16
test: add unit tests for type_builders and regression coverage for #523/#535

SoundMatt:
## Summary

This PR improves test coverage across two areas, motivated by the regression in #535 that went undetected until the real VSS spec was run in CI.

### 1. Unit tests for `s2dm/type_builders.py` (new file `tests/test_s2dm_type_builders.py`)

Functions that were previously only exercised indirectly through full CLI integration tests now have direct unit tests:

- **`_clean_enum_name`** — digit prefix, dot, dash, leading minus edge cases
- **`_parse_instances_simple`** — string list, `Row[1,2]` expansion, mixed list/expand
- **`resolve_datatype_to_graphql`** — all primitive types, array types, custom struct lookup, fallback
- **`create_unit_enums`** — quantity grouping, enum naming, new `angular_velocity`/`power` quantities

### 2. Regression tests for the unit case-sensitivity bug (#523 / #535)

After #523, `check_valid_unit` returns `unit_data.key` (original YAML casing, e.g. `"W"` for Watt) instead of the normalised lowercase key. `_get_unit_args` then calls `dynamic_units["W"]` but the dict only has `"w"` → `KeyError`. This is the regression fixed in #535.

The fix (`unit.lower()`) is included here so the regression tests can pass. If #535 merges first, this change is a no-op.

New test fixture `test_mixed_case_unit.vspec` uses both `unit: rpm` (lowercase key in YAML) and `unit: W` (uppercase key in YAML). The `TestUnitCaseSensitivityRegression` class verifies both generate valid unit enum fields.

**Why this catches what #535 would have caught:** A signal with `unit: W` would KeyError in `_get_unit_args` before the fix. The test would have failed on master, flagging the regression before it reached CI in the VSS repo.

### 3. `tree` exporter added to generic test matrix

The `tree` exporter had zero test coverage. Added `expected.tree` output files for 9 existing test directories (`test_datatypes`, `test_min_max`, `test_null`, `test_allowed`, `test_instances`, `test_overlay`, `test_comment`, `test_no_expand`, `test_extended`) and added `"tree"` to the exporters list in `test_generic.py`.

## Test plan

- [ ] `pytest tests/test_s2dm_type_builders.py` — 31 new tests, all pass
- [ ] `pytest tests/test_s2dm_exporter.py` — existing 15 tests still pass
- [ ] Generic `test_exporters` matrix now includes `tree` for 9 directories
- [ ] Verify `test_uppercase_unit_key_generates_schema` fails on master without the `.lower()` fix


erikbosch:
Great work.

As you may note the pre-commit check fails. We typically require a newline at end of line, and apparently the tree format as of today does not add a newline at end. I do not know if that is intentional @jdacoello There are two ways solving that, either we change so that the exporter adds a newline, or we change the config below so it ignores tree files. We already do it for some other file types. 

I will merge my PR to solve the regression in VSS repo (maybe gives an easy conflict here as same file is changed here). Then let this one be open to ask for feedback from @jdacoello

https://github.com/COVESA/vss-tools/blob/master/.pre-commit-config.yaml

Replace: 

```
      - id: end-of-file-fixer
        exclude_types: ["json", "proto"]
        exclude: ".*\\.[fidl|jsonschema]"
```


With: 

```
      - id: end-of-file-fixer
        exclude_types: ["json", "proto"]
        exclude: ".*\\.[fidl|jsonschema|tree]"
```

tests/vspec/test_allowed/expected.tree
```diff
@@ -0,0 +1,5 @@
+A
```
sschleemilch, line 1: I don't really see the point testing the tree exporter for those attributes - because it doesn't do anything with them

tests/vspec/test_comment/expected.tree
```diff
@@ -0,0 +1,11 @@
+A
+├── SingleLineNotQuoted
```
sschleemilch, line 2: Same here - the exporter does not use comment

tests/vspec/test_datatypes/expected.tree
```diff
@@ -0,0 +1,12 @@
+A
+├── UInt8
```
sschleemilch, line 2: same

sschleemilch (commented):
I don't think the tree exporter tests provide much value. The tree exporter is for dev purposes anyway and is ~10 LoC

SoundMatt:
Is there a downside to the tests? Avoids accidental changes in the future?

sschleemilch:
> Is there a downside to the tests? Avoids accidental changes in the future?

Of course tests are always welcome, if they test something of value. The `tree` exporter is a debug exporter that just prints the tree on stdout or into a file and is a very thin 10 line wrapper around `anytree`'s `RenderTree`.

erikbosch:
> > Is there a downside to the tests? Avoids accidental changes in the future?
> 
> Of course tests are always welcome, if they test something of value. The `tree` exporter is a debug exporter that just prints the tree on stdout or into a file and is a very thin 10 line wrapper around `anytree`'s `RenderTree`.

I think our generic tester as of today is constructed so that if there is no "expected.tree" file it in a folder it will just ignore testing the tree exporter for that scenario. So easy to remove an "expected.tree" file if we think it is useful. But now we have them in this PR and I see no reason in removing them. And having at least some tests for tree exporter makes sense, even if it is a small exporter.

erikbosch (approved):
LGTM

PR #544, opened 2026-07-17
`[yaml exporter]` Not sorting keys

### LINKED ISSUE #542: Enum generation in alphabetical order is difficult to follow in the .yaml files
In the .yaml files generated by the vspec tool the enum values are sorted in alphabetical order, and it's very difficult to follow the numbering when you are checking the code:

Image

It would be good to keep the order as it's in the .vspec files. The signals ordered by alphabet is ok.

sschleemilch:
# About

Not sorting keys of the exported yaml, which will keep the initial ordering of the vspec files which seems more natural.
Also does not shuffle around `enum` values alphabetically. Technically it's possible to just apply that for the `enum` dict with some additional effort. However, I think we might just preserve the original structure (as it might be intended by the writers).

Note that of course that is only relevant for human readers of the output, no difference when parsing as dicts are never guaranteeing an order when iterating.

fixes #542

### DIFF
--- a/src/vss_tools/exporters/yaml.py
+++ b/src/vss_tools/exporters/yaml.py
@@ -33,7 +33,7 @@ def export_yaml(file_name, content_dict):
             f,
             default_flow_style=False,
             Dumper=NoAliasDumper,
-            sort_keys=True,
+            sort_keys=False,
             width=1024,
             indent=2,
             encoding="utf-8",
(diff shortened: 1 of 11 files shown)


SebastianSchildt (approved):
Looks ok to me.

I was worrying a bit whether the order of signals is still "logical" in all cases but I think the proposed one-line change is better than doing some extra code to just keep enums sorted

And as mentioned: No difference when parsing anyway

marnova:
lgtm

adobekan:
MoM - Ready to merge.

PR #548, opened 2026-07-31
Bump minimum Python version to 3.11

erikbosch:
Also affects CI default

Reason:

- 3.10 reaches soon end of life See https://devguide.python.org/versions/

- Model ledger PR require 3.11

While doing this also changing so we test on 3.14 instead of 3.13.  

It seems to be difficult to get a stable ddsidl compiler environment. The one we used before does not support Python 3.11 onwards. So opting for removing those tests.

**What I would like to know - Is anyone objecting removing support for Python 3.10?**


slawr:
Makes sense given EOL and that vss-tools may be in use in production. Downstream users who can not yet migrate beyond 3.10 can always use existing version in short term.

sschleemilch (approved):

SebastianSchildt (approved):
Support increasing minimum version. Many distros have 3.14 as default now, and maintaining all is hard, as I have experienced some libraries have also split along the 10/11 line. e.g. either you have an old dependency that is not compatible with current Py versions, or current libs that do not support 3.10....

PR #550, opened 2026-08-04
feat: branch `description` optional, defaulting to ""

### LINKED ISSUE #549: avro output wrongly complains that branch descriptions are missing
See https://github.com/COVESA/vehicle_signal_specification/pull/932
The [Standard Build Check](https://github.com/COVESA/vehicle_signal_specification/actions/runs/30810936818/job/91677237997?pr=932) command 
vspec export avro -u ./spec/units.yaml --strict -s ./spec/VehicleSignalSpecification.vspec -t ./spec/VehicleDataTypes.vspec -o avro --namespace xxx
fails for an "asymmetric" wheel configuration with 2 front and 4 rear wheel vehicle. The error message is 
'Vehicle.Chassis.Axle.Row1' has 1 model          main.py:250
                    error(s):                                                   
                    [                                                           
                        {                                                       
                            'type': 'value_error',                              
                            'loc': (),                                          
                            'msg': 'Value error, All nodes in the               
                    final tree must have a description. Implicit                
                    branches are not allowed in final tree!',                   
                            'input': {'fqn':                                    
                    'Vehicle.Chassis.Axle.Row1', 'type': 'branch'}              
                        }                                                       
                    ]                                                           
The yaml tree has all the descriptions on place, including for the Vehicle.Chassis.Axle.Row1 branch, but the avro output fails.
What's wrong?

sschleemilch:
# About

The `description` is often not that useful and having it as required feels to be too much.

Additionally, we are less flexible with modelling and handling non expanded trees since you are not allowed to model signals that are attaching to expanded instances in that case.

Therefore, making `description` optional.

fixes #549


erikbosch (approved):
LGTM

PR #563, opened 2026-09-09
fix: reintroduce extended attributes support in jsonschema exporter

sebroy-oti:
- Reintroduce processing of extended attributes in jsonschema export. The functionality had disappeared between 5.0 and 6.0 in this commit: https://github.com/COVESA/vss-tools/commit/526754f388303f401584d100230c784fe2e62b4c
- Group extra attributes inside a x-extra object containing all the mappings, as requested by Sebastian Schleemilch
- Update doc accordingly.

### DIFF
--- a/src/vss_tools/exporters/jsonschema.py
+++ b/src/vss_tools/exporters/jsonschema.py
@@ -153,7 +153,7 @@ def add_node(
     if extend_all_attributes:
         add_x_attributes(schema, node)
     for field in vss_data.get_extra_attributes():
-        schema[field] = getattr(vss_data, field)        
+        schema[field] = getattr(vss_data, field)
     if isinstance(node.data, VSSDataDatatype):
         ref = schema
         if is_array(node.data.datatype):
--- a/src/vss_tools/exporters/jsonschema.py
+++ b/src/vss_tools/exporters/jsonschema.py
@@ -146,11 +146,14 @@ def add_node(
     require_all_properties: bool,
     extend_all_attributes: bool,
 ) -> None:
+    vss_data = node.get_vss_data()
     schema["type"] = "object"
-    if node.get_vss_data().description is not None:
-        schema["description"] = node.get_vss_data().description
+    if vss_data.description is not None:
+        schema["description"] = vss_data.description
     if extend_all_attributes:
         add_x_attributes(schema, node)
+    for field in vss_data.get_extra_attributes():
+        schema[field] = getattr(vss_data, field)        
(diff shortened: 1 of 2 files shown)


src/vss_tools/exporters/jsonschema.py
```diff
@@ -146,11 +146,14 @@ def add_node(
...
+    vss_data = node.get_vss_data()
     schema["type"] = "object"
-    if node.get_vss_data().description is not None:
-        schema["description"] = node.get_vss_data().description
+    if vss_data.description is not None:
+        schema["description"] = vss_data.description
     if extend_all_attributes:
         add_x_attributes(schema, node)
+    for field in vss_data.get_extra_attributes():
+        schema[field] = getattr(vss_data, field)
```
sschleemilch: Personally I would group them in e.g a `x-extra` object to make it explicit that those properties are not related to a JSON schema and to avoid possible conflicts.

sschleemilch (changes requested):

src/vss_tools/exporters/jsonschema.py
sebroy-oti: Added. I originally left it as it was in 5.0 to avoid breaking changes, but I agree it's better to group those custom fields separately.

erikbosch:
We are preparing the 6.1 release, and are currently at the state with release candidates. Do you see this as a fix that better should be included also in 6.1 (and not just the release following 6.1)

sschleemilch (approved):

sebroy-oti:
> We are preparing the 6.1 release, and are currently at the state with release candidates. Do you see this as a fix that better should be included also in 6.1 (and not just the release following 6.1)

We recently bumped vss-tools to 6.0 in our solution and had to create a fork that includes this fix as we require custom attributes to be accessible in the generated JSON Schema. In our case, it would be preferable if this fix made it into 6.1 so we could drop our vss-tools fork and use upstream 6.1. However, it's not essential it makes it to 6.1, as we can continue using a fork until the release following 6.1.

erikbosch:
Thanks @sebroy-oti - I think there will be no problem to include it. I will take a look at it and possibly then create a new release candidate later today or early next week including it

erikbosch:
Merged and cherry-picked to 6.1 branch. I created a new release candidate for 6.1 vss-tools (but have not yet created a new vss release candidate referencing it). 

How to install the release candidate:

`pip install --upgrade --pre vss-tools==6.1.0rc3`

PR #564, opened 2026-09-16
feat(avro): resolve struct-typed datatype references

Ekdall:
## Motivation

We ran into this gap while generating AVRO schemas for a struct catalogue that uses struct-typed properties (a property whose `datatype` points at another struct, e.g. a `Contact` struct with a `HomeAddress` property of type `Address`). The exporter currently only handles structs nested as direct tree children, so this pattern aborts with `ValueError: Unsupported VSS datatype for AVRO mapping`.

## What changed

The AVRO exporter previously only inlined structs that were nested as direct tree children. A property whose datatype references another struct (e.g. `datatype: Types.Address` or `datatype: Types.Address[]`) was not handled and caused:

```
ValueError: Unsupported VSS datatype for AVRO mapping: 'Types.Address'
```

This adds struct-reference resolution:

- `build_struct_index()` indexes every struct in the data type tree by FQN and short name.
- `resolve_struct_datatype()` resolves a property's datatype to a struct node (and whether it is an array-of-struct reference) when it is not a primitive.
- `collect_records_ordered()` replaces the nested-only record collection with a dependency-ordered (post-order), de-duplicated, cycle-safe traversal that includes both nested tree children and structs reached via a datatype reference. The top-level struct is always declared last.
- `generate_protocol()` and the CLI now build/pass this struct index so referenced structs are declared before use, exactly like nested structs already were.

Self-referential and circular struct definitions continue to be rejected by vss-tools' own model validation before this exporter runs, so the cycle guard here is defensive rather than something reachable via a valid vspec file today.

Backward compatible: existing nested-struct and primitive-datatype behaviour is unchanged, and `generate_protocol()`/`collect_enums()`/`collect_nested_structs_ordered()` keep their previous signatures and semantics for callers that do not need struct-reference resolution.

Adds Address/Contact test fixtures (single + array struct reference) and updates `docs/avro.md` accordingly.

## Testing

- `pytest tests/test_avro_exporter.py` passes (67 tests: 46 pre-existing + 21 new).
- `pre-commit run --files src/vss_tools/exporters/avro.py tests/test_avro_exporter.py tests/vspec/test_avro/types.vspec docs/avro.md` and `mypy src/vss_tools/exporters/avro.py` are clean.
- Verified end-to-end against a real-world struct catalogue with struct-typed properties (single reference and array-of-struct reference): all references resolve, dependency order is correct, and output is otherwise unchanged for catalogues without struct-typed properties.


erikbosch:
Thanks for your PR. I will present it at the VSS meeting next week

erikbosch:
MoM: Please review

PR #566, opened 2026-09-21
Task 562/fix handling of duplicate signals

Kostadin-Ivanov:
Updates done:

1. Fix handling of **top_vss_tree_unique_node_names** by setting more stable node key handling for better identification of repetitive nodes
2. Update handling of ttl_name so it can be used better for nodes property name definitions
3. Update handling of PREFERRED_NAME so to preserve node name casing and parent name casing, if needed, when human readable PREFERRED_NAME value is defined for the corresponding node
4. Update SAMM unit for Celsius and add a new SAMM unit for the VSS "ms" unit
5. Update string_helper to handle better CamelCase strings and split those for the more "human friendly" "split camel case strings"
6. Replace some hardcoded strings with corresponding SammCConcepts and SammConcepts constants


erikbosch:
MoM:

- Please review

